PR #4, opened 2022-10-23
fix: prevent vanilla from using itemID for comparing ammunition types

Darxo:
Vanilla BB commited one of the great sins in programming: Assign the same ID multiple times for different objects.
`powder_bag.nut` has the same ID as `large_powder_bag.nut`
`quiver_of_arrows.nut` has the same ID as `large_quiver_of_arrows.nut`
`quiver_of_bolts.nut` has the same ID as `large_quiver_of_bolts.nut`

**Why is that?** 
Because of some parts of the codebase that are responsible for checking whether you have the correct ammunition for the equipped ranged weapon. They compare the result of `getID()`of the equipped ammunition against the result of `getAmmoID()` of the equipped ranged weapon.
Meanwhile other parts of the code-base correctly use the id-entries of the enum `Const.Items.AmmoType` when checking for certain ammunition types

**Challenge:**
- Those old wrong checks between `getID()` and `getAmmoID()` are burried somewhat deep in three different functions. Overwriting those should be the last resort.
- Overwriting `getID()` is not an option. It is supposed to return an ItemID and not some ammunition-type.

**Solution:** (3 steps)
- (1) We define that from now on `function getAmmoID()` on ranged weapons is supposed to return a type from the `Const.Items.AmmoType` enum. This is easily fixable retroactively by refactoring the return values whenever that function is about to return one of the old ItemIDs
- (2) We do a Switcheroo for all the three wrong vanilla functions: When they are called we make it so `getID()` of the ammo item is switched out with the `getAmmoType()` of that same ammo item. And afterwards they are switched back. The Vanilla code stays the same but it now compared the right kind of types with each other.
- (3) We finally define new actual unique ID's for the three large ammunition variants. And moving forward we can have more ammunition container for existing ranged weapons with their own unique ID

### DIFF
--- a/mod_reforged/hooks/items/ammo/ammo.nut
+++ b/mod_reforged/hooks/items/ammo/ammo.nut
@@ -1,0 +1,76 @@
+::mods_hookExactClass("items/ammo/ammo", function(o)
+{
+    o.m.AmmoTypeName <- "ammo_type_name";	// Must be overwritten with an actual informative name
+    o.m.AmmoWeight <- 0.0;			// Every Ammo in this bag applies this value as a negative StaminaModifier
+	o.m.StaminaModifier <- 0;		// Flat StaminaModifier. In Vanilla this already exists on most other equipables
+
+    local create = o.create;
+    o.create = function()
+    {
+        create();
+		this.m.ItemType = ::Const.Items.ItemType.Ammo;		// This is a line that Vanilla just forgot to include in this shared parent class
+    }
+
+    local onEquip = o.onEquip;
+    o.onEquip = function()
+    {
+		onEquip();
+		this.addGenericItemSkill();		// Now that ammunition by design can inflict a Staminamodifier and potentially more effects we force an addGenericItemSkill call
+    }
+
+// Function overwrites of 'item.nut' functions
+
+    // New function that new ammunition items can use to reduce the amount of copying the same lines
+    o.getTooltip <- function()
+    {
+        local ret = this.item.getTooltip();   // Name + Description + Category + Value + image
+
+		if (this.getStaminaModifier() < 0)
+		{
+			ret.push({
+				id = 8,
+				type = "text",
+				icon = "ui/icons/fatigue.png",
+				text = "Maximum Fatigue " + ::MSU.Text.colorRed(this.getStaminaModifier())
+			});
+		}
+
+		if (this.m.Ammo != 0)
+		{
+			ret.push({
+				id = 6,
+				type = "text",
+				icon = "ui/icons/ammo.png",
+				text = "Contains " + ::MSU.Text.colorRed(this.m.Ammo) + " " + this.m.AmmoTypeName + "s"
+			});
+		}
+		else
+		{
+			ret.push({
+				id = 6,
+				type = "text",
+				icon = "ui/tooltips/warning.png",
+				text = ::MSU.Text.colorRed("Is empty and useless")
+			});
+		}
+        ret.push({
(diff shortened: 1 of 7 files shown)


mod_reforged/hooks/items/ammo/ammo.nut
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    local onEquip = o.onEquip;
+    o.onEquip = function()
+    {
+		if (this.m.AddGenericSkill)
+		{
+			this.addGenericItemSkill();
+		}
+    }
```
Enduriel, line 21: You never call the original function

mod_reforged/hooks/items/ammo/ammo.nut
```diff
@@ -0,0 +1,76 @@
...
+            icon = "ui/icons/asset_ammo.png",
+            text = "Refill cost per ammunition: [b]" + ::MSU.Text.colorRed(this.m.AmmoCost) + "[/b]"
+        });
+        return ret;
+    }
+
+    o.getStaminaModifier <- function()
+    {
+        return -1 * ::Math.ceil(this.getAmmo() * this.m.AmmoWeight);
+    }
```
Enduriel, line 70: This, along with the other AmmoWeight stuff, should be a separate (design) PR

mod_reforged/hooks/skills/special/no_ammo_warning.nut
```diff
@@ -0,0 +1,17 @@
...
+    o.isHidden = function()
+    {
+        local ammo = this.getContainer().getActor().getItems().getItemAtSlot(::Const.ItemSlot.Ammo);
+        if (ammo == null) return isHidden();
+        local oldGetID = ammo.item.getID;
+        ammo.item.getID = ammo.getAmmoType;
+
+        local ret = isHidden();
+
+        ammo.item.getID = oldGetID;
```
Enduriel, line 14: Not the *biggest* fan of this, I think it'd be better if we also handle the situation where someone creates a child getID function, but if you don't want to put in the work I understand since it is fairly niche.

Although reading back, you do this differently here than in actor.nut getRangedWeaponInfo, which is inconsistent.

mod_reforged/hooks/items/ammo/ammo.nut
```diff
@@ -0,0 +1,76 @@
+::mods_hookExactClass("items/ammo/ammo", function(o)
+{
+    o.m.AmmoTypeName <- "ammo_type_name";
+    o.m.AmmoWeight <- 0.0;
+    o.m.AddGenericSkill <- true;
```
Enduriel, line 5: When would this be false?

Enduriel (changes requested):
Overall I'm kind of on the fence regarding this being *necessary*, but I agree that the vanilla system here is ridiculous so I guess this is fair enough.

mod_reforged/hooks/items/ammo/ammo.nut
Darxo: I don't know. I copied the way Vanilla did it in other places. In those places this variable is false by default, probably so that those items don't polute the skill-container when not necessary.

I'll remove it

mod_reforged/hooks/skills/special/no_ammo_warning.nut
Darxo: I haven't considered the cases with childs having that function at all. Not sure how to best deal with that.

About the second issue: I revisited those three hooks and unified them now

PR #10, opened 2023-01-07
feat: implement burden mechanic for equipped items

Darxo:
### Explanation:
The **Burden** of an item is equal to the Fatigue penalty it inflicts onto the player **after** fatigue reduction perks like **Brawny**.
The **Burden** of an Actor is equal to the combined **Burden** from every item that they have equipped or that is in their bag.

(1) **Burden** is very useful for a more correct Initiative Calculation. Currently any buffs to Fatigue (Like the 'Strong' perk) also increase your Initiative by the same amount because in Vanilla there is no way to differentiate between Fatigue gained/lost from equipment and Fatigue gained/lost from injuries, effects or traits. All of them affect Initiative.
Now (after this proposed fix) if someone wants an effects to give an equal amount of Fatigue and Initiative they just increase both values explicitly.

(2) **Burden** allows new types of perks/effects which change how your Itemweight affects your Initiative via the **BurdenMult**.
A perk could reduce the Initiative Penalty from your **Burden** by 50% reducing the negative effect on the turn-order from heavy armor significantly.
A debuff could vastly increase the **BurdenMult** and punishing brothers with heavier armor.

(3) **Burden** allows new effects/perks that have interaction with this value of a character. They could grant any type of buff or debuff depending on the current **Burden** of the target.

### Implementation:
Currently the **Burden** is just equal to the Stamina difference that any type of equipment applies during their `onUpdateProperties` function. This does include the 30% effect of the perk **Brawny** and the effect of **Bags and Belts** because those are hard-coded inside these functions. 

### Current Problems:
- It does not however consider any other custom perks which try to realize those fatigue reducing effects outside. (Like **Wears It Well** from PTR)
- [positive] Fatigue Buffs from items are treated as negative Burden. In Vanilla this does not exist but if we introduce Accessories which have no inherit FatigueModifier but grant Fatigue through their magical effect then this will negate some of the Burden and therefor double-dip as an Initiative Boost again.


Darxo:
This is likely no longer needed as this branch makes it obsolete with a more complete Weight & Burden system:
https://github.com/Battle-Modders/mod-reforged/pull/51

PR #12, opened 2023-01-08
fix: prevent AI from targeting mortars

Darxo:
Mortars are not meant to be killed but in Vanilla, But AI still targets them. This is frustrating when allies waste their ranged attacks on mortars or your dogs get hung up on them instead of pinning down the engineers or fleeing enemies.

This PR reduces those instances to a minimum by tanking the value of mortars.

This is an exact copy from an AI-Fix-Collection mod of mine and I ran this in a longer campaign without issues. Though I can't remember this exact case of AI-Targeting was actually fixed.

### DIFF
--- a/mod_reforged/hooks/ai/tactical/behavior.nut
+++ b/mod_reforged/hooks/ai/tactical/behavior.nut
@@ -1,0 +1,11 @@
+::mods_hookBaseClass("ai/tactical/behavior", function(o) {
+    o = o[o.SuperName];
+
+    local queryTargetValue = o.queryTargetValue;
+    o.queryTargetValue = function( _entity, _target, _skill = null )
+    {
+        // Currently Mortars are near invincible and not meant to be killed. But AI still targets them sometimes. This line should make it so they are basically never targeted.
+        if (_entity.getType() == ::Const.EntityType.Mortar) return 0.01;
+        return queryTargetValue(_entity, _target, _skill);
+    }
+});


mod_reforged/hooks/ai/tactical/behavior.nut
```diff
@@ -0,0 +1,12 @@
+::mods_hookBaseClass("ai/tactical/behavior", function(o)
```
LordMidas, line 1: For hooks the convention is to use the bracket on the same line, and space after `function` (basically like adam wrote it in his mod hooks scripts). So: `function (o) {`

mod_reforged/hooks/ai/tactical/behavior.nut
```diff
@@ -0,0 +1,12 @@
...
+{
+    o = o[o.SuperName];
+
+    local queryTargetValue = o.queryTargetValue;
+    o.queryTargetValue = function( _entity, _target, _skill = null )
+    {
+        // Currently Mortars are near invincible and not meant to be killed. But AI still targets them sometimes. This line should make it so they are basically never targeted.
+        if(_entity.getType() == ::Const.EntityType.Mortar) return 0.01;
+        return queryTargetValue( _entity, _target, _skill = null );
```
LordMidas, line 10: Don't use `_skill = null` here. Also remove spaces before and after the args. So: `queryTargetValue(_entity, _target, _skill)`

mod_reforged/hooks/ai/tactical/behavior.nut
```diff
@@ -0,0 +1,12 @@
+::mods_hookBaseClass("ai/tactical/behavior", function(o)
+{
+    o = o[o.SuperName];
+
+    local queryTargetValue = o.queryTargetValue;
+    o.queryTargetValue = function( _entity, _target, _skill = null )
+    {
+        // Currently Mortars are near invincible and not meant to be killed. But AI still targets them sometimes. This line should make it so they are basically never targeted.
+        if(_entity.getType() == ::Const.EntityType.Mortar) return 0.01;
```
LordMidas, line 9: I see that the original `queryTargetValue` function takes into account the target's HP. Do you have an estimate of what the return value is for mortars? How big is it compared to this new `0.01`?

LordMidas (changes requested):

mod_reforged/hooks/ai/tactical/behavior.nut
Darxo: I just looked into the queryTargetValue and while I don't understand everything the following line should give some sort of answer:
`score = score + hitpoints * this.getProperties().TargetPriorityHitpointsMult;`
`hitpoints` is the fraction of expectedDamage/remainingHitpoints
`TargetPriorityHitpointsMult` is some multiplier value, between 0.15 and 0.50 across the board.

But apart from this there are about 20 other lines of code that all add to `score`. For Mortar most of them will likely not fire.
The most notable one that triggers is
`score = score + hitchance / 100.0 * this.getProperties().TargetPriorityHitchanceMult * this.Const.AI.Behavior.GlobalHitchanceValueMult;`
Every AI except Alps will consider Mortars an interesting target just because they have a realistic chance of hitting it. And this is basically always greater than 0.01.

Enduriel (approved):

PR #27, opened 2023-01-21
feat: add new trait 'swindler' which pretends to be another random trait

Darxo:


Darxo:
I will keep this branch and the idea locally but close the PR for now.
I will make it return at a later point

PR #30, opened 2023-01-22
feat: taxidermist beast part price mult addition

Uberbagel:
The base beast parts price mult is 0.15. This branch adds a 2.0 price mult for beast parts at settlements with a taxidermist. This is the same mult present in the Collectors settlement situation.


LordMidas (changes requested):
This PR is invalid in its current state. We will actually have to look at the codebase and implement a system in which such a thing as this PR intends to accomplish can be done - as from a cursory look I do not see a clean/efficient method available by default.

PR #46, opened 2023-02-07
new random names

Darxo:
- Increases the amount of random names by roughly 40%
- Knight-Names are now automatically generated from the random names instead of being a separate hard-coded list.
- CharacterNames and KnightNames are once checked for duplicates and those are discarded


mod_reforged/hooks/config/character_names.nut
```diff
@@ -0,0 +1,105 @@
...
+    "Wido",
+    "Wigand",
+    "Wighart",
+    "Wilfred",
+    "Wolfhart",
+    "Wulfric"
+]);
+
+// This will also void any custom names from other mods that execute before Reforged
+::Const.Strings.KnightNames = [];
```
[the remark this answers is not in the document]
Darxo: > These are not strong enough arguments in my opinion to completely overwrite the KnightNames array and therefore enforce incompatibility with mods loading before Reforged.
I don't see the point to even plan for knight-exclusive names at this point when they will be buried under 300 other names.
If I could I wouldn't even be using KnightNames and instead use a `function generateKnightName()` but that wouldn't be  backwards compatible

If those mods (We are talking about probably 0,1%) really want their custom unique KnightNames anyways which appear once every 300 spawns they can manipulate the load-order. 
Just like if a mod wanted to modify `onAnySkillUsed` for `thrust`, `slash`, `gash` etc.
They probably throw most of those new KnightNames into the CharacterNames array anyways.

They can even replace the KnightNames to go back to a small hand-picked collection of very exotic names.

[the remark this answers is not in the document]
LordMidas: There is a difference between a mechanical change that happens due to our design intention of how a skill should behave vs. a cosmetic mod that changes e.g. the names generated for entities. The argument for `thrust`, `slash`, `gash` etc. is not analogous here because the design of those skills is connected to the balance of the mod. Nevertheless, those changes to those skills are currently temporary and we can/should implement a more compatible system.

In my opinion, enforced breaking of compatibility for such a small cosmetic thing, based on the arguments you gave i.e. (a) a few names feel weird to you and (b) you don't want to deal with two separate arrays, is both unnecessary and unwise.

mod_reforged/hooks/msu.nut
```diff
@@ -8,6 +8,25 @@
 	return obj;
 }
 
+// Removes all duplicates from a string array and returns a new unique array
+::logInfo("Reforged::MSU -- adding ::MSU.Array.removeDuplicateStrings");
```
LordMidas, line 12: This is just a "btw comment": This should ideally be `::MSU.Array.removeDuplicates` function which works with any data type. But it's ok for now as I don't think we will be keeping this long term.

mod_reforged/hooks/msu.nut
```diff
@@ -8,6 +8,25 @@
...
+{
+	local _nonDuplicatesArray = [];
+	local _arrayTable = {};
+	foreach( string in _array )
+	{
+		::MSU.requireString(string);
+		_arrayTable[string] <- null;	// The value doesn't matter here. We just put in some random value
+	}
+
+	foreach( nonDuplicateString, value in _arrayTable )
```
LordMidas, line 23: Use `::MSU.Table.keys` as that has a more optimized algorithm instead of this local `_nonDuplicatesArray`. You can simply `return ::MSU.Table.keys(_arrayTable);`

mod_reforged/hooks/config/character_names.nut
```diff
@@ -0,0 +1,108 @@
...
+]);
+
+// This removes all duplicate CharacterNames. This line is redundant for Vanilla + Reforged because our additional names are already checke for duplicates
+// Any mod that adds names after ours may re-introduce duplicates. This is not game-breaking but still undesired
+::Const.Strings.CharacterNames = ::MSU.Array.removeDuplicateStrings(::Const.Strings.CharacterNames);
+
+local knightPrefix = "Sir ";
+foreach(characterName in ::Const.Strings.CharacterNames)
+{
+    ::Const.Strings.KnightNames.push(knightPrefix + characterName);
```
LordMidas, line 104: Why do you use a local variable `knightPrefix` instead of just `"Sir " + characterName`?

LordMidas (changes requested):
Looks fine. Just a few minor comments.

mod_reforged/hooks/config/character_names.nut
```diff
@@ -0,0 +1,108 @@
...
+    "Widald",
+    "Wido",
+    "Wigand",
+    "Wighart",
+    "Wilfred",
+    "Wolfhart",
+    "Wulfric"
+]);
+
+// This removes all duplicate CharacterNames. This line is redundant for Vanilla + Reforged because our additional names are already checke for duplicates
```
LordMidas, line 97: typo: `checked`

mod_reforged/hooks/config/character_names.nut
Darxo: Just personal preference at the point in time when I wrote that.

LordMidas (approved):

PR #49, opened 2023-02-10
fix: add missing tooltip line about the damage reduction of FollowUp

Darxo:


LordMidas:
I've instead added the damage reduction to the tooltip of the active skill and this information will then be viewable in the nested tooltip of the active skill in the perk tooltip.

PR #51, opened 2023-02-10
feat: implement the basic Weight & Burden mechanic

Darxo:
- 'Brawny' and 'Wears It Well' now use this new system
- fixes Initiative calculation bug in vanilla: Buffs and Debuffs to fatigue no longer modify Initiative
- StaminaMultiplier now scales Stamina before gear, not after


Enduriel (approved):
Overall Looks good, I'm a little worried about the complexity of this setup and the one suggestion I would have would be to add some helper functions to CharacterProperties so that
```squirrel
_properties.WeightMult[::Const.ItemSlot.Body] *= 0.7;
_properties.WeightMult[::Const.ItemSlot.Head] *= 0.7;
```
could be one line, but other than those two things technically speaking it looks fine.

Enduriel (changes requested):
Need some more time to review some concerns

PR #55, opened 2023-02-11
fix: add calculation for 'Malus' variable so FromAllSides actually works

Darxo:


LordMidas:
It required some further fixes too. So, instead of this PR I've implemented a different fix for this which includes a `getMalus()` function and also fixed how the stacks are being calculated.

Darxo:
Sure, your implementation fixed that but.
But I don't know what other fixes you are talking about. I'm certain this PR fixed all the issues with it

PR #57, opened 2023-02-12
feat: introduce and apply the two keywords 'trapped' and 'out of action'

Darxo:
These keywords would make it much easier to explain the synergies and conditions of several perks/skills in one word instead of having to count every variation of those effects by hand each time.

I'm not happy with the name **Out of Action**. It's not a keyword in the literal sense and weird using it in certain sentences like 
`grants immunity to [being] out of action`
`grants immunity to being trapped` sounds just fine in contrast.

There is also a small argument to skip **Out of Action** because it only encompasses two sub-effects. And I don't see us introducing a frozen effect anytime soon.


LordMidas:
While I appreciate the direction of this PR I'm not sure if the perk descriptions need to be updated everywhere to use this. E.g. let's say we want to modify Executioner to now ignore Webbed but still work on Rooted or Stunned - we'd have to go and update either the Trapped concept or the Executioner tooltip to now again mention all the status effects manually by hand. I'm not saying this is a good example - but it's a proof of concept.

Mentioning each status effect manually, while tedious, ensures that each perk mentions exactly the effects that are valid for it.

Having the two concepts, especially the OutOfAction is quite nice imo as they can be used in useful places. It would also be nice to have a central place for a list of status effects that "Trapped" and "OutOfAction" use - so that we can then refer to that in the code of the perk and then mention the concept in the tooltip.

PR #59, opened 2023-02-14
Shield rework

Uberbagel:
This is the first phase of shield damage and shield condition rework.

Phase One: Rebalance  Shield Damage and Shield Condition to make both shields and shield splitting useful.

Phase Two: Increase Shield Damage on regular attacks to be both useful and balanced.

Phase Three: Shield Damage that exceeds a Shield's remaining condition deals the remaining damage to the target.

This PR:
I've balanced our current lineup against each other and against each shield in the game. I feel strongly I have a pretty good balance here. You can see the whole list of shield damage by weapon in the weapons spreadsheet. You can use this link to quickly see shield conditions (https://battlebrothers.fandom.com/wiki/Armor#Shields). If the numbers are too hard to parse and come on up with an understanding of the balance, let me know and I can expand upon my design intentions.

Edit: I forgot I also rebalanced orc axe fatigue cost because it was wrong. I can split up this commit if necessary.

### DIFF
--- a/mod_reforged/hooks/items/shields/kite_shield.nut
+++ b/mod_reforged/hooks/items/shields/kite_shield.nut
@@ -1,0 +1,9 @@
+::mods_hookExactClass("items/shields/kite_shield", function(o) {
+	local create = o.create;
+	o.create = function()
+	{
+		create();
+		this.m.Condition = 42;
+		this.m.ConditionMax = 42;
+	}
+});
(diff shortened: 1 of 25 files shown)


Darxo:
These numbers are hard to parse for me.

I don't know what problems of the vanilla values you are trying to fix
You state the goal of you wanting

> to make both shields and shield splitting useful

So, how useful do you see shields before and after your change? And likewise with Shield Splitting.

Also for those balance passes some summary would be very helpful for the other devs. Something like these types of items/skills are on average this % weaker/stronger.

LordMidas:
We did an overview together and added the changes discussed in our meeting.

PR #61, opened 2023-02-15
refactor: change encumbrance to increase fatigue at turn start

LordMidas:
This _is_ how the encumbrance mechanic was originally conceptualized to work. Currently, the way vanilla works is that Fatigue Recovery is only relevant until you reach maximum fatigue. At that point, you always regain 15 Fatigue. This makes traits such as Asthmatic, and injuries which affect Fatigue Recovery basically irrelevant at that stage. It also makes it so that the Encumbrance mechanic doesn't work as intended.

To change the complete mechanic so that Fatigue Recovery is always used at Max Fatigue instead of a minimum of 15 would require overwriting actor.nut `onTurnStart` function or doing an ugly switcheroo on it as it touches many things. So, instead of doing that, this PR makes it so that only the "Encumbrance" mechanic overrides the min. 15 Fatigue recovery mechanic, which makes it work as intended and is, imo, an ok compromise for now.

EDIT: Based on the discussion on the PR we decided to change Encumbrance to now "increase Fatigue by x at the start of a turn" rather than change "FatigueRecovery" property.

### DIFF
--- a/scripts/skills/effects/rf_encumbrance_effect.nut
+++ b/scripts/skills/effects/rf_encumbrance_effect.nut
@@ -38,7 +38,7 @@ this.rf_encumbrance_effect <- ::inherit("scripts/skills/skill", {
 					id = 10,
 					type = "text",
 					icon = "ui/icons/fatigue.png",
-					text = "[color=" + ::Const.UI.Color.NegativeValue + "]-1[/color] Fatigue Recovery per turn"
+					text = "[color=" + ::Const.UI.Color.NegativeValue + "]+1[/color] Fatigue at the start of every turn"
 				});
 				break;
 
@@ -48,7 +48,7 @@ this.rf_encumbrance_effect <- ::inherit("scripts/skills/skill", {
 						id = 10,
 						type = "text",
 						icon = "ui/icons/fatigue.png",
-						text = "[color=" + ::Const.UI.Color.NegativeValue + "]-2[/color] Fatigue Recovery per turn"
+						text = "[color=" + ::Const.UI.Color.NegativeValue + "]+2[/color] Fatigue at the start of every turn"
 					},
 					{
 						id = 10,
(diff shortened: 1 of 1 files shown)


scripts/skills/effects/rf_encumbrance_effect.nut
```diff
@@ -130,26 +130,46 @@ this.rf_encumbrance_effect <- ::inherit("scripts/skills/skill", {
...
+				fatigue = 2;
+				break;
+
+			default:
+				fatigue = 3;
+				break;
+		}
+
+		local actor = this.getContainer().getActor();
+		actor.setFatigue(::Math.min(actor.getFatigueMax(), actor.getFatigue() + fatigue));
```
Darxo, line 173: This would fail on brothers with less than 15 usable fatigue.
A brother with 10/10 Fatigue would recover 10 Fatigue due to vanilla mechanic leaving him at 0/10 Fatigue at the start of their turn.
If they have the highest Encumbrance Level he now gains a penalty of 3 fatigue leaving him at 3/10 Fatigue.

What you actually want is him being at 10/10 Fatigue because you reduce the 15 Fatigue per turn down to 12 Fatigue per turn.

LordMidas: Lol yeah I realized this when I was going to sleep last night. It will fail also when someone recovers enough Fatigue to get to 0 but had excess fatigue recovery. E.g. a bro has 10 fatigue at turn start and has level 2 encumbrance. Due to fatigue recovery he should still go down to 0 fatigue but encumbrance will add fatigue to him.

So yeah this solution doesn't work. The only solution I can think of is to hook the actor.nut function which is unfortunate.

So instead of doing that perhaps encumbrance should increase fatigue cost of skills instead of affecting fatigue recovery? The goal is to eliminate the fatigue neutral cheese builds. As long as we achieve that goal in a sensible way it should be fine.

LordMidas: @Darxo , @Uberbagel If we want a clean and compatible solution, it would be to change the Encumbrance mechanic to say "When encumbered you build x Fatigue at the start of every turn". It accomplishes the goal of removing fatigue neutral cheese builds, and is then consistent with all edge cases.

PR #64, opened 2023-02-16
fix: reduce ap cost of split and swing on warbrand and rhomphaia by …

Uberbagel:
…1 as intended


LordMidas (approved):
Change both the PR title and the commit type to `fix` instead of `feat` and then consider it approved.

PR #65, opened 2023-02-16
feat: streamline and simplify racial damage reductions

Darxo:
Vanilla does racial damage reductions and increases hard-coded. This also causes several mistakes where sometimes a specific attack was missing causing that type of damage to deal full damage.
Reforged already did a first rework pass on these racial effects by mimicing them via the DamageType system.

**Overview of the changes:**
This PR tries to simplify those values so they are more presentable to players and understandable for devs.
Burning Ground no longer has inherit mutlipliers for different entity-types. Instead it uses the MSU damage type system and leaves the multipliers to the individual entities.
All races that are resistance to piercing attacks now all have 
- a 66% (or 75% in case of Schrat) damage reduction against ranged piercing (arrows damage is buffed, a few other random attacks here and there nerfed)
- and 50% against melee piercing. (for Schrats this is actually a new addition)
Burning ground does less damage to Schrats now but firelances do more damage against them.

**Additional Thoughts:**

While it does make sense that some projectiles like arrows should deal less damage than bolts to the same target it isn't the job of us hard-coding those differences via the racial effects. That could be solved via some new flexible rule like using the armor pen value of those attacks as an additional damage multiplier for certain races.

But I argue that the streamlining of those ranged piercing attacks doesn't matter in the grand scheme of things. 
It doesn't matter whether arrows deal 10% or 33% damage. They are bad either way against that enemy type and that brother should have better brought a throwing axe to that battle or stayed in reserve.

### DIFF
--- a/mod_reforged/hooks/config/tactical_entity_common.nut
+++ b/mod_reforged/hooks/config/tactical_entity_common.nut
@@ -1,0 +1,31 @@
+// We overwrite the vanilla function to remove all damage multiplier and add the burning damage type
+::Const.Tactical.Common.onApplyFire = function( _tile, _entity )
+{
+    if (_entity.getCurrentProperties().IsImmuneToFire) return;
+
+    ::Tactical.spawnIconEffect("status_effect_116", _tile, ::Const.Tactical.Settings.SkillIconOffsetX, ::Const.Tactical.Settings.SkillIconOffsetY, ::Const.Tactical.Settings.SkillIconScale, ::Const.Tactical.Settings.SkillIconFadeInDuration, ::Const.Tactical.Settings.SkillIconStayDuration, ::Const.Tactical.Settings.SkillIconFadeOutDuration, ::Const.Tactical.Settings.SkillIconMovement);
+    local sounds = [
+        "sounds/combat/dlc6/status_on_fire_01.wav",
+        "sounds/combat/dlc6/status_on_fire_02.wav",
+        "sounds/combat/dlc6/status_on_fire_03.wav"
+    ];
+    ::Sound.play(sounds[this.Math.rand(0, sounds.len() - 1)], ::Const.Sound.Volume.Actor, _entity.getPos());
+
+    local damage = this.Math.rand(15, 30);
+    local hitInfo = clone ::Const.Tactical.HitInfo;
+    hitInfo.DamageRegular = damage;
+    hitInfo.DamageArmor = damage;
+    hitInfo.DamageDirect = 0.1;
+    hitInfo.DamageType = ::Const.Damage.DamageType.Burning;     // Uses damagetype system from MSU
+    hitInfo.BodyPart = ::Const.BodyPart.Body;
+    hitInfo.BodyDamageMult = 1.0;
+    hitInfo.FatalityChanceMult = 0.0;
+    hitInfo.Injuries = ::Const.Injury.Burning;
+    hitInfo.IsPlayingArmorSound = false;
+    _entity.onDamageReceived(_entity, null, hitInfo);
+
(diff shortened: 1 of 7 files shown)


LordMidas:
Overall nice work. I think the changes are mostly acceptable - we'll iron out the details over time. I also agree that we could take into account the Direct Damage of a piercing skill to tweak the damage reduction further. Let's see what @Uberbagel's opinions are.

Uberbagel:
For reference with this change burning ground and firearms will both deal 1.32 x damage to schrats if my calc. is correct. Previously it was 1.33 for firelance and 3.0 for burning ground.

I think these are good changes, but that's a significant difference and I would to know of any other cases like this of significant change.

mod_reforged/hooks/skills/items/firearms_resistance_skill.nut
```diff
@@ -0,0 +1,14 @@
+::mods_hookExactClass("skills/items/firearms_resistance_skill", function(o) {
+    // We overwrite the vanilla behavior of that skill completely while replicating its effect pretty future proof
+    o.onBeforeDamageReceived = function( _attacker, _skill, _hitInfo, _properties )
+	{
+        if (_skill.getItem().isWeaponType(::Const.Items.WeaponType.Firearm))    // This covers all kinds of firearms
+        {
+            _properties.DamageReceivedRegularMult *= 0.66;
+        }
+        else if (_hitInfo.DamageType == ::Const.Damage.DamageType.Burning)  // This covers burning ground and firelances
```
LordMidas, line 9: Just want to point out that in Reforged I've changed the `WeaponType` of Firelance to be `Spear` only. I've removed the `Firearm` type from it because it conflicts with many things e.g. weapon masteries and this system. The `firearm` skill on that weapon is just a one-off use, so I think Spear fits it better anyway.

mod_reforged/hooks/config/tactical_entity_common.nut
```diff
@@ -0,0 +1,31 @@
+// We overwrite the vanilla function to remove all damage multiplier and add the burning damage type
+::Const.Tactical.Common.onApplyFire = function( _tile, _entity )
```
LordMidas, line 2: This `onApplyFire` change should be a separate `fix` commit because this was an oversight in the original implementation where I converted the racial skills to use MSU damage type system.

LordMidas (changes requested):
The changes to the damage reductions are fine by me. I've requested a change to the commits structure of the PR.

Darxo:
> For reference with this change burning ground and firearms will both deal 1.32 x damage to schrats if my calc. is correct. Previously it was 1.33 for firelance and 3.0 for burning ground.
> 
> I think these are good changes, but that's a significant difference and I would to know of any other cases like this of significant change.

What are you referencing exactly? It is supposed to just be 2.0 from both burning ground and firelance.

mod_reforged/hooks/skills/items/firearms_resistance_skill.nut
Darxo: Ok, I never expected it to be firearm in the first place.
And I already account for burning damage which the ignite firelance skill applies exclusively.

So I don't see anything that needs to be resolved here

Uberbagel:
My previous comment was wrong. Uber approved!

mod_reforged/hooks/skills/items/firearms_resistance_skill.nut
LordMidas: Yeah, I just wanted to mention that in case you didn't know it. Nothing to Resolve.

LordMidas (approved):

PR #66, opened 2023-02-17
feat: set Type of most racial effects to be only 'SkillType.Racial'

Darxo:
There is currently no reason that some of those racial effects randomly had the type 'Perk'.
That only causes issues with our tooltips trying to display those.

In vanilla there is no reason that some of those racial effects randomly had the type 'StatusEffect'
Afaik this is only relevant if you want to display an Icon and Tooltip to the player. But most of those effects have IsHidden at true which will never display those anyways.
The only exception here is 'schrat_racial'. Here I don't replace the Type

Overwriting those types will also give a small performance boost as those racials will no longer be queried.
When we decide to add descriptions for those and show the to the player we will re-add the type 'StatusEffect'

### DIFF
--- a/mod_reforged/hooks/skills/racial/ghost_racial.nut
+++ b/mod_reforged/hooks/skills/racial/ghost_racial.nut
@@ -1,0 +1,9 @@
+::mods_hookExactClass("skills/racial/ghost_racial", function(o) {
+	local create = o.create;
+	o.create = function()
+	{
+		create();
+		if (this.isType(::Const.SkillType.Perk))
+			this.removeType(::Const.SkillType.Perk);	// This effect having the type 'Perk' serves no purpose and only causes issues in modding
+	}
+});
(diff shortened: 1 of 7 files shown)


LordMidas:
While this is a "sensible" change (I agree that the vanilla code is very random in this case), I'd like to ask what is the _need_ for doing this? Does this accomplish something useful in the context of Reforged? If not, then I'd personally avoid making such a change as it can potentially mess with things as other mods don't expect these to be changed. But if it accomplishes something important in Reforged, then we should make this change.

Darxo:
It is one possible solution to this bug reported in the discord:
We display any skill that has the type **Perk** in our tooltip with no exception. Fake perks break our tooltip.
image

This can of course be handled/corrected in other ways aswell.
I just took it as a nice excuse to do some vanilla refactoring on the side

LordMidas (commented):
These things in vanilla are rather _benign_ and I don't see the _need_ to fix them. We can instead add proper descriptions/names to these effects so that they show up properly in our tooltip. But I also don't mind these changes - so I'm neutral on this. Take a final opinion from @Enduriel and if he is ok with this then merge it.

Darxo:
This will be opened under an new title and with less extreme suggested changes

PR #68, opened 2023-02-17
feat: give Saplings the Schrat-Racial effect

Darxo:
In return this will extract the Sapling-Spawn effect out of the Schrat-Racial and into its own new effect that is only given to Schrats for now.

Screenshot on how it looks ingame:
image

### DIFF
--- a/scripts/skills/effects/rf_sapling_harvest_effect.nut
+++ b/scripts/skills/effects/rf_sapling_harvest_effect.nut
@@ -1,0 +1,49 @@
+this.rf_sapling_harvest_effect <- ::inherit("scripts/skills/skill", {
+	m = {
+		DamageThreshold = 0.1,
+		SpawnedUnit = "scripts/entity/tactical/enemies/schrat_small"
+	},
+	function create()
+	{
+		this.m.ID = "effects.rf_sapling_harvest";
+		this.m.Name = "Sapling Harvest";
+		this.m.Icon = "skills/terrain_icon_06.png";
+		this.m.SoundOnUse = [];
+		this.m.Type = this.Const.SkillType.StatusEffect;
+		this.m.Order = this.Const.SkillOrder.Last;
+		this.m.IsActive = false;
+		this.m.IsStacking = false;
+		this.m.IsHidden = false;
+	}
+
+	function getDescription()
+	{
+		return ::Reforged.Mod.Tooltips.parseString("Whenever this character receives " + ::MSU.Text.colorGreen((this.m.DamageThreshold * 100.0) + "%") + " or more damage to [Hitpoints|Concept.Hitpoints] from a single hit, spawn a Sapling on an adjacent tile, if possible.");
+	}
+
+	function onDamageReceived( _attacker, _damageHitpoints, _damageArmor )
+	{
+		local actor = this.getContainer().getActor();
+		if (_damageHitpoints < (this.m.DamageThreshold * actor.getHitpointsMax())) return;
+
+		local candidates = [];
+		local myTile = actor.getTile();
+
+		for (local i = 0; i < 6; i++)
+		{
+			if (!myTile.hasNextTile(i)) continue;
+
+			local nextTile = myTile.getNextTile(i);
+			if (nextTile.IsEmpty && ::Math.abs(myTile.Level - nextTile.Level) <= 1)
+			{
+				candidates.push(nextTile);
+			}
+		}
+		if (candidates.len() == 0) return;
+
+		local spawnTile = candidates[::Math.rand(0, candidates.len() - 1)];
+		local sapling = ::Tactical.spawnEntity(this.m.SpawnedUnit, spawnTile.Coords);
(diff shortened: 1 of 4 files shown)


LordMidas:
I must say I really like your current implementation - but I just want to ask your opinion on what would be better from a submodding / compatibility with mods point of view: Would it be better to add a Boolean inside the `schrat_racial` and wrap the spawning of the sapling inside an `if` which checks for that Boolean. Then we set that Boolean to `true` for the racial skill that is added to the big schrat but set it to false for the sapling.

Darxo:
The advance of your described approach would be if some mod added another Schrat-Like enemy that behaved so similar to the existing one that they don't create a new racial effect for them. They wouldn't be missing out on the Sapling effect because those mods can't know that that is now a unique effect.

But the backwards compatibility train kinda already left the station with Saplings now having the `schrat_racial` effect including all effects which mods may add in the future and which were not meant to be given to Saplings.

I want the flexibility of splitting vanilla spaghetti code skills into several to 
- better display the different parts to the player where it makes sense via icons/description
- allow us to freely combine them for new types of enemies
If it's up to me then this is another one of those total overhaul decisions that [sub]mods unfortunately have to be adjusted to

scripts/skills/effects/rf_sapling_harvest.nut
```diff
@@ -0,0 +1,45 @@
+this.rf_sapling_harvest <- ::inherit("scripts/skills/skill", {
```
LordMidas, line 1: Our convention is to append the word `_effect` to every status effect file that goes in the `skills/effects` folder. So please change the filename and the classname here to `rf_sapling_harvest_effect`.

scripts/skills/effects/rf_sapling_harvest.nut
```diff
@@ -0,0 +1,45 @@
...
+                local nextTile = myTile.getNextTile(i);
+                if (nextTile.IsEmpty && this.Math.abs(myTile.Level - nextTile.Level) <= 1)
+                {
+                    candidates.push(nextTile);
+                }
+			}
+			if (candidates.len() == 0) return;
+
+			local spawnTile = candidates[this.Math.rand(0, candidates.len() - 1)];
+			local sapling = this.Tactical.spawnEntity("scripts/entity/tactical/enemies/schrat_small", spawnTile.Coords);
```
LordMidas, line 40: `::Tactical`

scripts/skills/effects/rf_sapling_harvest.nut
```diff
@@ -0,0 +1,45 @@
...
+
+	function onDamageReceived( _attacker, _damageHitpoints, _damageArmor )
+	{
+		local actor = this.getContainer().getActor();
+		if (_damageHitpoints >= actor.getHitpointsMax() * this.m.DamageThreshold)
+		{
+			local candidates = [];
+			local myTile = actor.getTile();
+
+			for( local i = 0; i < 6; i = ++i )
```
LordMidas, line 27: Two changes for convention's sake:
1. spacing: `for (local i =0; i < 6; i++)`
2. The last part is just `++i` or `i++` and not `i = ++i`

scripts/skills/effects/rf_sapling_harvest.nut
```diff
@@ -0,0 +1,45 @@
...
+			{
+				if (!myTile.hasNextTile(i)) continue;
+
+                local nextTile = myTile.getNextTile(i);
+                if (nextTile.IsEmpty && this.Math.abs(myTile.Level - nextTile.Level) <= 1)
```
LordMidas, line 32: `::Math`

LordMidas (changes requested):
Ok - with that logic, I think it's ok to include this PR. Just some minor changes as requested in my review. And one major change: This should be 2 commits. The first commit should be the splitting of `schrat_racial` and `sapling_harvest` and addition of `sapling_harvest` to the big schrats. The second commit should be the addition of `schrat_racial` to the small schrats. 

Finally, we should get a final approval from @Enduriel before merging.

scripts/skills/effects/rf_sapling_harvest_effect.nut
```diff
@@ -0,0 +1,49 @@
...
+		this.m.Type = this.Const.SkillType.StatusEffect;
+		this.m.Order = this.Const.SkillOrder.Last;
+		this.m.IsActive = false;
+		this.m.IsStacking = false;
+		this.m.IsHidden = false;
+	}
+
+	function getDescription()
+	{
+		"Whenever this character receives " + ::MSU.Text.colorGreen((this.m.DamageThreshold * 100.0) + "%") + " or more hitpoint damage from a single hit, spawn a Sapling on an adjacent tile, if possible.";
```
LordMidas, line 21: replace `or more hit point damage` with `or more damage to Hitpoints` to be more in line with vanilla style.

LordMidas (approved):
That minor language change, and then approved from my side.

PR #69, opened 2023-02-19
Weapon tier

Uberbagel:
Heavy and goblin crossbow reload now costs 6 AP but will be reduced by 2 AP by weapon mastery. Reduced fatigue costs on crossbows to reasonable, sensible and usable levels. Reduced extra AP cost of warbow aimed shot. Reduced woodcutters axe round swing fatigue cost to the proper level (wasn't reduced before when it should've been).


LordMidas:
Not necessary anymore with the elaborate weapon costs rework we've recently implemented.

PR #70, opened 2023-02-20
feat: change bonus granted by smoke effect from percentage based to flat

Darxo:


mod_reforged/hooks/skills/effects/smoke_effect.nut
```diff
@@ -0,0 +1,29 @@
+::mods_hookExactClass("skills/effects/smoke_effect", function (o) {
+    o.m.RangedDefenseBonus = 30;
```
LordMidas, line 2: `<-`

mod_reforged/hooks/skills/effects/smoke_effect.nut
```diff
@@ -0,0 +1,29 @@
...
+
+    local getTooltip = o.getTooltip;
+	o.getTooltip = function()
+	{
+        local ret = getTooltip();
+        foreach (entry in ret)
+        {
+            if (entry.id == 12 && entry.icon == "ui/icons/ranged_defense.png")
+            {
+                // We replace a single line of the vanilal tooltip without touching the rest
```
LordMidas, line 12: typo in comment: `vanilla`

mod_reforged/hooks/skills/effects/smoke_effect.nut
```diff
@@ -0,0 +1,29 @@
...
+            }
+        }
+        return ret;
+	}
+
+    local onUpdate = o.onUpdate;
+    o.onUpdate = function( _properties )
+    {
+        onUpdate(_properties);
+        if (this.m.IsGarbage == true) return;     // Simple way of checking whether the line above skipped the buffs because the smoke effect is not active anymore
```
LordMidas, line 24: Prefer to use `this.isGarbage()` instead of `this.m.IsGarbage`.

mod_reforged/hooks/skills/effects/smoke_effect.nut
```diff
@@ -0,0 +1,29 @@
...
+
+        _properties.RangedDefenseMult *= 0.5;   // Hard-Coded reversing of the vanilla buffs. They multiply it with 2.0
```
LordMidas, line 26: Instead of this reversal method, perhaps overwrite it completely? What if a mod hooks it to change `RangedSkillMult` by 3.0 during this function. Then we are not fully reversing it anyway - so I don't think we have to care about mod compatibility here as we are completely changing how this effect behaves.

LordMidas (changes requested):

mod_reforged/hooks/skills/effects/smoke_effect.nut
Darxo: But replacing that function doesn't fix the issue you are describing if that mod loads after our mod. Then their adjustments still take place and warp the effects of smoke bomb to an undesired level.

Going all-in into hooking here at least provides compatibility with most other adjustments that a modder or vanilla could do to the smoke-bomb effect

Darxo: I replaced that line with a better solution now. 
I don't reverse the effects but instead just reset that property to the value it was before the vanilla function was called.

LordMidas: I think overwriting the function is better in this case than resetting the property like that. Because if you want to be incompatible with a mod's changes, better make it explicit. Otherwise this can lead to partial changes being overwritten and partial not - which is hard to track down in the code's logic and catch with multiple mods. @Enduriel please opinionate.

mod_reforged/hooks/skills/actives/throw_smoke_bomb_skill.nut
```diff
@@ -0,0 +1,31 @@
...
+			ret.remove(index);	// We remove that entry because we want to split it into two seperate entries that are next to each other
+			break;
+		}
+
+		ret.extend([
+			{
+				id = 10,
+				type = "text",
+				icon = "ui/icons/ranged_skill.png",
+				text = "Reduces Ranged Skill by " + ::MSU.Text.colorRed("50%") + " for anyone inside"
```
LordMidas, line 19: Reword it as: 50% reduced Ranged Skill

mod_reforged/hooks/skills/actives/throw_smoke_bomb_skill.nut
```diff
@@ -0,0 +1,31 @@
...
+			},
+			{
+				id = 11,
+				type = "text",
+				icon = "ui/icons/ranged_defense.png",
+				text = "Increases Ranged Defense by " + ::MSU.Text.colorGreen("30") + " for anyone inside"
```
LordMidas, line 25: Reword it as: +30 Ranged Defense

LordMidas, line 25: There is no need to mention `for anyone inside` because the effect only shows up on characters who are inside the smoke.

LordMidas (changes requested):
Alright, based on our discussion in Discord, I am ok with the complete reverting of changes to `RangedDefenseMult` in our hook on `onUpdate`. I've requested some changes to the language of the tooltip, after which I'll be ok for this to merge.

mod_reforged/hooks/skills/effects/smoke_effect.nut
LordMidas: Based on our discussion in Discord, I am ok with the complete reverting of changes to RangedDefenseMult in our hook on onUpdate.

mod_reforged/hooks/skills/actives/throw_smoke_bomb_skill.nut
```diff
@@ -0,0 +1,31 @@
+::mods_hookExactClass("skills/actives/throw_smoke_bomb_skill", function(o) {
+	local getTooltip = o.getTooltip;
+	o.getTooltip = function()
+	{
+		local ret = getTooltip();
+        foreach (index, entry in ret)
+        {
+            if (entry.id != 5) continue;
+			if (entry.text.find("Increases Ranged Defense by ") == null) continue;	// Vanilla has two entries with id 5. That's why we check for a phrase aswell
```
LordMidas, line 9:
```squirrel
if (entry.id != 5 || entry.text.find("Ranged Defense") == null)
    continue;
```

In fact for this such a small case the readability is way better if you do:
```squirrel
if (entry.id == 5 && entry.text.find("Ranged Defense") != null)
{
    ret.remove(index);
    break;
}
```
Either way, please combine both conditions into one `if` block.

LordMidas, line 9: Vanilla has the text `Ranged Defense` not `Increases Ranged Defense by`, so you should check for the former.

mod_reforged/hooks/skills/actives/throw_smoke_bomb_skill.nut
Darxo: We are inside an active ability here: `throw_smoke_bomb_skill`. I find it very reasonable if we clarify who exactly gains the bonus if that bonus is mentioned in a separate line.
Otherwise one might think that the thrower gains this bonus.

I think you confused this script with the `smoke_effect`

mod_reforged/hooks/skills/actives/throw_smoke_bomb_skill.nut
Darxo: I wouldn't say `way better`
But I'm ok changing it in this case

mod_reforged/hooks/skills/actives/throw_smoke_bomb_skill.nut
LordMidas: Ah right. Yes, sorry I confused it.

LordMidas (approved):

PR #72, opened 2023-02-21
refactor: set action points to 7 but reduce AP cost of movement

LordMidas:
With this they can move 7 tiles on normal ground (compared to 6 in vanilla). They can attack twice with Falchion and then Bite, or attack thrice with Spear and then Bite.

Currently they have 13 AP which means that with Falchion they can attack 4 times, and then Bite (so 5 total attacks) and with Spear they can attack 6 times and then Bite (so 7 total attacks). This is broken. I believe the vanilla devs gave them 13 AP to allow them to have a large movement - and not so many attacks. But in Reforged because we have AP cost reduction on goblin weapons and then from the Two for One perk which further reduces it - it allows them to attack so much.

After the changes in this PR, their attacks are normal in number, but they still get to move roughly similar number of tiles as vanilla.

### DIFF
--- a/mod_reforged/hooks/entity/tactical/enemies/goblin_wolfrider.nut
+++ b/mod_reforged/hooks/entity/tactical/enemies/goblin_wolfrider.nut
@@ -53,6 +53,8 @@
 		this.m.Skills.add(this.new("scripts/skills/perks/perk_backstabber"));
 
 		// Reforged
+		this.m.BaseProperties.ActionPoints = 7;
+		this.m.BaseProperties.MovementAPCostAdditional -= 1;
 		this.m.Skills.add(::new("scripts/skills/perks/perk_rf_vigorous_assault"));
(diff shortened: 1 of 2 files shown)


Uberbagel:
->After the changes in this PR, their attacks are normal in number, but they still get to move roughly similar number of tiles as vanilla.

They aren't normal in number compared to other goblins though. Skirmishers can attack 3 times with goblin falchion (intended) and 4 times with goblin spear (unintended imo). I think wolfriders should have the same number of 'goblin' attacks as skirmishers and the bite should be one extra attack on top of that.

Your numbers aren't completely making sense to me. Wolfriders have 1 additional action point cost that we don't remove in Reforged. Should that not increase the AP cost of falchion to 4, giving wolfriders 3 falchion attacks and one free wolf bite attack after that?

I do think we should remove two for one from non-miniboss goblins though, as 2 ap attacks from goblins with spears is not what we want imo.

PR #75, opened 2023-02-22
refactor: remove the Type 'Perk' from all racial effects

Darxo:
There is currently no reason that some of those racial effects randomly had the type 'Perk'. This currently causes issues with our tooltips trying to display those. Also it will cause the Oblivion Potion to remove those racial effects if they are ever to appear on "brothers".

**Important:** This PR requires MSU functions which are not yet in MSU. Enduriel said that he wants to get them in ASAP


LordMidas (approved):

LordMidas:
Commit should be type `refactor: ` and not `fix: `.

PR #76, opened 2023-02-22
fix: lower volume of annoying kid ambience sound

Darxo:
Out of all ambience sound files this is the most annoying one. When it randomly plays twice close to each other it is the most recognizable one. I think most of that is because this sounds volume is above average.


LordMidas (commented):
I've never been bothered by any ambient sound in BB so far. So, in my opinion, don't include this in Reforged. It could be a separate sub-mod.

LordMidas:
I remember we had a small talk about this on the dev Discord and the team wasn't in favor of including this change in Reforged. So I'll close the PR.

PR #78, opened 2023-02-23
rewrite vampire life leech

Darxo:
feat: add new life-leech effect

refactor: replace life-leech effect from racial with new separate effect
As a result this also fixes the bug where necrosavants were leeching from everything, regardless of whether it bleeds

image

### DIFF
--- a/mod_reforged/hooks/skills/racial/vampire_racial.nut
+++ b/mod_reforged/hooks/skills/racial/vampire_racial.nut
@@ -6,24 +6,10 @@
 		this.m.Name = "Vampire";
 		this.m.Description = "";	// Vanilla has "TODO" written here. We don't want that to display
 		this.m.Icon = "/ui/orientation/vampire_01_orientation.png";
-		this.m.IsHidden = false;
+		this.m.IsHidden = true;		// We currently don't have any description to display
 		this.addType(::Const.SkillType.StatusEffect);	// We now want this effect to show up on the enemies
 		if (this.isType(::Const.SkillType.Perk))
 			this.removeType(::Const.SkillType.Perk);	// This effect having the type 'Perk' serves no purpose and only causes issues in modding
-	}
-
-	o.getTooltip <- function()
-	{
-		local ret = this.skill.getTooltip();
-		ret.extend([
-			{
-				id = 10,
-				type = "text",
-                icon = "ui/icons/regular_damage.png",
-				text = "Heal " + ::MSU.Text.colorGreen("100%") + " of Hitpoint damage inflicted on enemies"
-			}
-		]);
-		return ret;
 	}
 
 	o.onAdded <- function()
@@ -33,5 +19,10 @@
 		baseProperties.IsAffectedByInjuries = false;
 		baseProperties.IsAffectedByNight = false;
 		baseProperties.IsImmuneToPoison = true;
+
+		this.getContainer().add(::new("scripts/skills/effects/rf_life_leech_effect"));
 	}
+
+	// Vanilla used to handle life-leech in this function but we outsourced that effect. That's why we need to disable the vanilla function now by overwriting it
+	o.onTargetHit = function( _skill, _targetEntity, _bodyPart, _damageInflictedHitpoints, _damageInflictedArmor ) {}
 });
--- a/scripts/skills/effects/rf_life_leech_effect.nut
+++ b/scripts/skills/effects/rf_life_leech_effect.nut
@@ -4,6 +4,7 @@ this.rf_life_leech_effect <- ::inherit("scripts/skills/skill", {
         LeechDistance = 1,	// 1 only allows leeching from adjacent enemies
 		ForceHeal = false   // Temporary helper variable
 	},
+
 	function create()
 	{
 		this.m.ID = "effects.rf_life_leech";
@@ -50,7 +51,7 @@ this.rf_life_leech_effect <- ::inherit("scripts/skills/skill", {
 
 	function onTargetHit( _skill, _targetEntity, _bodyPart, _damageInflictedHitpoints, _damageInflictedArmor )
 	{
-		if (_damageInflictedHitpoints <= 0) return;
+		if (_damageInflictedHitpoints == 0) return;
 		if (this.m.ForceHeal)      // If our attack killed the target we instead handle the leeching in here
 		{
 			this.leechLife(_damageInflictedHitpoints);
--- a/scripts/skills/effects/rf_life_leech_effect.nut
+++ b/scripts/skills/effects/rf_life_leech_effect.nut
@@ -1,0 +1,98 @@
+this.rf_life_leech_effect <- ::inherit("scripts/skills/skill", {
+	m = {
+		LifeLeechValue = 1.0,   // 1.0 equals to 100% damage leeched as life
+        LeechDistance = 1,	// 1 only allows leeching from adjacent enemies
+		ForceHeal = false   // Temporary helper variable
+	},
+	function create()
+	{
+		this.m.ID = "effects.rf_life_leech";
+		this.m.Name = "Life Leech";
+		this.m.Description = "This character recovers Hitpoints when dealing damage to enemies with blood.";
+		this.m.Icon = "skills/status_effect_09.png";
+		this.m.Overlay = "status_effect_09";
+		this.m.SoundOnUse = [
+			"sounds/enemies/vampire_life_drain_01.wav",
+			"sounds/enemies/vampire_life_drain_02.wav",
+			"sounds/enemies/vampire_life_drain_03.wav"
+		];
+		this.m.Type = ::Const.SkillType.StatusEffect;
+		this.m.Order = this.Const.SkillOrder.Perk;
+		this.m.IsActive = false;
+		this.m.IsRemovedAfterBattle = false;
+		this.m.IsStacking = false;
+	}
+
+	function getTooltip()
+	{
+		local ret = this.skill.getTooltip();
+
+		local withinText = this.m.LeechDistance == 1 ? "adjacent enemies" : "enemies within " + ::MSU.Text.colorGreen(this.m.LeechDistance) + " tiles";
+        ret.push({
+            id = 11,
+            type = "text",
+            icon = "ui/icons/health.png",
+            text = "Heal " + ::MSU.Text.colorGreen((this.m.LifeLeechValue * 100) + "%") + " of Hitpoint damage inflicted on " + withinText + " that have blood"
+        });
+
+		return ret;
+	}
+
+	function onCombatStarted()
+	{
+		this.m.ForceHeal = false;
+	}
+
+	function onCombatFinished()
+	{
+		this.m.ForceHeal = false;
+	}
+
+	function onTargetHit( _skill, _targetEntity, _bodyPart, _damageInflictedHitpoints, _damageInflictedArmor )
+	{
+		if (_damageInflictedHitpoints <= 0) return;
+		if (this.m.ForceHeal)      // If our attack killed the target we instead handle the leeching in here
+		{
+			this.leechLife(_damageInflictedHitpoints);
+			this.m.ForceHeal = false;
+			return;
+		}
+		local actor = this.getContainer().getActor();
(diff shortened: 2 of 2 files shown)


scripts/skills/effects/rf_life_leech_effect.nut
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	function onCombatFinished()
+	{
+		this.m.ForceHeal = false;
+	}
+
+	function onTargetHit( _skill, _targetEntity, _bodyPart, _damageInflictedHitpoints, _damageInflictedArmor )
+	{
+		if (_damageInflictedHitpoints <= 0) return;
```
LordMidas, line 55: If the damage is negative, it seems like a bug. Therefore, it is preferable to use `==` here to show the intent clearly.

scripts/skills/effects/rf_life_leech_effect.nut
```diff
@@ -0,0 +1,100 @@
...
+		if (this.m.ForceHeal)      // If our attack killed the target we instead handle the leeching in here
+		{
+			this.lifesteal(_damageInflictedHitpoints);
+			this.m.ForceHeal = false;
+			return;
+		}
+		local actor = this.m.Container.getActor();
```
LordMidas, line 62: Prefer to use `getContainer()`

scripts/skills/effects/rf_life_leech_effect.nut
```diff
@@ -0,0 +1,100 @@
...
+		if (actor.getHitpoints() == actor.getHitpointsMax()) return;
+		if (_targetEntity == null || !_targetEntity.isAlive()) return;      // Otherwise we crash the game when leeching from a killing blow
+		if (_targetEntity.getCurrentProperties().IsImmuneToBleeding) return;
+		if (actor.getTile().getDistanceTo(_targetEntity.getTile()) > this.m.LeechDistance) return;
```
LordMidas, line 66: Please combine all these `if` checks into one `if` check using `||` operator.

scripts/skills/effects/rf_life_leech_effect.nut
```diff
@@ -0,0 +1,100 @@
...
+
+		this.lifesteal(_damageInflictedHitpoints);
+	}
+
+	function onTargetKilled( _targetEntity, _skill )
+	{
+		if (_targetEntity.getCurrentProperties().IsImmuneToBleeding) return;
+		local actor = this.m.Container.getActor();
```
LordMidas, line 74: Prefer to use `getContainer()`

scripts/skills/effects/rf_life_leech_effect.nut
```diff
@@ -0,0 +1,100 @@
...
+		if (actor.getHitpoints() == actor.getHitpointsMax()) return;
+		if (actor.getTile().getDistanceTo(_targetEntity.getTile()) > this.m.LeechDistance) return;
```
LordMidas, line 76: Combine the checks into one `if` with `||` operator

scripts/skills/effects/rf_life_leech_effect.nut
```diff
@@ -0,0 +1,100 @@
...
+		this.m.ForceHeal = true;
+	}
+
+	function lifesteal( _damageInflictedHitpoints )
+	{
+		local actor = this.m.Container.getActor();
```
LordMidas, line 82: Prefer to use `getContainer()`

scripts/skills/effects/rf_life_leech_effect.nut
```diff
@@ -0,0 +1,100 @@
...
+        LeechDistance = 1,
+		ForceHeal = false   // Temporary helper variable
+	},
+	function create()
+	{
+		this.m.ID = "effects.rf_life_leech";
+		this.m.Name = "Life Leech";
+		this.m.Description = "This character recovers Hitpoints when dealing damage to enemies with blood.";
+		this.m.Icon = "skills/status_effect_09.png";
+		this.m.IconMini = "";   // We currently have no fitting mini-icon for this
```
LordMidas, line 13: I can create one, using the bigger icon

scripts/skills/effects/rf_life_leech_effect.nut
```diff
@@ -0,0 +1,100 @@
...
+		this.m.IsActive = false;
+		this.m.IsRemovedAfterBattle = false;
+		this.m.IsStacking = false;
+	}
+
+	function getTooltip()
+	{
+		local ret = this.skill.getTooltip();
+
+        local withinText = "adjacent enemies";
```
LordMidas, line 31: Seems like a perfect candidate for a ternary use.
```squirrel
local withinText = this.m.LeechDistance == 1 ? "adjacent enemies" : "enemies within " + ::MSU.Text.colorGreen(this.m.LeechDistance) + " tiles";
```

scripts/skills/effects/rf_life_leech_effect.nut
```diff
@@ -0,0 +1,100 @@
...
+	function onTargetKilled( _targetEntity, _skill )
+	{
+		if (_targetEntity.getCurrentProperties().IsImmuneToBleeding) return;
+		local actor = this.m.Container.getActor();
+		if (actor.getHitpoints() == actor.getHitpointsMax()) return;
+		if (actor.getTile().getDistanceTo(_targetEntity.getTile()) > this.m.LeechDistance) return;
+		this.m.ForceHeal = true;
+	}
+
+	function lifesteal( _damageInflictedHitpoints )
```
LordMidas, line 80: Function names that do something should generally be clear that they are verbs. So, I would prefer a function name like `doLifesteal` or `stealLife` or `heal`

LordMidas, line 80: Please rename the variable to something more generic e.g. `_amount`, because this function doesn't need to know if it was "damage" or anything else.

scripts/skills/effects/rf_life_leech_effect.nut
```diff
@@ -0,0 +1,100 @@
...
+		if (_damageInflictedHitpoints <= 0) return;
+		if (this.m.ForceHeal)      // If our attack killed the target we instead handle the leeching in here
+		{
+			this.lifesteal(_damageInflictedHitpoints);
+			this.m.ForceHeal = false;
+			return;
+		}
+		local actor = this.m.Container.getActor();
+		if (actor.getHitpoints() == actor.getHitpointsMax()) return;
+		if (_targetEntity == null || !_targetEntity.isAlive()) return;      // Otherwise we crash the game when leeching from a killing blow
```
LordMidas, line 64: `_targetEntity` can never be null in this function, so you can skip the null check.

LordMidas (changes requested):

scripts/skills/effects/rf_life_leech_effect.nut
Darxo: But I also argue that this does not need a mini-icon.
Mini-Icons should be used sparingly to not overload the UI.

Especially effects which are always active like Life-Leech should therefor be the first to **not** be using mini-icons

scripts/skills/effects/rf_life_leech_effect.nut
Darxo: I find those single line guard clauses much better coding practice in most cases than using long if statements or god forbid multiple nested if's that cause all the code to indent to the right outside of the screen.

Vanilla does it the same way btw. Except that they use 5 lines for each if:
image

scripts/skills/effects/rf_life_leech_effect.nut
Darxo: I agree that negative damage dealt should be considered a bug.

I was copying vanilla in this case. In the end I didn't write the vanilla code and don't know for sure whether negative values are possible.

scripts/skills/effects/rf_life_leech_effect.nut
LordMidas: Long `if` clauses can be broken into multiple lines and they can be very easy to read. For example:
```squirrel
if (_script == "items/weapons/greenskins/goblin_bow" ||
    _script == "items/weapons/greenskins/goblin_crossbow" ||
    _script == "items/weapons/greenskins/goblin_heavy_bow" ||
    _script == "items/weapons/named/named_goblin_heavy_bow"
)
{
	::mods_hookNewObject(_script, function(o) {
		o.m.FatigueOnSkillUse -= 2;
	});
}
```

scripts/skills/effects/rf_life_leech_effect.nut
LordMidas: Sure, yes that is fine.

scripts/skills/effects/rf_life_leech_effect.nut
LordMidas: As this is primarily a code style discussion, we should see what @Enduriel says.

scripts/skills/effects/rf_life_leech_effect.nut
Darxo: Alright.
But for the protocol, vanilla also did it this way. I didn't put too much thought into it and just copied that funciton

Darxo: 95%+ of Lifeleech effects in other games scale from some sort of damage dealt. I think it's widely known and understood to function this way.
I will change this variable to just mean `_damageInflicted` because this may allow leeching from damage inflicted to armor or fatigue.

With something generic like `_amount` it isn't certain for me anymore what kind of value is expected to arrive into this function.

LordMidas: Ok that is fine.

scripts/skills/effects/rf_life_leech_effect.nut
LordMidas: It can never be negative, in vanilla. actor.nut `onDamageReceived` function:
```squirrel
damage = this.Math.max(0, this.Math.max(this.Math.round(damage), this.Math.min(this.Math.round(_hitInfo.DamageMinimum), this.Math.round(_hitInfo.DamageMinimum * p.DamageReceivedTotalMult))));
_hitInfo.DamageInflictedHitpoints = damage;
```
which is then passed to the `onTargetHit` skill event in `skill.nut` function `onScheduledTargetHit`:
```squirrel
_info.Container.onTargetHit(_info.Skill, _info.TargetEntity, hitInfo.BodyPart, hitInfo.DamageInflictedHitpoints, hitInfo.DamageInflictedArmor);
```

scripts/skills/effects/rf_life_leech_effect.nut
Enduriel: I strongly prefer this format
```squirrel
if (_script == "items/weapons/greenskins/goblin_bow" ||
    _script == "items/weapons/greenskins/goblin_crossbow" ||
    _script == "items/weapons/greenskins/goblin_heavy_bow" ||
    _script == "items/weapons/named/named_goblin_heavy_bow"
)
```
It achieves both purposes of having a new line for each statement while also minimizing wasted spaces. One thing to keep in mind is to make sure you continue to use only tabs for this alignment.

Darxo: I applied this way of formatting now.

But the reasons you give right here are not convincing. My approach had a new line for each statement & minimized wasted space. Even more so that this suggested approach here.
You lose a whole line for the `)` and two more lines for `{` and `}`.

The argumentation should instead have been readability or staying with a style that most are probably already used to

mod_reforged/hooks/skills/racial/vampire_racial.nut
```diff
@@ -6,32 +6,23 @@
...
-
 	o.onAdded <- function()
 	{
 		local baseProperties = this.getContainer().getActor().getBaseProperties();
 
 		baseProperties.IsAffectedByInjuries = false;
 		baseProperties.IsAffectedByNight = false;
 		baseProperties.IsImmuneToPoison = true;
+
+		this.getContainer().add(this.new("scripts/skills/effects/rf_life_leech_effect"));
```
LordMidas, line 23: prefer `::new`

LordMidas (approved):

PR #79, opened 2023-02-24
feat: rework several vanilla getHitFactors

Darxo:


mod_reforged/hooks/skills/skill.nut
```diff
@@ -5,4 +5,57 @@
...
+						ret[index].text = ::MSU.Text.colorRed(this.m.HitChanceBonus + "% ") + ret[index].text;
+					}
+					else
+					{
+						ret[index].text = ::MSU.Text.colorGreen(this.m.HitChanceBonus + "% ") + ret[index].text;
+					}
+					break;
+
+				case "Height disadvantage":
+					ret[index].text = ::MSU.Text.colorRed((::Const.Combat.LevelDifferenceToHitMalus * (this.m.Container.getActor().getTile().Level - _targetTile.Level)) + "% ") + ret[index].text;
```
LordMidas, line 51: You probably need to multiply by -1 to remove the negative sign here (negative sign should be removed because the icon is already showing a negative sign).

mod_reforged/hooks/skills/skill.nut
```diff
@@ -5,4 +5,57 @@
...
+				case "Resistance against ranged weapons":	// This was never followed through for anything other than skeletons
+				case "Resistance against piercing attacks":	// This was never followed through for anything other than skeletons
+				case "Nighttime":
+					ret.remove(index);
+					break;
+
+				// We add additional hit-chance numbers for some other entries
+				case "Too close":
+				case this.getName():
+					if (this.m.HitChanceBonus == 0) break;	// MSU pushes a raw 'this.getName()' string sometimes even when HitChanceBonus is 0. Currently this is with ranged attacks based on AdditionalAccuracy
```
LordMidas, line 39: This part is a bit fragile because we have skills e.g. Pattern Recognition which show up in hit factors with their name. IF such skills would have a non-zero `this.m.HitChanceBonus` then their entry would get messed up here. But that is probably a very unlikely case to happen in practice. Just something to be aware of.

mod_reforged/hooks/skills/skill.nut
```diff
@@ -5,4 +5,57 @@
...
+				// For these entries we switch out the Icon for a better one
+				case "Immune to stun":
+				case "Immune to being rooted":
+				case "Immune to being disarmed":
+				case "Immune to being knocked back or hooked":
+				ret[index].icon = "ui/tooltips/warning.png";	// These cases should be a warnings and not hide in-between hit-chance lines
+					break;
+
+				// We remove these entries from ever displaying because they are mostly redundant and take away space
+				case "Fast Adaption":		// This should be handled by the perk itself if it really feels like this information is important
```
LordMidas, line 26: Anything that affects directly affects hit-chance should show up in this tooltip. Therefore, Fast Adaptation and Nighttime specifically are important in my opinion. Nighttime especially as it is something that is negated by a certain items in the game (and maybe more from mods) - showing when it is applying and when not is relevant information.

Darxo: Nighttime is already listed under effects and it is directly affecting your attributes in real time which you can also see in the tooltip. That is my cut-off point for this PR, that's where I chose to draw the line whether to remove a vanilla entry.

Fast Adaption should be shown, but by the perk itself and not by the `skill.nut` class in a hardcoded way.

mod_reforged/hooks/skills/skill.nut
Darxo: That will not be an issue here. I only check for entries which are **just** a name but Pattern Recognition will automatically display the exact percentage when it appears

mod_reforged/hooks/config/character.nut
```diff
@@ -36,3 +36,11 @@ local getClone = ::Const.CharacterProperties.getClone;
 ::Const.ProjectileSprite.push("rf_projectile_flaming_arrow");
 
 ::Const.Movement.AutoEndTurnBelowAP = 1;
+
+::Const.CharacterProperties.getSurroundedBonus <- function( _targetEntity )
```
LordMidas, line 40: This should be a function in actor, not properties. Then in the function in the actor, you use the values from the character's CurrentProperties.

mod_reforged/hooks/skills/perks/perk_fast_adaption.nut
```diff
@@ -0,0 +1,13 @@
+::mods_hookExactClass("skills/perks/perk_fast_adaption", function (o) {
+
```
LordMidas, line 2: Please remove the empty line

mod_reforged/hooks/skills/skill.nut
```diff
@@ -5,4 +5,77 @@
...
+						local bonus = this.getContainer().getActor().getCurrentProperties().getSurroundedBonus(_targetTile.getEntity());
+						ret[index].text = ::MSU.Text.colorGreen(bonus + "%") + " Surrounded";
+					}
+					break;
+
+			}
+		}
+
+		// New Entries
+		if (!this.m.IsShieldRelevant && _targetTile.IsOccupiedByActor && _targetTile.getEntity().isArmedWithShield())
```
LordMidas, line 69: I like this addition. Perhaps also include a `IsShieldwallRelevant` entry (even though no vanilla skill uses that currently)?

mod_reforged/hooks/skills/skill.nut
```diff
@@ -5,4 +5,77 @@
...
+		for (local index = (ret.len() - 1); index >= 0; index--)	// We traverse it in reverse because we want to remove entries during the loop
+		{
+			switch (ret[index].text)
+			{
+				// For these entries we switch out the Icon for a better one
+				case "Immune to stun":
+				case "Immune to being rooted":
+				case "Immune to being disarmed":
+				case "Immune to being knocked back or hooked":
+				ret[index].icon = "ui/tooltips/warning.png";	// These cases should be a warnings and not hide in-between hit-chance lines
```
LordMidas, line 22: This is missing a tab indentation.

LordMidas (changes requested):

PR #81, opened 2023-02-26
feat: introduce new rumors for undiscovered unique locations

Darxo:


LordMidas (commented):
This is rather low impact flavor and I don't have strong opinions on this. So, I am going to defer to Taro/Uber for design review and Enduriel for code review.

mod_reforged/hooks/entity/world/location.nut
```diff
@@ -5,4 +5,12 @@
...
\ No newline at end of file
+
+	local onInit = o.onInit;
+	o.onInit = function()
+	{
+		onInit();
+		// Quick way to reduce the visibility of all unique locations by 30%. This does not affect the tundra elk location because it will be set to 0.8 at a later point
+		if (this.isLocationType(::Const.World.LocationType.Unique)) this.m.VisibilityMult *= 0.7;
+	}
+});
```
TaroEld, line 16: Why are you doing this?

Darxo: I don't want legendary locations to be easier to find so this nerf should go alongside the buff from the new rumors.

Combing the whole world should not be the main strategy for finding all legendary locations. 
You will still be able to find them by accident but that's now probably twice as hard with this nerf.

Instead if is now roughly split between you finding half the locations by accident and the other half by following clues from rumors. And especially the last legendary locations will be much more deterministic to find because all legendary rumors will point to them.

Darxo: Now revisiting this my method here may be flawed.
I assumed that the onInit function is only ever called once for every location ever: right after the `create()`

But if onInit is called every time that they are loaded from a savegame then this will continuously decrease their visionMult every time you save and load

Darxo: I removed this vision penalty now for this PR.

mod_reforged/hooks/entity/world/settlements/buildings/tavern_building.nut
```diff
@@ -0,0 +1,48 @@
...
+            if (s.isDiscovered()) continue;     // We don't want to show the player already discovered unique locations
+            if (s.getVisibilityMult() == 0.0) continue;     // These legendary locations are not meant to be found yet (tundra_elk_location)
+
+            local d = s.getTile().getDistanceTo(this.m.Settlement.getTile()) - ::Math.rand(1, 10);  // Small variance so we don't always talk about the nearest unique location
+            if (d > bestDist) continue;
+
+            bestDist = d;
+            bestLocation = s;
+        }
+        if (bestLocation == null) return ret;	// This should only happen if all legendary locations on the map are discovered
```
TaroEld, line 31: Probably just put this in a `getLegendaryLocationForRumor` function.

mod_reforged/hooks/entity/world/settlements/buildings/tavern_building.nut
```diff
@@ -0,0 +1,48 @@
...
+        this.m.Location = ::WeakTableRef(bestLocation);
+        local candidates = ::Const.Strings.RumorsUniqueLocation[::Math.rand(0, 1)];     // 50% of the time you get a generic tip with no name of the location
+
+        local rumor = "";
+        if (_isPaidFor) rumor = rumor + ::Const.Strings.PayTavernRumorsIntro[::Math.rand(0, ::Const.Strings.PayTavernRumorsIntro.len() - 1)];
+        else  rumor = rumor + "The patrons talk about this and that.";
+
+        rumor = rumor + "\n\n[color=#bcad8c]\"";
+        rumor = rumor + candidates[::Math.rand(0, candidates.len() - 1)];
+        rumor = rumor + "\"[/color]\n\n";
```
TaroEld, line 42: Same here, `+=` and `::MSU.Array.rand()`

mod_reforged/hooks/entity/world/settlements/buildings/tavern_building.nut
```diff
@@ -0,0 +1,48 @@
...
+            bestLocation = s;
+        }
+        if (bestLocation == null) return ret;	// This should only happen if all legendary locations on the map are discovered
+
+        this.m.Location = ::WeakTableRef(bestLocation);
+        local candidates = ::Const.Strings.RumorsUniqueLocation[::Math.rand(0, 1)];     // 50% of the time you get a generic tip with no name of the location
+
+        local rumor = "";
+        if (_isPaidFor) rumor = rumor + ::Const.Strings.PayTavernRumorsIntro[::Math.rand(0, ::Const.Strings.PayTavernRumorsIntro.len() - 1)];
+        else  rumor = rumor + "The patrons talk about this and that.";
```
TaroEld, line 38: += instead of rumor = rumor + ...
Alternatively, local rumor = _isPaidFor ? ::MSU.Array.rand(::Const.Strings.PayTavernRumorsIntro) : "The patrons talk about this and that."
Use MSU.Array.rand regardless I think

mod_reforged/hooks/config/rumors.nut
```diff
@@ -0,0 +1,68 @@
...
+        "There is an {urban | old} {legend | story} about a {place | location} called %location%. {Some | Some folk} {say | insist} it\'s {%distance% while others estimate it %wrongDistance% | %wrongDistance% while others estimate it %distance%}.",
+        "Just {the other day | yesterday | this morning} a drunk guy {was talking | came up to} to me, said he {found | stumbled upon} {a | some} {place | location} called %location% on one of his adventures. First he claimed it was {%direction%, then he corrected himself saying it was %wrongDirection% | %wrongDirection%, then he corrected himself saying it was %direction%}. In the end he {insisted | sweared} on it being {%direction% | %wrongDirection%}",
+        "{Some | %randomname%s} kid told me {this | a} story {the other day | yesterday | this morning}. It was about {a | some} {place | location} {he calls | called} %location%. He told me it\'s supposed to be {%distance% | %wrongDistance%} to the {%direction% | %wrongDirection%} from here, {%terrain% | %wrongTerrain%}. You {know, I think that story is | believe any of this? Me neither, it\'s} a load of {horse | horse | horse | horse | unhold} shit."
+    ]
+]
```
TaroEld, line 68: The way I understood you, the message would include both the true and false direction/distance, but for some of these it's randomly true or false.

mod_reforged/hooks/config/rumors.nut
```diff
@@ -0,0 +1,68 @@
+// We add up 4 new variables for texts to be build with.
+// Three are wrong distances, direction and terrain for the purpose of creating more interesting rumors.
+// One is an indirect adjective for a legendary location if you don't wanna name it directly
+// This hook will add those vars for contracts and events aswell but they will not make use of those
+local buildTextFromTemplate = ::buildTextFromTemplate;
+::buildTextFromTemplate = function( _text, _vars )
```
TaroEld, line 6: Since we only actually use this fake info in `tavern_building`, this should just be hooking `buildText` in there instead.

mod_reforged/hooks/config/rumors.nut
```diff
@@ -0,0 +1,68 @@
...
+	"spectacular",
+    "surreal"
+];
+
+::Const.Strings.RumorsUniqueLocation <- [
+    [   // Doesn't include the name but the information is correct
+        "{I keep hearing talk | Some traveller mentioned something | Some pilgrim mentioned something} about a %legendaryLocationAdjective% {place | location} {%direction% from here | %terrain% | %distance%}"
+    ],
+    [   // Include the exact name but the other information is usually randomly true or fake
+        "{An explorer | A cartographer} just recently came by, said he found {a | some} %legendaryLocationAdjective% {place | location} %distance% {that he | } {called | named} %location%. If only some more details slipped out of his mouth...",
```
TaroEld, line 62: Something's wrong with `{that he | }`

TaroEld (changes requested):

mod_reforged/hooks/config/rumors.nut
Darxo: That's correct.

Some messages will contain right&wrong alongside so you will know instantly that one of those must be correct and can go after both hints one after another.
And in other cases you only are given one hint that could be true or false

mod_reforged/hooks/config/rumors.nut
Darxo: Oh that's on purpose. I want to increase the variety of these messages so I add these branches whenever possible and it makes sense.

Either it's a location `... that he called Witch Hut ... `
Or it's a location `... called Witch Hut ...`

mod_reforged/hooks/contracts/contracts/return_item_contract.nut
```diff
@@ -0,0 +1,47 @@
...
+    {
+        onPrepareVariables(_vars);
+        foreach (var in _vars)
+        {
+            if (var[0] != "bribe") continue;
+            var[1] = "[b]" + var[1] + "[/b]";     // Bribe is bold now so its easier to spot and compare
+            break;
+        }
+    }
+});
```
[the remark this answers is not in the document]
Darxo: It should be removed now. Don't know how that slipped in

mod_reforged/hooks/config/rumors.nut
Darxo: I don't see how that would be possible.
I have no way accessing the `_vars` while hooking the `buildText` function. It's being assembled in that funtion and then passed directly to the `buildTextFromTemplate`

PR #83, opened 2023-02-27
fix: add missing information about refundability of various perks

Darxo:


mod_reforged/hooks/skills/perks/perk_student.nut
```diff
@@ -0,0 +1,3 @@
+::mods_hookExactClass("skills/perks/perk_student", function (o) {
```
LordMidas, line 1: You should hook `create` and set `this.m.IsRefundable` to false during the create.

LordMidas (changes requested):

LordMidas (approved):

PR #86, opened 2023-03-02
feat: improve quality of life of a few vanilla contracts

Darxo:
- Deliver Item and Caravan Contracts now display their estimated time in days + hours
- Return Item Contract now displays the original payout during a CounterOffer event
- Return Item Contract thieves are a bit easier to catch should they choose to run (10% slower)
- Contracts with a settlement as their destination will now try to display an image of that during the contract dialog

### DIFF
--- a/mod_reforged/config/text/text.nut
+++ b/mod_reforged/config/text/text.nut
@@ -1,0 +1,28 @@
+::Reforged.Text <- {
+    function getDayHourString( _seconds )
+    {
+        local days = ::Math.floor(_seconds / ::World.getTime().SecondsPerDay);
+        local hours = ::Math.floor(_seconds / ::World.getTime().SecondsPerHour) % 24;
+		local dayHourString = "";
+		switch (days)
+		{
+			case 0:
+				break;	// empty string
+			case 1:
+				dayHourString = "1 day and ";
+				break;
+			default:
+				dayHourString = days + " days and ";
+		}
+        dayHourString += (hours == 0) ? "" : hours + " hours";
+        return dayHourString;
+    }
+
+    function getDaysAndHalf( _seconds )
+    {
+        local doubleDays = ::Math.round(2.0 * (_seconds / ::World.getTime().SecondsPerDay));
+        local daysAndHalf = (doubleDays / 2.0);
+        if (daysAndHalf <= 1.0) return "a day"
+        return "" + daysAndHalf + " days";
+    }
+}
--- a/mod_reforged/hooks/contracts/contract.nut
+++ b/mod_reforged/hooks/contracts/contract.nut
@@ -1,0 +1,31 @@
+::mods_hookBaseClass("contracts/contract", function(o)
+{
+	while(!("EmployerID" in o.m)) o = o[o.SuperName]; 	// find the base class
+
+	local getUICharacterImage = o.getUICharacterImage;
+	o.getUICharacterImage = function( _index = 0 )
+	{
+		if ((!("Characters" in this.m.ActiveScreen) || !this.m.ActiveScreen.Characters.len()) &&
+			(!("Banner" in this.m.ActiveScreen) || _index == 0) &&
+			"ShowEmployer" in this.m.ActiveScreen &&
+			this.m.ActiveScreen.ShowEmployer &&
+			_index != 0 &&
+			"Destination" in this.m &&
+			this.m.Destination != null &&
+			!this.m.Destination.isNull() &&
+			this.m.Destination.isLocation() &&
+			this.m.Destination.isLocationType(::Const.World.LocationType.Settlement) &&
+			this.m.Destination.isDiscovered()
+		)
+		{
+			return {
+				Image = this.m.Destination.getImagePath(),
+				IsProcedural = false
+			};
+		}
+		else
+		{
+			return getUICharacterImage(_index);
+		}
+	}
+});
--- a/mod_reforged/hooks/contracts/contracts/deliver_item_contract.nut
+++ b/mod_reforged/hooks/contracts/contracts/deliver_item_contract.nut
@@ -1,0 +1,14 @@
+::mods_hookExactClass("contracts/contracts/deliver_item_contract", function(o) {
+    local onPrepareVariables = o.onPrepareVariables;
+    o.onPrepareVariables = function( _vars )
+    {
+        onPrepareVariables(_vars);
+        foreach (var in _vars)
+        {
+            if (var[0] != "days") continue;
+            local seconds = ::MSU.Time.getSecondsRequiredToTravel(this.m.Flags.get("Distance"), ::Const.World.MovementSettings.Speed, true);
+            var[1] = ::Reforged.Text.getDaysAndHalf(seconds);
+            break;
+        }
+    }
+});
--- a/mod_reforged/hooks/contracts/contracts/escort_caravan_contract.nut
+++ b/mod_reforged/hooks/contracts/contracts/escort_caravan_contract.nut
@@ -1,0 +1,14 @@
+::mods_hookExactClass("contracts/contracts/escort_caravan_contract", function(o) {
+    local onPrepareVariables = o.onPrepareVariables;
+    o.onPrepareVariables = function( _vars )
+    {
+        onPrepareVariables(_vars);
+        foreach (var in _vars)
+        {
+            if (var[0] != "days") continue;
+            local seconds = ::MSU.Time.getSecondsRequiredToTravel(this.m.Flags.get("Distance"), ::Const.World.MovementSettings.Speed * 0.6, true);
+            var[1] = ::Reforged.Text.getDaysAndHalf(seconds);
+            break;
+        }
+    }
+});
--- a/mod_reforged/hooks/contracts/contracts/return_item_contract.nut
+++ b/mod_reforged/hooks/contracts/contracts/return_item_contract.nut
@@ -1,0 +1,47 @@
+::mods_hookExactClass("contracts/contracts/return_item_contract", function(o) {
+    local createStates = o.createStates;
+    o.createStates = function()
+    {
+        createStates();
+        foreach (state in this.m.States)
+        {
+            if (state.ID != "Offer") continue;
+            local end = state.end;
+            state.end = function()
+            {
+                end();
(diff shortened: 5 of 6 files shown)


mod_reforged/hooks/msu.nut
```diff
@@ -61,6 +61,15 @@
 	}
 }
 
+
+::logInfo("Reforged::MSU -- adding ::MSU.Time.getSecondsRequiredToTravel");
+::MSU.Time.getSecondsRequiredToTravel <- function(_numTiles, _speed, _onRoadOnly)
```
LordMidas, line 66: The `::MSU.Time` namespace does not exist, so you need to create it first.

LordMidas, line 66: Space before/after parameters in function declaration.

mod_reforged/hooks/msu.nut
```diff
@@ -61,6 +61,15 @@
...
+{
+	local speed = _speed * ::Const.World.MovementSettings.GlobalMult;
+	if (_onRoadOnly) speed = speed * ::Const.World.MovementSettings.RoadMult;
```
LordMidas, line 69: Sounds like `onRoadOnly` should be an optional parameter that defaults to `false`?

mod_reforged/config/text/text.nut
```diff
@@ -0,0 +1,10 @@
+::Reforged.Text <- {
+    function getDayHourString( _seconds )
+    {
+        local days = ::Math.floor(_seconds / ::World.getTime().SecondsPerDay);
+        local hours = ::Math.floor(_seconds / ::World.getTime().SecondsPerHour);
+        local dayHourString = (days == 0) ? "" : (days == 1) ? "1 day and " : days + " days and ";
```
LordMidas, line 6: Generally avoid nested ternaries. Enduriel and I have discussed this over at MSU and we are not in favor of it. In this particular case, it is still readable with the nested ternaries, so let's see what @Enduriel says.

mod_reforged/config/text/text.nut
```diff
@@ -0,0 +1,10 @@
...
+        dayHourString += (hours == 0) ? "" : hours + " hours"
```
LordMidas, line 7: missing `;`

mod_reforged/config/text/text.nut
```diff
@@ -0,0 +1,10 @@
+::Reforged.Text <- {
+    function getDayHourString( _seconds )
+    {
+        local days = ::Math.floor(_seconds / ::World.getTime().SecondsPerDay);
+        local hours = ::Math.floor(_seconds / ::World.getTime().SecondsPerHour);
```
LordMidas, line 5: As far as I remember, the vanilla "day" doesn't have 24 hours in it. I think it has something like 16. But I may be misremembering it, do you know how many hours it has?

mod_reforged/hooks/contracts/contracts/deliver_item_contract.nut
```diff
@@ -0,0 +1,14 @@
+::mods_hookExactClass("contracts/contracts/deliver_item_contract", function(o) {
+    local onPrepareVariables = o.onPrepareVariables;
+    o.onPrepareVariables = function( _vars )
+    {
+        onPrepareVariables(_vars);
+        foreach (var in _vars)
+        {
+            if (var[0] != "days") continue;
+            local seconds = ::Reforged.Text.getSecondsRequiredToTravel(this.m.Flags.get("Distance"), ::Const.World.MovementSettings.Speed, true);
```
LordMidas, line 9: Personally I don't like showing "days + hours" as it is "too exact" and doesn't fit the atmosphere of the game. What I would prefer is that you round it to the nearest number of half days. So if it takes closer to 1 day, you say 1 day. If it takes closer to 1.5 you say 1 and a half days. If it takes closer to 2 days you say 2 days. And so on. That is still "accurate enough" information for the player to compare, and is immersive in the game's universe.

mod_reforged/hooks/contracts/contracts/return_item_contract.nut
```diff
@@ -0,0 +1,46 @@
+::mods_hookExactClass("contracts/contracts/return_item_contract", function(o) {
+    local createStates = o.createStates;
+    o.createStates = function()
+    {
+        createStates();
+        foreach (state in this.m.States)
+        {
+            if (state.ID != "Offer") continue;
```
LordMidas, line 8: You _only_ want to change something in the `Offer` state. So, please do a `break` at the end of the loop, or enclose this in a positive `if (state.ID == "Offer")` and `break` it inside that if block. In my opinion, when you have a single condition, it is better (for readability) to do a positive check than an extracted inverse check like this. Extraction and inversion is useful when you have several checks.

mod_reforged/hooks/contracts/contracts/return_item_contract.nut
```diff
@@ -0,0 +1,46 @@
...
+            local end = state.end;
+            state.end = function()
+            {
+                end();
+                this.Contract.m.Target.setMovementSpeed(this.Contract.m.Target.getBaseMovementSpeed() * 0.85);  // Thieves are easier to catch
```
LordMidas, line 13: This seems like an _unnecessary_ freebie for the player. I don't think I've ever had the thieves be impossible to catch. So this isn't a QoL change anymore, but rather a reduction in difficulty. I don't think we need this.

mod_reforged/hooks/contracts/contracts/escort_caravan_contract.nut
```diff
@@ -0,0 +1,14 @@
+::mods_hookExactClass("contracts/contracts/escort_caravan_contract", function(o) {
+    local onPrepareVariables = o.onPrepareVariables;
+    o.onPrepareVariables = function( _vars )
+    {
+        onPrepareVariables(_vars);
+        foreach (var in _vars)
+        {
+            if (var[0] != "days") continue;
+            local seconds = ::Reforged.Text.getSecondsRequiredToTravel(this.m.Flags.get("Distance"), ::Const.World.MovementSettings.Speed * 0.6, true);
```
LordMidas, line 9: Same comment as the one about days + hours on deliver item contract.

LordMidas (changes requested):
Overall very good changes. I've requested some changes.

mod_reforged/hooks/contracts/contracts/return_item_contract.nut
Darxo: I also never had thieves impossible to catch. But there were cases where they start running immediately and it took like 12 hours to catch up.

I believe the fleeing is just part of the vanilla AI. Any party will flee from you if you are stronger than them. So during the early game those thieves will always wait for you patiently. And during the lategame when your party is slower because if its bigger size they start fleeing. At which case catching them is most annoying.

I say, choose one of those mechanics: 
- Either always flee (more realistic) 
- or they never flee (simple mechanic for a simple contract).
I prefer the first one. And I still think there should be some minor movement speed penalty on them. Even if it's just 5%

mod_reforged/hooks/msu.nut
```diff
@@ -61,6 +61,17 @@
 	}
 }
 
+
+::logInfo("Reforged::MSU -- adding ::MSU.Time.getSecondsRequiredToTravel");
+::MSU.Time <- {
+	function getSecondsRequiredToTravel( _numTiles, _speed, _onRoadOnly = false )	// This is a close copy of how vanilla calculates their distance duration
+	{
+		local speed = _speed * ::Const.World.MovementSettings.GlobalMult;
```
LordMidas, line 69: You don't _need_ the local variable. You can just do:
```squirrel
_speed *= ::Const.World.MovementSettings.GlobalMult;
if (_onRoadOnly) _speed *= ::Const.World.MovementSettings.RoadMult;
return _numTiles * 170.0 / _speed;
```

mod_reforged/hooks/contracts/contracts/return_item_contract.nut
LordMidas: Yeah I agree that it is a bit inconsistent that they don't flee in the early game (when you are weak) and then start fleeing later on. Of the two options you proposed, I agree with your suggestion i.e. **always flee but have a speed penalty of 5%**.

mod_reforged/config/text/text.nut
Enduriel: Yeah unfortunately as much as it's not too bad in this case I think we should avoid them completely. Really makes you wish for a match statement but without one just use a short switch probably for this unfortunately.

mod_reforged/hooks/contracts/contracts/return_item_contract.nut
Darxo: I don't mind going for either solution in the future but I don't wanna deal with implementing/fixing that right now in this PR.
I changed the movement speed multiplier to be 0.9, so 10% slower than in vanilla.

mod_reforged/hooks/contracts/contracts/return_item_contract.nut
Darxo: How will this case now work under this SOP point?:

> 13. Don't use guard clauses for only one level of indentation.

With the following for-loop we now have two levels of indentation which this guard clause helps to minimize back to one.
I feel like this is exactly the case that guard clauses under `13.` is for now. Keeping nested indentations to a minimum

LordMidas: Guards to me feel like "Run the code for all situations except for the condition in the guard". But what you want here is "Run the code for this one particular condition". It just trips me when I look at this. E.g. I'd be ok with a guard here if the guard was `if (state.ID == "Offer") continue`

mod_reforged/hooks/contracts/contracts/return_item_contract.nut
LordMidas: Fine for now.

LordMidas (approved):

mod_reforged/hooks/contracts/contract.nut
```diff
@@ -0,0 +1,31 @@
+::mods_hookBaseClass("contracts/contract", function(o)
+{
+	while(!("EmployerID" in o.m)) o = o[o.SuperName]; 	// find the base class
```
LordMidas, line 3: Just use `o = o[o.SuperName]`. There is no need for a while loop because `hookBaseClass` hooks the children of the class, so the SuperName will always be the baseclass itself.

PR #87, opened 2023-03-02
feat: randomize hiring cost of brothers between +-20%

Darxo:
I briefly tested this ingame and on the first town it worked fine.

### DIFF
--- a/mod_reforged/config/math/math.nut
+++ b/mod_reforged/config/math/math.nut
@@ -1,0 +1,33 @@
+::Reforged.Math <- {
+	// Returns a random integer between _min and _max
+	// If there is any _luck present then you have the chance for one or more rerolls returning only the value closest to _target
+	function luckyRoll( _min, _max, _target, _luck = 0 )
+	{
+		local _randomNumber = ::Math.rand(_min, _max);
+		for (; _luck > 0; _luck -= 100)
+		{
+			if (::Math.rand(1, 100) <= _luck);
+			{
+				_randomNumber = this.__getCloser(_randomNumber, ::Math.rand(_min, _max), _target);
+			}
+		}
+		return _randomNumber;
+	}
+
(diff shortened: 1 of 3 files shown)


LordMidas:
I feel this is rather unnecessary as hiring cost is already varied by equipment. But I don't see any reason to **not add** this. So, I don't have any strong opinions about it. Please get a review from @Enduriel about the code and then feel free to merge.

LordMidas (approved):

mod_reforged/config/math/math.nut
```diff
@@ -0,0 +1,33 @@
...
+        return _value;
+    }
+
+    // Private helper functions
+    function getCloser( _oldValue, _newValue, _target)
+    {
+        local oldDistance = ::Math.abs(_oldValue - _target);
+        local newDistance = ::Math.abs(_newValue - _target);
+        return (newDistance < oldDistance) ? _newValue : _oldValue;
+    }
```
Enduriel, line 32: Should be double underscored to identify as a private function

mod_reforged/config/math/math.nut
```diff
@@ -0,0 +1,33 @@
...
+			}
+		}
+		return _randomNumber;
+	}
+
+	// A negative _decimalPlaceOffset will result in an equal amounts of 0's left of the decimal place.
+	// ceil(1234, -2) = 1300; ceil(1234, -1) = 1240; ceil(1234.5678, 2) = 1234.57;
+	function ceil( _value, _decimalPlaceOffset = 0 )
+	{
+		_value *= ::Math.pow(10, _decimalPlaceOffset);
```
LordMidas, line 21: We should benchmark the built-in squirrel `pow` function against this `::Math.pow`

LordMidas: I did some benchmark and `::Math.pow` is slightly slower. So I'd say just use `pow`.

PR #91, opened 2023-03-05
feat: mark armor with attachements that you unequip as to-be-repaired

Darxo:
I tested this successfully ingame.

When we expand attachements and layers we probably want to revisit this option and put it behind a mod setting.
But right now it pure upside in 99% of the cases

### DIFF
--- a/mod_reforged/hooks/items/armor/armor.nut
+++ b/mod_reforged/hooks/items/armor/armor.nut
@@ -1,0 +1,8 @@
+::mods_hookExactClass("items/armor/armor", function(o) {
+	local onUnequip = o.onUnequip;
+	o.onUnequip = function()
+	{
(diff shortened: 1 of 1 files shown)


Uberbagel:
I don't understand why this is necessary? What problem is it solving?

Darxo:
It saves you a click every time that you switch out a damaged armor (with an attachement) for any fresh one.
Currently you have to mark them as to be repaired each and every time.

Uberbagel:
Is there a reason this is specific to armors with attachments?

Darxo:
It's the most obvious category of armor which you pretty much always want to have repaired.

If I invest an expensive attachement on a piece of armor then I will want to keep it and reuse it for a long time. So when a player un-equips such an armor it will be in 95%+ of the cases because it is damaged and needs to be repaired or because they want to switch armor around.

Uberbagel:
I don't really like any change that breaks convention in a non-obvious or unnecessary way. It's inconsistent.

PR #93, opened 2023-03-07
refactor: tweak adjacency mechanic and rework polearm mastery

Darxo:
- Adjacency Mechanic now always ignores the first ally
- Polearm Mastery no longer reduces the AP cost by 1
- Polearm Mastery now lets the user ignore adjacent allies
- Polearm Mastery now lets hook and repel ignore shield bonus


LordMidas:
Overall there are some changes in this PR that are interesting, so I'd like to keep it for future reference. For the immediate case, we are going with #94 for a simpler change.

PR #95, opened 2023-03-07
feat: introduce new skill for players to prevent bleeding damage

Darxo:
I tested this ingame and could confirm it working.
The bleed icons still show up at the end of the turn but they deal not damage and tick down normally
I was able to stall a grazed_neck on 1 HP for multiple turns.


LordMidas:
I like the concept - however it needs some work on both the tech and design side:
1. We need to figure out a way to avoid the hard-coded and tightly coupled damage values. This, for me, is on the level of "essential" for this PR.
2. On the design side - we can discuss whether it would be better to have the damage blocked and let the Bleed still tick, or just prevent the Bleed from ticking (i.e. it doesn't apply the damage but doesn't reduce turns either).

PR #97, opened 2023-03-09
feat: only allow AI to rotate if the user or target has adjacent enemy

LordMidas:
Currently, sometimes AI uses Rotation even in the very first round of combat while they are all standing in formation a million miles away from the player. The AI will rotate with an adjacent dude and then move 1 tile and end their turn. This wastes the AI's actions and builds unnecessary Fatigue.

With this change, the AI should hopefully use Rotation in more meaningful circumstances.

### DIFF
--- a/mod_reforged/hooks/skills/actives/barbarian_fury_skill.nut
+++ b/mod_reforged/hooks/skills/actives/barbarian_fury_skill.nut
@@ -1,0 +1,37 @@
+::mods_hookExactClass("skills/actives/barbarian_fury_skill", function(o) {
+	local onVerifyTarget = o.onVerifyTarget;
+	o.onVerifyTarget = function( _originTile, _targetTile )
+	{
+		local ret = onVerifyTarget(_originTile, _targetTile);
+		local user = this.getContainer().getActor();
+
+		// Only allow AI to use the skill if at least one of the two characters has an adjacent enemy
+		if (ret && !user.isPlayerControlled())
+		{
+			ret = false;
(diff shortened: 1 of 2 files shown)


mod_reforged/hooks/skills/actives/barbarian_fury_skill.nut
```diff
@@ -0,0 +1,37 @@
+::mods_hookExactClass("skills/actives/barbarian_fury_skill", function(o) {
```
Darxo, line 1: Barbarian Fury only costs 3 AP and 5 Fatigue. I will not fatigue out barbarians in any way at the start of the battle.
Are you you have seen them use it at the very first turn?

LordMidas: I don't remember if I have seen Barbarians in particular doing this.

mod_reforged/hooks/skills/actives/rotation.nut
```diff
@@ -0,0 +1,37 @@
+::mods_hookExactClass("skills/actives/rotation", function(o) {
```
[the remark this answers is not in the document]
LordMidas: I've seen bandits rotate two frontline dudes and then move 1 tile forward. This was a complete waste of that bandit's resources.

LordMidas:
I don't think this is necessary - the few cases where the AI uses Rotation wastefully are probably not very frequent. We don't have to worry about it. Closing without merging.

PR #106, opened 2023-03-14
New enemy effect: soul link

Darxo:


mod_reforged/hooks/entity/tactical/enemies/skeleton_priest.nut
```diff
@@ -26,7 +26,7 @@
 		if (::Reforged.Config.IsLegendaryDifficulty)
     	{
     		this.m.Skills.add(::new("scripts/skills/perks/perk_anticipation"));
-    		this.m.Skills.add(::new("scripts/skills/perks/perk_nimble"));
+    		this.m.Skills.add(::new("scripts/skills/effects/rf_soul_link_effect"));
```
LordMidas, line 29: I don't find it very thematic for skeleton priests to have this skill. For now, let's add it only on Hexe and Necromancer.

scripts/skills/effects/rf_soul_link_effect.nut
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	function onDamageReceived( _attacker, _damageHitpoints, _damageArmor )
+	{
+		if (_damageHitpoints == 0) return;
+		if (!this.hasLink()) return;
+
+		local transferedDamageMult = (1.0 / (1.0 - this.m.TransferedPart)) - 1.0;
+		this.m.TransferInfo = {
+			Attacker = _attacker.weakref(),
```
LordMidas, line 54: Use `Attacker = ::MSU.asWeakTableRef(_attacker)` instead of doing a manual `.weakref()`

scripts/skills/effects/rf_soul_link_effect.nut
```diff
@@ -0,0 +1,122 @@
...
+			::Sound.play(this.m.SoundOnUse[::Math.rand(0, this.m.SoundOnUse.len() - 1)], ::Const.Sound.Volume.RacialEffect, target.getPos());
+		}
+
+		local hitInfo = clone ::Const.Tactical.HitInfo;
+		hitInfo.DamageRegular = _damage;
+		hitInfo.DamageDirect = 1.0;
+		hitInfo.BodyPart = ::Const.BodyPart.Body;
+		hitInfo.BodyDamageMult = 1.0;
+		hitInfo.FatalityChanceMult = 0.0;
+		target.onDamageReceived(_attacker, this, hitInfo);	// Forwarding the attacker skill instead of 'this' would be nice but we don't have access to that one
```
LordMidas, line 120: You can store the `_skill` of the attacker in the `this.m.TransferInfo` from `onBeforeDamageReceived`. However, I also see arguments for dealing the damage to the linked target from this character and not the `_attacker`. In that case we should be passing `this.getContainer().getActor()` instead of `_attacker`.

For reference, vanilla `hex_slave_effect` applies the Hex damage from the hexed character to himself.

LordMidas (changes requested):

scripts/skills/effects/rf_soul_link_effect.nut
```diff
@@ -0,0 +1,122 @@
...
+		TransferedPart = 0.4,	// This percentage of incoming damage is prevented
+		TransferedMult = 1.0,	// This part of the prevented damage is inflicted to the nearby linked target
+
+		TransferInfo = null
+	},
+	function create()
+	{
+		this.m.ID = "effects.rf_soul_link";
+		this.m.Name = "Soul Link";
+		this.m.Description = "This character can link their soul with adjacent lesser allies. A portion of all incoming damage to Health is transfered to a random adjacent ally. Doesn't work with allies that also have Soul Link.";
```
LordMidas, line 12: replace `Doesn't` with `Does not`

scripts/skills/effects/rf_soul_link_effect.nut
```diff
@@ -0,0 +1,122 @@
...
+		if (this.getLink() == null) return false;
+		return true;
+	}
+
+	function getLink()
+	{
+		local candidates = [];
+		local actor = this.getContainer().getActor();
+		local myTile = actor.getTile();
+		for( local i = 0; i < 6; i = ++i )
```
LordMidas, line 89: This _is_ how the decompiler writes for loops. However, we have the convention of writing them as:
`for (local i = 0; i < 6; i++)`
- Or you can write `++i`, that is also fine.
- Space between `for` and `(`
- No space next to brackets inside the statement

scripts/skills/effects/rf_soul_link_effect.nut
```diff
@@ -0,0 +1,122 @@
...
+		{
+			if (myTile.hasNextTile(i) == false) continue;
+
+			local nextTile = myTile.getNextTile(i);
+			if (!nextTile.IsOccupiedByActor) continue;
+
+			local nextEntity = nextTile.getEntity();
+			if (!nextEntity.isAlliedWith(actor)) continue;
```
LordMidas, line 97: These two ifs should be combined into a single `if` with `||`

scripts/skills/effects/rf_soul_link_effect.nut
```diff
@@ -0,0 +1,122 @@
...
+			"sounds/enemies/dlc2/hexe_hex_damage_04.wav"
+		];
+		this.m.Overlay = "rf_worn_down_effect";
+		this.m.Type = ::Const.SkillType.StatusEffect;
+		this.m.IsRemovedAfterBattle = true;
+	}
+
+	function getName()
+	{
+		if (this.hasLink()) return (this.m.Name + " (Active)");
```
LordMidas, line 29: A better way, imo, is to have this as a `Perk | StatusEffect` skill. And then just hide it when it is not active.

scripts/skills/effects/rf_soul_link_effect.nut
Darxo: I don't want experience/loot to be voided when we kill a body guard via the soul link damage.
So I assumed passing the attacker here will make sure that the player doesn't lose out on any loot

LordMidas: That's a good point. Yeah then let's pass the `_attacker`. But then please store the skill used and pass that too.

Darxo: Ok will do that. But I don't know whether that would be the best way.
Is it really that attacker skill or the soul link that caused the kill? - Somewhat philosophical question

LordMidas: Perhaps passing `this` _is_ better because then the KilledString and other effects related to the death of the linked character are handled properly e.g. we wouldn't want Overhead Strike on a Necromancer to decapitate a linked zombie.

LordMidas (approved):

PR #110, opened 2023-03-15
Perks rebalance

LordMidas:
The changes in this PR are based on my 145 days play test, and also based on discussions with players in the public server. The intention is to reduce perks that are considered non-competitive with other options at their tiers. Also, to reduce the overall number of "mediocre" perks. This is achieved as follows:

- Bestial Vigor, moved from tier 5 to tier 2. Ideally this should be tier 1. It's basically an improved Recover, which is available to the Wildling only and is one use per combat. But on tier 1 we have Pathfinder, so it goes on tier 2 for now.
- Savage Strength is a strong perk. It moves from tier 3 to tier 5.
- Heft is a no-brainer perk for axes right now. So it has been removed and a lite version of its bonus (10% instead of 30%) has been baked into the Cull perk.
- Internal Hemorrhage is a lackluster option at tier 3. So it has been removed and baked directly into the Deep Impact perk at tier 6.
- Dent Armor is a weak and very limited pick at tier 7. It has been moved to tier 3, so it is a competitive pick even in the early game, allowing you to e.g. dent orc warriors in a relatively early stage of the game with Military Picks or Two Handed Wooden Hammers.
- Fluid Weapon and Tempo separately are less competitive perks when your perk points are so limited. So, Fluid Weapon has been baked into Tempo. Tempo now has two parts: the +15 Initiative per attack is available for everyone, but the Fluid Weapon part requires a sword.

### DIFF
--- a/mod_reforged/hooks/config/strings.nut
+++ b/mod_reforged/hooks/config/strings.nut
@@ -1701,13 +1701,23 @@ foreach (vanillaDesc in vanillaDescriptions)
  	}),
 	RF_Tempo = ::UPD.getDescription({
  		Fluff = "By keeping ahead of your opponent, you set the terms of engagement!",
- 		Effects = [{
-			Type = ::UPD.EffectType.Passive,
-			Description = [
-				"Every attack, hit or miss, against a target who acts after you in the current round increases your [Initiative|Concept.Initiative] by " + ::MSU.Text.colorGreen("+15") + ".",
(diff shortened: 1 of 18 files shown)


Uberbagel (commented):
I don't know if users will like the dent armor change without the smackdown style addition we discussed. Otherwise I approve everything else.

LordMidas:
> I don't know if users will like the dent armor change without the smackdown style addition we discussed. Otherwise I approve everything else.

Yeah I've kept the smackdown part for later. It's quite some work to implement it properly and I want to brainstorm it a bit more with you guys.

PR #111, opened 2023-03-15
feat: set coat with spider poison action point cost to 0 during first…

Uberbagel:
… round


LordMidas:
There was a typo that I fixed before merging.

PR #116, opened 2023-03-18
feat: introduce new shared functions for loaded ranged weapon items

Darxo:
These functions include a fix for
- wasting ammo from reloading at the very end of a fight. Now every loaded-ranged-weapon returns its loaded ammo to the ammo-supply pool of the player
- generating ammo out of nowhere at the start of fights because loading-ranged-weapons magically start loaded at the start of each combat. Not they take 1 ammo out of the equipped ammo or start unloaded


mod_reforged/config/make_weapon_loaded.nut
```diff
@@ -0,0 +1,120 @@
...
+	{
+		o.isLoaded <- function()
+		{
+			return this.m.IsLoaded;
+		}
+	}
+
+	if (!"setLoaded" in o)
+	{
+		o.setLoaded <- function( _loaded, _ammoID = null, _ammoCost = 1 )
```
LordMidas, line 20: Why do we allow `null` as ammoID? Shouldn't we always require an ammoID?

mod_reforged/config/make_weapon_loaded.nut
```diff
@@ -0,0 +1,120 @@
...
+		}
+	}
+
+	// On CombatFinished we now unload the weapon
+	if (!"onCombatFinished" in o)
+	{
+		o.onCombatFinished <- function()
+		{
+			this.weapon.onCombatFinished();
+			this.setLoaded(true);
```
LordMidas, line 64: It feels that `this.setLoaded(true);` here is a mistake?

mod_reforged/hooks/skills/actives/reload_bolt.nut
```diff
@@ -10,7 +10,8 @@
 	o.onUse = function( _user, _targetTile	)
 	{
 		this.consumeAmmo();
-		this.getItem().setLoaded(true);
+		local ammoItem = this.getContainer().getActor().getItems().getItemAtSlot(this.Const.ItemSlot.Ammo);
+		this.getItem().setLoaded(true, ammoItem.getID(), ammoItem.getAmmoCost());
```
LordMidas, line 14: I see here that ammo cost is somehow a function of ammo Item. This feels backwards to me. A weapon should define how much of ammo it consumes. What are your thoughts on this?

LordMidas (changes requested):

mod_reforged/config/make_weapon_loaded.nut
Darxo: Part of this being defaulted to null is to make the function backwards compatible with what vanilla and vanilla mods may expect.

This ID is currently only meant as an optional tool for new quiver variants to use. They can at any time check whether the loaded ammunition came from their quiver/pouch and only then apply their passive effects.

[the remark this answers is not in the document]
Darxo: You are thinking of a different ammoCost. This AmmoCost defines how expensive a bolt or bullet is (in terms of Ammunition Supplies)

Enduriel (changes requested):
LoadedAmmunitionID -> Reference to quiver
LoadedAmmunitionCost -> remove
setLoaded -> pass quiver
```if (!"onCombatStarted" in o)```
use null and combine big if
Move `::Reforged.Items.makeWeaponLoaded(this);` outside of `create`

File: mod_reforged/config/make_weapon_loaded.nut [hunk 4]
```diff
@@ -0,0 +1,80 @@
...
+	// In onCombatFinished we now unload the weapon.
+	// This overwrites all vanilla functions because they all set this weapon to loaded at this point.
+	o.onCombatFinished <- function()
+	{
+		if (this.isLoaded())
+		{
+			::World.Assets.addAmmo((this.m.LoadedAmmunitionItem == null) ? 1 : this.m.LoadedAmmunitionItem.getAmmoCost());
+			this.setLoaded(false);
+		}
+	}
```
Enduriel, line 55: I still think we should call the original function and then unload it ourselves if needed to maximize compatibility.

mod_reforged/config/make_weapon_loaded.nut
```diff
@@ -0,0 +1,80 @@
...
+	o.onCombatStarted <- function()
+	{
+		if (onCombatStarted == null)
+		{
+			this.weapon.onCombatStarted();
+		}
+		else
+		{
+			onCombatStarted();
+		}
```
Enduriel, line 68: the better way that we thought of to do this is
```squirrel
local parentName = o.SuperName;
local onCombatStarted = ("onCombatStarted" in o) ? o.onCombatStarted : null;
o.onCombatStarted <- function()
{
	local ret = onCombatStarted == null ? this[parentName].onCombatStarted() : onCombatStarted();
```
it's more reliable if someone makes a weapon inherit from something else before that inherits from weapon, eg an abstract bow or something.

Enduriel (changes requested):

PR #117, opened 2023-03-19
refactor: rework inspiring presence to not require a banner anymore

Darxo:
Inspiring Presence now only turns on for the brother with the highest resolve if multiple perks are present 
Inspiring Presence only affects allies that have less resolve than the emitter


Darxo:
For now this is implemented in Hardened. EVentually I will bring it up again

PR #120, opened 2023-03-21
Perks rebalance

LordMidas:
Based on player feedback and discussion in the dev team, this PR intends to consolidate lackluster perks into more useful and attractive options, while rebalancing some others.

Key changes:
- Retribution is merged as a passive into the Indomitable perk.
- Trip Artist is merged as a passive into the Offhand Training perk.
- Between the Ribs (requires Dagger) is merged as a passive into the Backstabber perk.
- Weapon Master reworked: No longer grants a free swap. Instead grants a key perk from the equipped weapon's perk group as long as you have that perk in your perk tree.
- Formidable Approach moved from tier 5 to tier 7.
- Sweeping Strikes moved from tier 6 to tier 5.
- Rotation skill Fatigue Cost dropped from 25 to 20.
- Banner now has a Reach of 5, and Bolster now has an exception to work with the Banner.
- Removed Between the Ribs from Serpent and added Force Enabled Backstabber (due to the changes to Backstabber above).

EDIT:
New changes:
- Set Banner reach to 6 instead of setting to 5 and adding exception in Bolster.
- Convert Weapon Master to a special perk that is rolled only if you have throwing perk group or both swift and shield perk groups.


Uberbagel (approved):

mod_reforged/hooks/config/strings.nut
Darxo, line 1: This between the ribs backstabber commit is missing changes to the perk groups where you remove between the ribs from appearing in the game by itself

mod_reforged/hooks/items/tools/faction_banner.nut
```diff
@@ -0,0 +1,8 @@
+::mods_hookExactClass("items/tools/faction_banner", function(o) {
+	local create = o.create;
+	o.create = function()
+	{
+		create();
+		this.m.Reach = 5;
```
Darxo, line 6: Setting the Reach to 6 and removing the custom exception in Bolster seems a much better solution

mod_reforged/hooks/config/strings.nut
```diff
@@ -1322,7 +1322,8 @@ foreach (vanillaDesc in vanillaDescriptions)
 			Type = ::UPD.EffectType.Passive,
 			Description = [
 				"Once per [turn|Concept.Turn], the first use of the tool in your offhand costs no [Action Points|Concept.ActionPoints].",
-				"Once per [turn|Concept.Turn], switching a buckler or tool in your offhand costs no [Action Points|Concept.ActionPoints]. Does not stack with other free item swapping skills e.g. Quick Hands."
+				"Once per [turn|Concept.Turn], switching a buckler or tool in your offhand costs no [Action Points|Concept.ActionPoints]. Does not stack with other free item swapping skills e.g. Quick Hands.",
+				"When equipped with a net, gain the [Trip Artist|NullEntitySkill+rf_trip_artist_effect] effect."
```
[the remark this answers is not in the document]
LordMidas: Yea I agree with you and I also like doing that generally. But in this particular case these are two separate effects, and the issue is that Trip Artist and Offhand Training both show up as status effects and they both mean different things. So I want the player to be very clear that Trip Artist is a separate effect so the player recognizes it separately when he sees the icon and knows where it is coming from. I did the same thing e.g. with Retribution / Indomitable and Tempo / Fluid Weapon.

PR #121, opened 2023-03-22
refactor: add dynamic self multiplier for ranged perk group

LordMidas:
- At least one ranged weapon perk group must be present (bow, crossbow or throwing). Otherwise the ranged perk group cannot be rolled.
- For each ranged weapon perk group present, the multiplier increases by 1. 
- For each talent star in Ranged Skill the multiplier increases by 1.


Darxo (approved):
I'm suprised to receive a review request for a branch this small and with such little impact to the gameplay.

LordMidas:
> I'm suprised to receive a review request for a branch this small and with such little impact to the gameplay. Usually you commit these directly onto the dev branch

Nah, I always discuss changes to design with at least someone in the design team. Usually I discuss on Discord before opening a PR but this seemed like a rather straightforward change which is why I opened a PR for it and requested review from you both so you can comment (I didn't expect much discussion on it).

Uberbagel:
With this change, the Ranged Group weight multiplier from Ranged Skill talent of 2 should be removed as this is now handled as a self-weight on Ranged Group.

PR #122, opened 2023-03-22
Swordmaster perks hybrid swords

LordMidas:
- Allow hybrid swords to be used for Swordmaster perks.
- Allow Sword/Cleaver hybrids to be a valid weapon for Metzger (in addition to southern swords).

If we run into particular issues where the perks feel weird with certain weapons, we can then exclude them as necessary.

### DIFF
--- a/mod_reforged/hooks/config/strings.nut
+++ b/mod_reforged/hooks/config/strings.nut
@@ -1665,13 +1665,13 @@ foreach (vanillaDesc in vanillaDescriptions)
  	}),
 	RF_SwordmasterMetzger = ::UPD.getDescription({
  		Fluff = "A sword, too, can take someone\'s head off just fine!",
- 		Requirement = "Southern Sword"
+ 		Requirement = "Southern Sword or a Sword/Cleaver Hybrid"
  		Effects = [
 	 		{
 				Type = ::UPD.EffectType.Passive,
 				Description = [
 					"Adds the [Decapitate|Skill+decapitate] skill to southern swords.",
-					"Grants the [Sanguinary|Perk+perk_rf_sanguinary] and [Bloodbath|Perk+perk_rf_bloodbath] perks for free as long as a southern sword is equipped.",
+					"Grants the [Sanguinary|Perk+perk_rf_sanguinary] and [Bloodbath|Perk+perk_rf_bloodbath] perks for free as long as a valid weapon is equipped.",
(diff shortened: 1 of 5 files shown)


scripts/skills/perks/perk_rf_swordmaster_metzger.nut
```diff
@@ -16,7 +16,8 @@ this.perk_rf_swordmaster_metzger <- ::inherit("scripts/skills/perks/perk_rf_swor
 		local ret = this.perk_rf_swordmaster_abstract.isEnabled();
 		if (ret)
 		{
-			if (!this.getContainer().getActor().getMainhandItem().isItemType(::Const.Items.ItemType.RF_Southern))
+			local weapon = this.getContainer().getActor().getMainhandItem();
+			if (!weapon.isItemType(::Const.Items.ItemType.RF_Southern) && weapon.isWeaponType(::Const.Items.WeaponType.Cleaver))
```
Darxo, line 20: The perk description mentions that it requires a southern sword ...
But a pure northern sword will pass this check because it is not a Cleaver.

This function currently only checks against Northern Cleavers (that are by extensions also swords because of the previous check)
Apart from norther swords above this condition will not check against other hybrid sword types like goblin spear/swords or sword/daggers

LordMidas: In the change in this PR, this perk requires a Southern Sword or a Sword/Cleaver hybrid. So the condition says:

So, if the thing is a southern sword, then this `if` will fail. But if the thing _is_ a southern sword, then it checks if it is a cleaver. If it is a cleaver that means it is a Sword/Cleaver hybrid (because `ret` can only be true for swords).

So to me the condition says, either be a southern sword, or be a cleaver (i.e. sword/cleaver hybrid). Which seems correct. 

Lemme know if there's a mistake in my thinking.

mod_reforged/hooks/config/strings.nut
```diff
@@ -1686,7 +1686,7 @@ foreach (vanillaDesc in vanillaDescriptions)
  	}),
  	RF_SwordmasterPrecise = ::UPD.getDescription({
  		Fluff = "Let me show you how to thread a needle... blindfolded!",
- 		Requirement = "Non-hybrid Sword"
+ 		Requirement = "Sword"
```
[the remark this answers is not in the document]
LordMidas: Will fix that

Uberbagel (approved):

PR #123, opened 2023-03-24
refactor: combine sprint into footwork and rename it "Escape Artist"

Darxo:
- "Footwork" perk is renamed to "Escape Artist"
- Sprint is now part of "Escape Artist" perk.
- Sprint now also costs 1 AP and no longer increases fatigue cost for moving


LordMidas:
As discussed on Discord, I think Sprint fits better balance-wise in Pathfinder (as balancing it in Footwork becomes very hard for the combo of Pathfinder + Footwork across multiple terrain types, whereas in the current state it balances itself). And based on some discussion that I had with several playtesters - the skill feels in a good place at the moment. So, I would hold off on this change.

PR #124, opened 2023-03-24
refactor: slightly tweak knife and dagger to reduce highrolling

Darxo:
Both lose 5 maximum damage but gain 5% armor damage/pen


LordMidas:
1. Knife and Dagger having higher armor penetration than Rondel Dagger seems wrong from a weapon design perspective.
2. This seems like a band-aid patch solution for a problem that should be fixed in a more elegant way.

PR #126, opened 2023-03-26
feat: introduce new craftable item "ayahuasca" to increase perk tier

Darxo:


Darxo:
This is currently implemented in my crock pot mod so I will close it here for the time being

PR #128, opened 2023-03-31
refactor: Give root skill a 3 turn cooldown

Uberbagel:


mod_reforged/hooks/skills/actives/root_skill.nut
```diff
@@ -0,0 +1,21 @@
+::mods_hookExactClass("skills/actives/root_skill", function(o) {
```
LordMidas, line 1: This is not the correct way. Look at our hook on rally_skill to see how to implement a better cooldown system. Then when you add the skill to the entity you specify the cooldown via `::MSU.new` (for this see how we add rally_skill to brigand_leader)

LordMidas (changes requested):

Uberbagel:
I just pulled the code directly from PTR. I'll let tech handle it and focus on getting ahead on design instead.

LordMidas:
> I just pulled the code directly from PTR. I'll let tech handle it and focus on getting ahead on design instead.

Yes that's fine. I'll do it.

PR #137, opened 2023-07-13
feat: implement first turn initiative AP bonus

Darxo:
During the first turn of every combat every characters Action Points are slightly modified depending on their current Initiative at that time.

image

Possible Todo: 
- Maybe get a unique Icon for this effect. Currently it uses the Impatient-Trait Icon
- Move the two "public" variables over to the `Characters.nut` so that other skills like perks can influence these values

### DIFF
--- a/mod_reforged/hooks/entity/tactical/actor.nut
+++ b/mod_reforged/hooks/entity/tactical/actor.nut
@@ -120,6 +120,16 @@
 		return __original(_change, _difficulty, _type, _showIconBeforeMoraleIcon, _noNewLine);
 	}
 
+	q.onRoundStart = @(__original) function()
+	{
+		if (this.isAlive() && ::Time.getRound() == 1)
+		{
+			this.getSkills().add(::new("scripts/skills/special/rf_first_turn_initiative"));
+		}
+
+		__original();
+	}
+
 	q.getSurroundedCount = @(__original) function()
 	{
 		local startSurroundCountAt = this.m.CurrentProperties.StartSurroundCountAt;
--- a/scripts/skills/special/rf_first_turn_initiative.nut
+++ b/scripts/skills/special/rf_first_turn_initiative.nut
@@ -1,0 +1,74 @@
+this.rf_first_turn_initiative <- ::inherit("scripts/skills/skill", {
+	m = {
+		// Config
+		InitiativePerAP = 40.0,	// For every this much Initiative an entity gains 1 AP. The received bonus can be negative
+		FlatAPBonus = -2,	// This is to balance out this new bonus for most entities
+		TotalMinimum = -3,
+		TotalMaximum = 7,	// Maximum is just in case so that enemies that accidentally have extreme initiative don't one-shot you Turn one
+
+		// Const
+		IconPositive = "ui/traits/trait_icon_46.png",	// Impatient Trait
+		IconNegative = "ui/traits/trait_icon_25.png",	// Hesitant Trait
+
+		// Private
+		CurrentAPBonus = 0,
+	},
+	function create()
+	{
+		this.m.ID = "special.rf_first_turn_initiative";
+		this.m.Name = "First Turn Initiative";
+		this.m.Description = "During the first turn of every combat, this character\'s [Action Points|Concept.ActionPoints] are modified depending on their [Initiative|Concept.Initiative] at that point.";
+		this.m.Icon = "ui/traits/trait_icon_46.png";
+		this.m.Type = ::Const.SkillType.StatusEffect;
+		this.m.Order = ::Const.SkillOrder.Last;		// Some effects may set the Action Points to a flat value instead of increasing them, like 'possessed_undead_effect'
+		this.m.IsSerialized = false;
+		this.m.IsRemovedAfterBattle = true;
+	}
+
+	function getDescription()
+	{
+		local ret = this.skill.getDescription();
+
+		ret += "\nThis character gains 1 [Action Point|Concept.ActionPoints] for every " + ::MSU.Text.colorGreen(this.m.InitiativePerAP) + " [Initiative|Concept.Initiative] which is then modified by a flat " + ::MSU.Text.colorizeValue(this.m.FlatAPBonus) + " [Action Points|Concept.ActionPoints].";
+
+		return ::Reforged.Mod.Tooltips.parseString(ret);
+	}
+
+	function getTooltip()
+	{
+		local ret = this.skill.getTooltip();
+
+		ret.push({
+			id = 10,
+			type = "text",
+			icon = "ui/icons/action_points.png",
+			text = ::Reforged.Mod.Tooltips.parseString(::MSU.Text.colorizeValue(this.m.CurrentAPBonus) + " [Action Point(s)|Concept.ActionPoints] for this [turn|Concept.Turn]"),
+		});
+
+		return ret;
+	}
+
+	function isHidden()
+	{
+		return this.m.CurrentAPBonus == 0;
+	}
+
+	function onUpdate( _properties )
+	{
+		_properties.ActionPoints += this.m.CurrentAPBonus;
+	}
+
+	function onNewRound()
+	{
+		this.m.CurrentAPBonus = this.m.FlatAPBonus;
+		this.m.CurrentAPBonus += ::Math.floor(this.getContainer().getActor().getInitiative() / this.m.InitiativePerAP);
+		this.m.CurrentAPBonus = ::Math.min(this.m.TotalMaximum, ::Math.max(this.m.TotalMinimum, this.m.CurrentAPBonus));
+
+		this.m.Icon = this.m.CurrentAPBonus > 0 ? this.m.IconPositive : this.m.IconNegative;
+	}
+
+	function onTurnEnd()
+	{
+		this.removeSelf();
+	}
+});


scripts/skills/special/rf_first_turn_initiative.nut
```diff
@@ -0,0 +1,77 @@
...
+	{
+		return (this.m.APBonus == 0);
+	}
+
+	function onUpdate( _properties )
+	{
+		_properties.ActionPoints += this.m.APBonus;
+	}
+
+	function onNewRound()
```
LordMidas, line 58: Wouldn't it be sufficient to do this in `onCombatStarted()`? Then you won't need the switcheroo and it can be much cleaner. Similarly you could add the effect in `actor.nut` during the `actor.onCombatStart` function before calling the original function.

scripts/skills/special/rf_first_turn_initiative.nut
```diff
@@ -0,0 +1,77 @@
...
+
+		this.m.APBonus = this.m.FlatAPBonus;
+		this.m.APBonus += ::Math.floor(this.getContainer().getActor().getInitiative() / this.m.InitiativePerAP);
+		this.m.APBonus = ::Math.min(this.m.TotalMaximum, ::Math.max(this.m.TotalMinimum, this.m.APBonus));
+
+		if (this.m.APBonus < 0) this.m.Icon = this.m.IconNegative;
+	}
+
+	// This effect only affects the very first round of each combat
+	function onRoundEnd()
```
LordMidas, line 73: It would be more efficient to remove it `onTurnEnd()` instead of waiting for the round to end.

scripts/skills/special/rf_first_turn_initiative.nut
```diff
@@ -0,0 +1,77 @@
...
+			type = "text",
+			icon = "ui/icons/action_points.png",
+			text = ::MSU.Text.colorizeValue(this.m.APBonus) + " Action Points for this turn"
+		});
+		return tooltip;
+	}
+
+	function isHidden()
+	{
+		return (this.m.APBonus == 0);
```
LordMidas, line 50: Parantheses are unnecessary and should be removed.

scripts/skills/special/rf_first_turn_initiative.nut
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	function getTooltip()
+	{
+		local tooltip = this.skill.getTooltip();
+		tooltip.push({
+			id = 10,
+			type = "text",
+			icon = "ui/icons/action_points.png",
+			text = ::MSU.Text.colorizeValue(this.m.APBonus) + " Action Points for this turn"
```
LordMidas, line 43: Please add hyperlinks for Action Points and turn concepts.

scripts/skills/special/rf_first_turn_initiative.nut
```diff
@@ -0,0 +1,77 @@
...
+		// Const
+		IconNegative = "ui/traits/trait_icon_25.png"	// Hesitant Trait
+		TotalMinimum = -3,
+		TotalMaximum = 7,	// Maximum is just incase so that enemies that accidentally have extreme initiative don't one-shot you Turn one
+	},
+	function create()
+	{
+		this.m.ID = "special.rf_first_turn_initiative";
+		this.m.Name = "First Turn Initiative";
+		this.m.Description = "During the first turn of every combat, this characters [Action Points|Concept.ActionPoints] are modified depending on their [Initiative|Concept.Initiative] at that point.";
```
LordMidas, line 19: `character\'s`

scripts/skills/special/rf_first_turn_initiative.nut
```diff
@@ -0,0 +1,77 @@
...
+		InitiativePerAP = 40.0,	// For every this much Initiative an entity gains 1 AP. The bonus can be negative
+		FlatAPBonus = -2,	// This is to balance out this new bonus for most entities
+
+		// Private
+		APBonus = 0
+
+		// Const
+		IconNegative = "ui/traits/trait_icon_25.png"	// Hesitant Trait
+		TotalMinimum = -3,
+		TotalMaximum = 7,	// Maximum is just incase so that enemies that accidentally have extreme initiative don't one-shot you Turn one
```
LordMidas, line 13: `in case`

LordMidas (changes requested):

scripts/skills/special/rf_first_turn_initiative.nut
Darxo: I get that, and this is mainly personal preference. I like me some extra parentheses here and there to emphasize certain operations and hopefully improve readability.

scripts/skills/special/rf_first_turn_initiative.nut
Darxo: I don't see where the `onCombatStart` exists in the actor.nut. I looked in vanilla, msu and reforged. No one seems to declare it.

Darxo: I'm unsure if `onCombatStarted` from the skill.nut is the answer here. It looks to be called in the same way as `onNewRound` is. In both function the `IsUpdating` is set to true, which prevents manual update calls.

scripts/skills/special/rf_first_turn_initiative.nut
LordMidas: In my opinion this does not improve readability and is detrimental as I have to parse the parantheses when reading it.

scripts/skills/special/rf_first_turn_initiative.nut
LordMidas: Ah yes you are correct about `actor.nut` not having `onCombatStart`. In vanilla that only exists in `player.nut` which we, over at MSU, are "fixing" by adding it also to `actor.nut` because this is what causes `onCombatStarted` to not be called for non-player actors in vanilla which we consider a [bug](http://battlebrothersgame.com/forums/topic/oncombatstarted-is-not-called-for-ai-characters/).

Regarding the update, you are correct `IsUpdating` is true during `onCombatStarted`. However, this means that as long as `onNewRound` is called **after** `onCombatStarted` the character has already updated at the end of `onCombatStarted` and therefore the properties should be up to date. So you actually don't need to sneak in a new update here.

scripts/skills/special/rf_first_turn_initiative.nut
```diff
@@ -0,0 +1,71 @@
+this.rf_first_turn_initiative <- ::inherit("scripts/skills/skill", {
+	m = {
+		// Public
```
LordMidas, line 3: Public is rather misleading as these aren't meant to be accessed/changed from outside necessarily. If you're gonna put a comment here I'd suggest renaming it to Config.

scripts/skills/special/rf_first_turn_initiative.nut
```diff
@@ -0,0 +1,71 @@
...
+		InitiativePerAP = 40.0,	// For every this much Initiative an entity gains 1 AP. The received bonus can be negative
+		FlatAPBonus = -2,	// This is to balance out this new bonus for most entities
+
+		// Const
+		IconPositive = "ui/traits/trait_icon_46.png",	// Impatient Trait
+		IconNegative = "ui/traits/trait_icon_25.png",	// Hesitant Trait
+		TotalMinimum = -3,
+		TotalMaximum = 7,	// Maximum is just in case so that enemies that accidentally have extreme initiative don't one-shot you Turn one
```
LordMidas, line 11: Both these Total fields can then move to the Config part above.

scripts/skills/special/rf_first_turn_initiative.nut
```diff
@@ -0,0 +1,71 @@
...
+	function getDescription()
+	{
+		local description = this.m.Description;
+		description += "\nThis character gains 1 [Action Point|Concept.ActionPoints] for every " + ::MSU.Text.colorGreen(this.m.InitiativePerAP) + " [Initiative|Concept.Initiative] which is then modified by a flat " + ::MSU.Text.colorizeValue(this.m.FlatAPBonus) + " [Action Points|Concept.ActionPoints].";
+		return ::Reforged.Mod.Tooltips.parseString(description);
+	}
+
+	function getTooltip()
+	{
+		local tooltip = this.skill.getTooltip();
```
LordMidas, line 37: Change to `local ret`

scripts/skills/special/rf_first_turn_initiative.nut
```diff
@@ -0,0 +1,71 @@
...
+		CurrentAPBonus = 0,
+	},
+	function create()
+	{
+		this.m.ID = "special.rf_first_turn_initiative";
+		this.m.Name = "First Turn Initiative";
+		this.m.Description = "During the first turn of every combat, this character\'s [Action Points|Concept.ActionPoints] are modified depending on their [Initiative|Concept.Initiative] at that point.";
+		this.m.Icon = "ui/traits/trait_icon_46.png";
+		this.m.Type = ::Const.SkillType.StatusEffect;
+		this.m.Order = ::Const.SkillOrder.Last + 100;		// Some effects may set the Action Points to a flat value instead of increasing them, like 'possessed_undead_effect'
```
LordMidas, line 23: I don't like the `+ 100` part. I'd just stick to `Last`. We can deal with potential edge cases when they happen.

scripts/skills/special/rf_first_turn_initiative.nut
```diff
@@ -0,0 +1,71 @@
...
+			type = "text",
+			icon = "ui/icons/action_points.png",
+			text = ::Reforged.Mod.Tooltips.parseString(::MSU.Text.colorizeValue(this.m.CurrentAPBonus) + " [Action Point(s)|Concept.ActionPoints] for this [turn|Concept.Turn]")
+		});
+		return tooltip;
+	}
+
+	function isHidden()
+	{
+		return (this.m.CurrentAPBonus == 0);
```
LordMidas, line 49: Please remove the parantheses.

scripts/skills/special/rf_first_turn_initiative.nut
```diff
@@ -0,0 +1,71 @@
...
+		this.m.Icon = "ui/traits/trait_icon_46.png";
+		this.m.Type = ::Const.SkillType.StatusEffect;
+		this.m.Order = ::Const.SkillOrder.Last + 100;		// Some effects may set the Action Points to a flat value instead of increasing them, like 'possessed_undead_effect'
+		this.m.IsSerialized = false;
+		this.m.IsRemovedAfterBattle = true;
+	}
+
+	function getDescription()
+	{
+		local description = this.m.Description;
```
LordMidas, line 30: Would be good to change to `local ret` here too

mod_reforged/hooks/entity/tactical/actor.nut
```diff
@@ -120,6 +120,12 @@
 		return __original(_change, _difficulty, _type, _showIconBeforeMoraleIcon, _noNewLine);
 	}
 
+	q.onRoundStart = @(__original) function()
+	{
+		if (this.isAlive() && ::Time.getRound() == 1) this.getSkills().add(::new("scripts/skills/special/rf_first_turn_initiative"));
```
LordMidas, line 125: In early parts of the code base we didn't always new-line `if` statements, but I know Enduriel is very adamant on this that they should always be newlined. So it should be:
```squirrel
if (this.isAlive() && ::Time.getRound() == 1)
	this.getSkills().add(::new("scripts/skills/special/rf_first_turn_initiative"));
```

LordMidas (changes requested):
Just a few minor changes requested.

LordMidas (approved):

PR #139, opened 2023-07-26
Fix combat tooltip out of bounds

Darxo:
The first commit will remove an our initial attempt on fixing the out of bounce issue. We hooked the wrong function back then and never realised that.

The second commit will fix the out of bounce issue regarding the top border. If the tooltip is too longer than the window is high then the bottom part will still go out of bounce. But the bottom part is always the less important one anyways.

The third commit will fix a bug with regular UI-Tooltips that are too long to be displayed above the curser and also too long to be displayed below the cursor.
Currently this only happens to me when viewing certain famed weapons in shops on a 1600*900 resolution.
This commit will make it so their full tooltip is shown instead of it overflowing out of the bottom screen border


Enduriel:
@Darxo please adjust the names of the commits, you mean out of bounds not out of bounce, would be annoying to have to remember that when searching for it in the future. All good though English is annoying with similar words lol

Enduriel:
Once that's done assume I've approved as the rest looks good from my side

PR #140, opened 2023-07-26
refactor: rework anticipation perk to grant damage reduction instead

Darxo:


mod_reforged/hooks/config/strings.nut
```diff
@@ -186,6 +186,21 @@ local vanillaDescriptions = [
...
 	},
+	{
+		ID = "perk.anticipation",
+		Key = "Anticipation",
+		Description = ::UPD.getDescription({
+			Fluff = "I saw these coming!",
+	 		Effects = [{
+ 				Type = ::UPD.EffectType.Passive,
+ 				Description = [
+					"Take reduced damage from the first " + ::MSU.Text.colorGreen(2) + " attacks you receive each battle.",
```
LordMidas, line 197: These both points should be merged into a single point that reads much better: `Gain percentage damage reduction equal to your Ranged Defense for the first two attacks, ranged or melee, against you in every battle.`

mod_reforged/hooks/config/strings.nut
```diff
@@ -186,6 +186,21 @@ local vanillaDescriptions = [
...
+					"This reduction is equal to your current [Ranged Defense|Concept.RangeDefense] + an additional " + ::MSU.Text.colorGreen("10%") + " for each tile between the attacker and you."
```
LordMidas, line 198: Please remove the 10% for each additional tile part. I believe it is too strong; the perk is going to be very good even without it.

mod_reforged/hooks/skills/perks/perk_anticipation.nut
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	o.onCombatFinished <- function()
+	{
+		this.m.UsesRemaining = this.m.UsesMax;	// So that for the purposes of the tooltip everything looks good
+	}
+
+	o.getDescription <- function()
+	{
+		return ("Take reduced damage from the first " + ::MSU.Text.colorGreen(this.m.UsesMax) + " attacks you receive each battle. This reduction is dependant on your current Ranged Defense and the distance between the attacker and you");
```
LordMidas, line 40: Try to avoid overwriting `getDescription` as much as possible. Instead of this, set `this.m.Description` in the `create` hook to `Use your fast senses to anticipate attacks against you, taking reduced damage.` or something like that so that it is a static string.

Then in the `getTooltip` function, add an entry that shows how many next attacks the effect is valid for.

mod_reforged/hooks/skills/perks/perk_anticipation.nut
```diff
@@ -0,0 +1,115 @@
...
+
+
+	// private functions
+	o.isEnabled <- function()
+	{
+		return (this.m.UsesRemaining > 0);
+	}
+
+	// returns a value between 0 and 100 indicitation the damage reduction in %
+	o.calculateDamageReduction <- function( _attacker = null )
```
LordMidas, line 105: Once you remove the 10% per tile of distance part, then the `_attacker` parameter can be removed from this function.

mod_reforged/hooks/skills/perks/perk_anticipation.nut
```diff
@@ -0,0 +1,115 @@
...
+	o.onBeforeDamageReceived <- function( _attacker, _skill, _hitinfo, _properties )
+	{
+		if (!this.isEnabled() || _skill == null || !_skill.isAttack())		// _skill can be null, for example when burning damage is applied
+		{
+			this.m.IsAboutToConsumeUse = false;
+			return;
+		}
+
+		this.m.TempDamageReduction = this.calculateDamageReduction(_attacker);	// We save this so that we can later display it in the combat log
+		local damageReceivedMultiplier = (1.0 - (this.m.TempDamageReduction / 100.0));
```
LordMidas, line 81: Saving a variable like this is unnecessary and inefficient. You can and should inline this as `_properties.DamageReceivedTotalMult *= 1.0 - (this.m.TempDamageReduction / 100.0)`

mod_reforged/hooks/skills/perks/perk_anticipation.nut
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	o.onDamageReceived <- function( _attacker, _damageHitpoints, _damageArmor )
+	{
+		if (this.m.IsAboutToConsumeUse == false) return;	// We only consume one use for each registrated attack. Something like the vanilla three headed flail might otherwise use up all of this perks charges
+		this.m.IsAboutToConsumeUse = false;
+
+		local actor = this.getContainer().getActor();
+		this.m.UsesRemaining = ::Math.max(0, this.m.UsesRemaining - 1);
+		::Tactical.EventLog.logEx(::Const.UI.getColorizedEntityName(this.getContainer().getActor()) + " anticipated the attack of " + ::Const.UI.getColorizedEntityName(_attacker) + ", reducing its damage by " + ::MSU.Text.colorGreen(this.m.TempDamageReduction + "%"));
```
LordMidas, line 94: `reducing damage received by` instead of `reducing its damage by`

mod_reforged/hooks/skills/perks/perk_anticipation.nut
```diff
@@ -0,0 +1,115 @@
...
+		}
+
+		this.m.TempDamageReduction = this.calculateDamageReduction(_attacker);	// We save this so that we can later display it in the combat log
+		local damageReceivedMultiplier = (1.0 - (this.m.TempDamageReduction / 100.0));
+		_properties.DamageReceivedTotalMult *= damageReceivedMultiplier;
+
+		this.m.IsAboutToConsumeUse = true;	// We need this variable because in "onDamageReceived" we have no information whether that was caused by an "attack"
+	}
+
+	o.onDamageReceived <- function( _attacker, _damageHitpoints, _damageArmor )
```
LordMidas, line 87: Do note that this function is not called if the entity dies from the attack. So, in that case the combat log won't show up. Which is probably not a big deal.

mod_reforged/hooks/skills/perks/perk_anticipation.nut
```diff
@@ -0,0 +1,115 @@
...
+	{
+		if (this.m.IsAboutToConsumeUse == false) return;	// We only consume one use for each registrated attack. Something like the vanilla three headed flail might otherwise use up all of this perks charges
```
LordMidas, line 89: I don't see how this system stops the vanilla three-headed flail from proccing this perk multiple times in one attack. In every attack of the three-headed flail `onBeforeDamageReceived` will be called first and then `onDamageReceived` - and each attack will be treated individually so it seems it will consume all charges.

In order to stop it from consuming all charges, you should use something similar to what vanilla uses in e.g. Overwhelm. Keep track of the SkillCounter and the attacker. And if the skill counter didn't increase, and its the same attacker, then you know that it is a three-headed flail attack.

LordMidas (changes requested):
Overall an interesting design. I have comments on the implementation which need fixing. The additional 10% reduction per tile needs to be removed.

mod_reforged/hooks/skills/perks/perk_anticipation.nut
Darxo: You're right, this won't prevent such a case. So I will change the comment.
However this line would prevent weird additional non-attack damage from the same attack from using up more than one charge. I don't know of any actual skills that do this though. They would need to manually call an additional onDamageReceived on the target.

mod_reforged/hooks/config/strings.nut
Darxo: From my testing it's the most essential part of this perk and not too strong. 
Please keep in mind that this only prevents the **first two attacks** in each battle and it quite literally follows in the footsteps of the old Anticipation, offering Player and Enemy alike a protection against getting sniped from far away.

LordMidas: Oh, I thought each bro has like 35 base Ranged Defense. I was confusing it with Ranged Skill in my mind. Then disregard my comment - your design seems fine.

mod_reforged/hooks/skills/perks/perk_anticipation.nut
LordMidas: Ideally I want to add a `onPerformedAttack` and `onReceivedAttack` events via MSU to handle such cases. In vanilla the presence of `onDamageReceived` is too limiting and not "exact" in terms of detecting received attacks.

mod_reforged/hooks/skills/perks/perk_anticipation.nut
Darxo: It's saved so that it can be displayed in the combat log when the damage is received in the next event-function.
I will add a comment explaining this variables existance.

LordMidas: I'm talking about the `local damageReceivedMultiplier`

mod_reforged/hooks/config/strings.nut
Darxo: I prefer the more verbose way of describing the effect, especially since this perk description has the space to spare for it.
Both of my bullet points convey their own important effect and this separation makes them more readable.
I view this way of formatting as an exception that we can sometimes employ if we see it fit.

LordMidas: Yeah because we're keeping the 10% tile per tile part as well, it's better to split it for now. But we have to say that the reduction is a percentage. The current statement is inaccurate i.e. it says the reduction is `equal to your Ranged Defense + an additional 10%...`. Whereas it should say the reduction is `equal to a percentage equal to your current Ranged Defense + an additional 10%...`. 

Please replace `+` with `and` or `plus`. It's cleaner to keep tooltips free of mathematical symbols and that's in line with the other tooltips too.

mod_reforged/hooks/skills/perks/perk_anticipation.nut
Darxo: fixed

LordMidas (approved):

PR #142, opened 2023-07-30
refactor: separate name and title with newline in combat tooltip

Darxo:
Currently every actor that a title will have that one displayed alongside their names on the combat tooltip.

Problem:
Both strings are stuffed into the same line overflowing to the next one.
A brother called **Basti** with the title **Nimble Forge Angler** will have it displayed the following:

``` 
Basti Nimble Forge
Angler
```

This PR will add a newline in such cases so that the title is displayed on the second line.
This should improve readability on the battlefield. For example to quickly identify fodder brothers.

So the example above would be displayed like so:
``` 
Basti 
Nimble Forge Angler
```

### DIFF
--- a/mod_reforged/hooks/entity/tactical/actor.nut
+++ b/mod_reforged/hooks/entity/tactical/actor.nut
@@ -59,6 +59,12 @@
 		if (this.isDiscovered() == false) return ret;
 		if (this.isHiddenToPlayer()) return ret;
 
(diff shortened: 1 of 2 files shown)


LordMidas:
Can you please show a screenshot of before/after. It's difficult for me to visualize the importance of what this PR is solving without viewing it in action.

PR #144, opened 2023-08-16
refactor(balance): half the relative and minimum recovery rate for mood

Darxo:


Darxo:
My first brief test (over like 15 ingame days) with this feature had all my brothers "permanently" happy.

This is not a indicator whether this change is good.
But it made me realise that positive effects will also last way longer to a point where we also need to address sources of positive mood in order to achieve the intended results (bad mood being more impactful).

LordMidas:
Yeah I was thinking about it the other day that this might happen. I wanted
to suggest that only mood increase should be slower. Mood decrease should
remain at vanilla speed.

PR #146, opened 2023-08-21
Flip melee inventory images

Darxo:


LordMidas (changes requested):
The consistency that this PR aims to achieve will break down the moment the user adds another mod that adds items and doesn't follow this convention. And there is no way for us to enforce the usage of this convention on other modders. Therefore, I'd not make this part of Reforged.

PR #149, opened 2023-08-28
feat: add item tables

Uberbagel:


LordMidas (changes requested):
The ItemTables of named items should be populated dynamically using the lists of vanilla named items. We should look into how to do that effectively.

LordMidas:
For now merge it in as is. Once I am back I can figure out a way to automatically populate the item tables for named items.

LordMidas (approved):

PR #152, opened 2023-08-29
feat: implement numeral adjectives

Darxo:
This will implement this mod of mine into reforged: https://www.nexusmods.com/battlebrothers/mods/530

The Idea:
- World Parties do not show their exact number anymore, instead they use the same system as Troop-Compositions (e.g. in Combat Dialog)
- We introduce new indirect numeral adjectives
- We define new ranges for all numeral adjectives where it makes sense

But unlike my mod I linked
- We don't give the player an option to display the exact number by default but we support it still. So certain game mechanics in reforged or submods can always turn it back on.

### DIFF
--- a/mod_reforged/config/indirect_numeral_adjectives.nut
+++ b/mod_reforged/config/indirect_numeral_adjectives.nut
@@ -1,0 +1,202 @@
+::Reforged.Numerals <- {
+	// Config
+	ExactPartySize = false,
+	ExactDialogSize = false,
+	PartyRange = false,
+	DialogRange = true
+
+	// Data
+	IndefiniteNumerals = [],
+
+	function register( _name, _startsAt, _suffix = "" )
+	{
+		::Reforged.Numerals.IndefiniteNumerals.push({
+			Name = _name,
+			StartsAt = _startsAt,
+			Suffix = _suffix,
+			Range = ""
+		});
+	}
+
+	function setExactPartySize( _bool )
+	{
+		this.ExactPartySize = _bool;
+		this.updateAllParties();
+	}
+
+	function setPartyRange( _bool )
+	{
+		this.PartyRange = _bool;
+		this.updateAllParties();
+	}
+
+
+	// This always needs to be called after all new numerals of your mod are registered.
+	// This will sort all Numerals and generate their ranges automatically
+	function calculateRanges()
+	{
+		this.IndefiniteNumerals.sort(@(a, b) a.StartsAt <=> b.StartsAt);
+
+		local minAmount = 1;
+		local maxAmount = 1;
+
+		for (local i = 1; i < this.IndefiniteNumerals.len(); i++)
+		{
+			maxAmount = (this.IndefiniteNumerals[i].StartsAt - 1);
+			if (minAmount != maxAmount || minAmount != 1)	// We hard-code that the (1) for One would never displayed
+			{
+				this.IndefiniteNumerals[i - 1].Range = "" + minAmount;
+				if (minAmount != maxAmount) this.IndefiniteNumerals[i - 1].Range += "-" + maxAmount;
+			}
+			minAmount = maxAmount + 1;
+		}
+		this.IndefiniteNumerals.top().Range = minAmount + "+";
+	}
+
+	function getNumeralString( _amount, _isWorldParty )
+	{
+		if (this.IndefiniteNumerals.len() == 0) return _amount;
+		if (this.ExactPartySize && _isWorldParty) return _amount;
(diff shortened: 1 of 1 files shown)


mod_reforged/config/indirect_numeral_adjectives.nut
```diff
@@ -0,0 +1,202 @@
+::Reforged.Numerals <- {
```
LordMidas, line 1: I'd change this to `::Reforged.TroopNumerals`

mod_reforged/config/indirect_numeral_adjectives.nut
```diff
@@ -0,0 +1,202 @@
...
+		this.ExactPartySize = _bool;
+		this.updateAllParties();
+	}
+
+	function setPartyRange( _bool )
+	{
+		this.PartyRange = _bool;
+		this.updateAllParties();
+	}
+
```
LordMidas, line 32: Extra line can be removed

mod_reforged/config/indirect_numeral_adjectives.nut
```diff
@@ -0,0 +1,202 @@
...
+
+	// This always needs to be called after all new numerals of your mod are registered.
+	// This will sort all Numerals and generate their ranges automatically
+	function calculateRanges()
```
LordMidas, line 36: One can make it so that the numerals must be registered during mod queueing. Which then means that this `calculateRanges()` can be called by us automatically during msu `AfterQueue` (or the AfterQueue bucket in Modern Hooks once we switch to that). Then modders don't have to call it.

mod_reforged/config/indirect_numeral_adjectives.nut
```diff
@@ -0,0 +1,202 @@
...
+		local numeralString = numeralEntry.Name;
+		if (!_isWorldParty) numeralString += numeralEntry.Suffix;
+		if (this.PartyRange && _isWorldParty || this.DialogRange && !_isWorldParty)
+		{
+			if (numeralEntry.Range != "") numeralString += " (" + numeralEntry.Range + ")";
+		}
+		return numeralString;
+	}
+
+	// Can be used by other mods to adjust individuell entries
```
LordMidas, line 73: English: `individual`

mod_reforged/config/indirect_numeral_adjectives.nut
```diff
@@ -0,0 +1,202 @@
...
+	function getNumeralEntryByName( _name )
+	{
+		foreach (numeral in this.IndefiniteNumerals)
+		{
+			if (numeral.Name == _namet) return numeral;
```
LordMidas, line 78: typo: `_name`

mod_reforged/config/indirect_numeral_adjectives.nut
```diff
@@ -0,0 +1,202 @@
...
+		}
+		return null;
```
LordMidas, line 80: `return null` is the default return of every function so this line is redundant.

mod_reforged/config/indirect_numeral_adjectives.nut
```diff
@@ -0,0 +1,202 @@
...
+	// Returns an indirect numeral, given an _amount. This requires 'this.IndefiniteNumerals' to be sorted
+	function getNumeralEntryByAmount( _amount )
+	{
+		local arrayPosition = -1;   // Somewhere the integer needs to be translated into an array position (decrementing by 1) so why not here
+		foreach (numeral in this.IndefiniteNumerals)
+		{
+			if (_amount < numeral.StartsAt) break;		// This should never happen for the first entry
+			++arrayPosition;
+		}
+		arrayPosition = ::Math.min(arrayPosition, this.IndefiniteNumerals.len() - 1);
```
LordMidas, line 102: Because `arrayPosition` is initialized as `-1`, I don't think it can ever be greater than  `this.IndefiniteNumerals.len() - 1`, so this line seems redundant.

LordMidas (changes requested):
I have some minor comments. I think @Enduriel should take a look at the code too.

Enduriel:
I'll take a look at how I'd implement this myself, but at first glance for what we are achieving this seems overcomplicated. I'd rather have a toggleable (via MSU boolean setting) system that just appends/patches the numerical information to relevant strings dynamically. Then if another mod wants to add this full functionality (like yours) it should force disable the reforged setting and then handle it itself :shrug:

Darxo:
As a sidenote, the current implementation also overwrites the vanilla array permanently.
That will cause issues when other mods call it interdependently for their own uses.

It's a rather simple addition to fix this compatibility issue. I just didn't bother doing it in the first prototype

PR #161, opened 2023-09-16
refactor(balance): add initiative bonus for falcon accessory

Darxo:


mod_reforged/hooks/items/accessory/falcon_item.nut
```diff
@@ -0,0 +1,12 @@
+::mods_hookExactClass("items/accessory/falcon_item", function(o) {
+	o.m.InitiativeModifier <- 15;
```
LordMidas, line 2: The naming of this field may conflict with the Weight/Burden rework we were working on where I suggested an `InitiativeModifier` field for items in line with `StaminaModifier`. Therefore, I would suggest to rename this to `InitiativeBonus`.

Consider approved after the rename.

LordMidas (approved):

PR #162, opened 2023-09-16
refactor(balance): reduce damage reduction of Anticipation by 10%

Darxo:
Fix Anticipation to work like it was initially envisioned to grant no natural damage reduction against adjacent entities.

### DIFF
--- a/mod_reforged/hooks/skills/perks/perk_anticipation.nut
+++ b/mod_reforged/hooks/skills/perks/perk_anticipation.nut
@@ -52,7 +52,7 @@
 				id = 11,
 				type = "text",
 				icon = "ui/icons/ranged_defense.png",
-				text = ::MSU.Text.colorRed(this.m.DamageReductionPerTile + "%") + " additional reduced damage received from all attacks for every tile distance between the attacker and you"
+				text = ::MSU.Text.colorRed(this.m.DamageReductionPerTile + "%") + " additional reduced damage received from all attacks for every tile between the attacker and you"
 			},
 			{
 				id = 15,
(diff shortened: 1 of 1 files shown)


mod_reforged/hooks/skills/perks/perk_anticipation.nut
```diff
@@ -99,12 +99,17 @@
...
 		damageReduction += ::Math.max(0, this.getContainer().getActor().getCurrentProperties().getRangedDefense());
-		damageReduction += distanceToAttacker * this.m.DamageReductionPerTile;
+
+		local tilesBetweenEntities = 0;
+		if (_attacker != null)
+		{
+			tilesBetweenEntities = _attacker.getTile().getDistanceTo(this.getContainer().getActor().getTile());
+			tilesBetweenEntities--;		// The tiles between adjacent entities is 0. So it is always 1 lower than their tile distance.
+		}
+		damageReduction += tilesBetweenEntities * this.m.DamageReductionPerTile;
```
LordMidas, line 111: Lines 105 to 111 are unnecessarily verbose. Can be squashed into a single line which still has good readability:
```squirrel
damageReduction += this.m.DamageReductionPerTile * (_attacker == null ? 0 : _attacker.getTile().getDistanceTo(this.getContainer().getActor().getTile()) - 1); // -1 because adjacent entities count at a distance of 1 and we want only the number of tiles between the entities
```
Putting the `this.m.DamageReductionPerTile` variable first in the multiplication makes it clear that the next expression enclosed in `()` is the number of tiles.

LordMidas (changes requested):

mod_reforged/hooks/skills/perks/perk_anticipation.nut
LordMidas: Actually, on second thought, any time we access tiles and distances we need to do `isPlacedOnMap` checks to be sure that there won't be a crash. Even then you can squash this all into 3 lines:
```squirrel
if (_attacker != null && _attacker.isPlacedOnMap() && this.getContainer().getActor().isPlacedOnMap())
{
   damageReduction += this.m.DamageReductionPerTile * (_attacker.getTile().getDistanceTo(this.getContainer().getActor().getTile()) - 1); // -1 because adjacent entities count at a distance of 1 and we want only the number of tiles between the entities
}
```

Darxo: > Lines 105 to 111 are unnecessarily verbose. Can be squashed into a single line which still has good readability:

I had the `- 1` in the same line at first but quickly decided that this little detail is easily overlooked and the resulting line with the comment is way too long.
That's why I went for multiple lines and I'm still in favor of the line 111

LordMidas (approved):

PR #163, opened 2023-09-16
feat: implement new agents for fast- and leader bandits

Darxo:
Change Leaders to stay closer to their allies more often. 
Change Fast Bandits to flank more often


LordMidas (commented):
Code and design looks good to me. I don't think we need the `bandits` subfolder within `agents` though. Vanilla puts all agents directly in the `agents` folder and we should do that too imo.

Darxo:
> Code and design looks good to me. I don't think we need the `bandits` subfolder within `agents` though. Vanilla puts all agents directly in the `agents` folder and we should do that too imo.
> 
> Please also include a commit where you update the relevant entities to use these new agents.

Yes, Vanilla throws all agents into the same folder, which is far more confusing that it has to be, especially now that we add many more faction specific agents.
I would like to start using a simple folder structure for new agents that are clearly faction specific.

For Items and Events I get it that we wanna use the same folder structure so that other mods can instantly find our new additions. But for Agents I can't think of a scenario where this is useful. So this argument does not hold true in the case of new agents.

LordMidas:
- Agents are not necessarily always used for the same faction/entity. E.g. you can use the `bandit_leader_agent` on a new entity which isn't a bandit leader or isn't even a bandit at all but still you want similar AI for that entity. 
- I don't see a confusion because the agent's name is very descriptive in terms of what it is.
- In any case, your approach isn't necessarily "wrong" tbh. I'll say get @Enduriel's comment on it and then finalize it.

LordMidas (approved):

PR #168, opened 2023-09-20
Ai tweaks onslaught holdsteady

LordMidas:
The previous ai package for these skills wasn't even working properly due to some logical issues in the algorithm. I've tested this new implementation and it works. We can tweak the AI over time more based on feedback.

It now uses the number of enemies adjacent to relevant allies and checks the hit-chance of allies vs enemies to determine whether to hold steady or to onslaught. Basically if enemies have a much greater hitchance than allies, then usage of Hold Steady is preferred, and if allies have hitchance close to enemy hit-chance or better than enemy hitchance, then usage of Onslaught is preferred. In both cases at least 3 relevant allies are required, otherwise the skills won't be triggered.

### DIFF
--- a/scripts/ai/tactical/behaviors/ai_rf_hold_steady.nut
+++ b/scripts/ai/tactical/behaviors/ai_rf_hold_steady.nut
@@ -36,54 +36,52 @@ this.ai_rf_hold_steady <- ::inherit("scripts/ai/tactical/behavior", {
 
 		score = score * this.getFatigueScoreMult(this.m.Skill);
 		local myTile = _entity.getTile();
-		local allies = ::Tactical.Entities.getInstancesOfFaction(_entity.getFaction());
-		local enemies = ::Tactical.Entities.getHostileActors(_entity.getFaction());
 		local useScore = 0.0;
 		local numTargets = 0;
 
-		local relevantAllies = [];		
-
-		foreach( enemy in enemies )
+		foreach (ally in ::Tactical.Entities.getInstancesOfFaction(_entity.getFaction()))
 		{
-			if (!enemy.hasZoneOfControl())
-			{
-				continue;
-			}
-			
-			local bestTarget = this.queryBestMeleeTarget(enemy, null, allies).Target;			
-			if (bestTarget == null || bestTarget.getMoraleState() == ::Const.MoraleState.Fleeing || bestTarget.getTile().getDistanceTo(myTile) > 4 || bestTarget.getSkills().hasSkill("effects.rf_hold_steady"))
+			local allyTile = ally.getTile();
+			if (allyTile.getDistanceTo(myTile) > 4 || !ally.hasZoneOfControl() || ally.getSkills().hasSkill("effects.rf_hold_steady"))
 			{
 				continue;
 			}
 
-			local enemyAttack = enemy.getSkills().getAttackOfOpportunity();
-			if (!enemyAttack.verifyTargetAndRange(bestTarget.getTile(), enemy.getTile()))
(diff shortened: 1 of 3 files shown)


scripts/ai/tactical/behaviors/ai_rf_hold_steady.nut
```diff
@@ -36,54 +36,51 @@ this.ai_rf_hold_steady <- ::inherit("scripts/ai/tactical/behavior", {
...
-			if (relevantAllies.find(bestTarget.getID()) == null) 
-			{
-				relevantAllies.push(bestTarget.getID());
+				local enemyHitChance = adjacentEntity.getSkills().getAttackOfOpportunity().getHitchance(ally);
+				if (enemyHitChance < 20) continue;
+
+				if (allyAttack.getHitchance(adjacentEntity) < enemyHitChance * 0.7)
+				{
+					numTargets++;
+					useScore += 40 * zocCount;
```
Darxo, line 73: there should be a break here. You dont want to count the same entity multiple times for each adjacent enemy

scripts/ai/tactical/behaviors/ai_rf_onslaught.nut
```diff
@@ -36,48 +36,51 @@ this.ai_rf_onslaught <- ::inherit("scripts/ai/tactical/behavior", {
...
 
-			numTargets++;
-			useScore = useScore + hitChance - 40;
+				local allyHitChance = allyAttack.getHitchance(adjacentEntity);
+				if (allyHitChance < 20) continue;
+
+				if (allyHitChance > adjacentEntity.getSkills().getAttackOfOpportunity().getHitchance(ally) * 0.7)
+				{
+					numTargets++;
+					useScore += 40 * zocCount;
```
Darxo, line 73: same as above: a `break;` is missing

scripts/ai/tactical/behaviors/ai_rf_hold_steady.nut
```diff
@@ -36,54 +36,51 @@ this.ai_rf_hold_steady <- ::inherit("scripts/ai/tactical/behavior", {
...
 		local numTargets = 0;
 
-		local relevantAllies = [];		
-
-		foreach( enemy in enemies )
+		foreach (ally in ::Tactical.Entities.getInstancesOfFaction(_entity.getFaction()))
 		{
-			if (!enemy.hasZoneOfControl())
+			local allyTile = ally.getTile();
+			if (allyTile.getDistanceTo(myTile) > 4 || !ally.hasZoneOfControl() || ally.getSkills().hasSkill("effects.rf_hold_steady"))
```
Darxo, line 45: Hold Steady grants defensive stats so it shouldnt matter whether that ally emits zone of control. They could have a bow or be sleeping. Both of these scenarios can still receive the buff and profit from it.

Darxo (changes requested):
the hasZoneOfControl check doesn't have to change. It a very minor detail that makes the code a bit more complicated.

The missing breaks are problematic though

scripts/ai/tactical/behaviors/ai_rf_hold_steady.nut
LordMidas: Technically you are correct - however, for the AI specifically I'd prefer that they use this skill in situations that provide the best bang for the buck. Using it to add a little bit of Defense to ranged units is probably still going to end up in those ranged units dying. So, for the AI specifically I want it to only consider melee units who are engaged with enemy and are in the "fighting" line.

Darxo (approved):

PR #170, opened 2023-09-20
Feat display ambition icons

Darxo:
### Current State

- All Ambitions which can spawn in vanilla now have unique circle-icon associated which are downscaled to 50x50 pixel.
- Those come from Orientation, Backgrounds, Perks, Effects or Traitts; using mostly Vanilla but also some Reforged Artwork
- Duplicate Ambitions which just require a higher number have the same icon. This includes "Do Contracts" and "Do Southern Contracts".
- "Choose No Ambition" has no icon. This is intended

### Remaining Issues:

- Icons which come from Traits have a golden border while all other have a grey border. This may look disconnected.
- Reforged Icons which are used could change any time in the future making them not fit anymore with the Ambitions (same if reforged icons are removed)
- Weapon Mastery Ambition and Named Item Set Ambition currently have the exact same icon of Weapon Master Perk. This is because its a perfect choice for the former ambition while still being a good choice for the latter. And I just couldn't find an alternative for the Named Set one that comes close to Weapon Master Perk

### My Thoughts:

- We should not go as far as to create custom images for Ambitions, not even for new ones. I find that a waste of effort for such a minor feature. I'd rather accept that some icons are "acceptable" at best.
- The existing Icons that I chose may also not even be the best choice from the available icons. But we should rather improve these icons over time from player feedback or our own experience ingame.


mod_reforged/hooks/ambitions/ambitions/trade_ambition.nut
```diff
@@ -0,0 +1,7 @@
+::mods_hookExactClass("ambitions/ambitions/trade_ambition", function (o) {
+	local create = o.create;
+	o.create = function()
+	{
+		create();
```
LordMidas, line 5: Missing icon

LordMidas (changes requested):

Darxo:
A remaining issue is which Icon we are gonna show for (No Ambition) option and what we are doing for Ambitions from other mods which do not know about this feature.

The solution to those two cases should be part of this PR

Darxo:
> A remaining issue is which Icon we are gonna show for (No Ambition) option and what we are doing for Ambitions from other mods which do not know about this feature.
> 
> The solution to those two cases should be part of this PR

One solution for this is to not show an icon in those cases. And that behavior already exists in this PR.
I like that solution.

LordMidas:
> > A remaining issue is which Icon we are gonna show for (No Ambition) option and what we are doing for Ambitions from other mods which do not know about this feature.
> > The solution to those two cases should be part of this PR
> 
> One solution for this is to not show an icon in those cases. And that behavior already exists in this PR. I like that solution.

- For no ambition option, there should be no icon.
- For ambitions without any icon, I think we should have a default icon and that is the golden cup vanilla icon for ambitions. We can create a trait-like icon from that and use that. I can try to prepare one and discuss over at Discord.

LordMidas (approved):
I think the icon should be placed in `gfx/ui/icons/` folder and not directly in the `gfx/ui/` folder. After that move, consider this approved from my side.

LordMidas:
You should get rid of the commit that changes the Renown entry, as that feature is implemented already in development in another way.

mod_reforged/hooks/ambitions/ambition.nut
```diff
@@ -1,27 +1,41 @@
...
 ::Reforged.QueueBucket.Late.push(function() {
 	::Reforged.HooksMod.hook("scripts/ambitions/ambition", function(q) {
 		q.getButtonTooltip = @(__original) function()
 		{
 			local ret = __original();
-			foreach (i, entry in ret)
+			if (this.getRenownOnSuccess() != 0)
 			{
-				if (("text" in entry) && entry.text == "Your Renown will increase, which means higher pay for contracts and potentially unlocking new types of contracts.")
+				local ret = __original();
```
LordMidas, line 18: These changes to `getButtonTooltip` regarding Renown seem to have come into this PR by mistake. Please edit them out.

PR #171, opened 2023-09-20
feat: show exact renown reward on all ambitions

Darxo:


mod_reforged/hooks/ambitions/ambition.nut
```diff
@@ -0,0 +1,18 @@
...
+
+	local getButtonTooltip = o.getButtonTooltip;
+	o.getButtonTooltip = function()
+	{
+		local ret = getButtonTooltip();
+		foreach(entry in ret)
+		{
+			if (entry.text.find("Your Renown will increase, which means higher pay for contracts") != null)
+			{
+				entry.text = "Gain " + ::MSU.Text.colorGreen("+" + this.getRenownOnSuccess()) + " Renown";
```
LordMidas, line 12: Better to replace it by `Your Renown will increase by <AMOUNT>, which means higher pay for contracts`

LordMidas (changes requested):

mod_reforged/hooks/ambitions/ambition.nut
Darxo: For reference, this is the full vanilla text:

> "Your Renown will increase, which means higher pay for contracts and potentially unlocking new types of contracts."

I do not only want to show the exact number, I also want it to all fit into a single line and trim the "repetitive fat" of vanilla descriptions.

How about just:

Your Renown will increase by <AMOUNT>

?

LordMidas: I agree. Let's replace it with something like `The company's Renown will increase by AMOUNT`.

LordMidas (approved):

PR #174, opened 2023-09-21
Ai behaviors assignment following msu update

LordMidas:
Previous MSU build automatically added `AIBehaviorID` to all vanilla skills. This meant that the behavior for said skills was also automatically added to any entity who got the skill. This resulted in Zombies using stuff like Gash, Decapitate, Puncture etc. 

The recent MSU build fixed this by no longer adding `AIBehaviorID` to skills automatically. This means mods now have to add them manually, but then also have to care about which agents they don't want those behaviors to be added to.

### DIFF
--- a/mod_reforged/hooks/ai/tactical/agent.nut
+++ b/mod_reforged/hooks/ai/tactical/agent.nut
@@ -1,0 +1,17 @@
+::mods_hookNewObject("ai/tactical/agent", function (o) {
+	local addBehavior = o.addBehavior;
+	o.addBehavior = function( _behavior )
+	{
+		local obj = this;
+		while ("SuperName" in obj)
+		{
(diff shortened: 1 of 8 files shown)


mod_reforged/hooks/ai/tactical/agent.nut
```diff
@@ -0,0 +1,20 @@
+::mods_hookNewObject("ai/tactical/agent", function (o) {
+	local addBehavior = o.addBehavior;
+	o.addBehavior = function( _behavior )
+	{
+		local className;
```
Darxo, line 5: Why is this variable declared outside?
It looks like it's used only a single time during the if. You might not even need to use a variable at all here

Darxo (approved):
After you resolved my other comment however you feel this can be seen as approved

mod_reforged/hooks/ai/tactical/agent.nut
LordMidas: Done.

PR #176, opened 2023-09-22
refactor: add equal number of low and high rolls to two star talents

LordMidas:
Instead of pure randomization. This makes it so that the total attribute change over the course of 10 levels remains the same as vanilla behavior (i.e. equal to a max roll every level). However, it adds randomization to some rolls: the number of rolls with -1 will be equal to the number of rolls with +1.

### DIFF
--- a/mod_reforged/hooks/entity/tactical/player.nut
+++ b/mod_reforged/hooks/entity/tactical/player.nut
@@ -149,7 +149,7 @@
 				local numIndicesToRandomize = ::Math.rand(2, _amount);
 				if (numIndicesToRandomize % 2 != 0) numIndicesToRandomize--; // Ensure that we have an even number to randomize
 
-				local change = ::Math.rand(-1, 1);
+				local change = 1;
 				for (local j = 0; j < numIndicesToRandomize; j++)
 				{
 					this.m.Attributes[i][indices.remove(::Math.rand(0, indices.len() - 1))] += change;
--- a/mod_reforged/hooks/entity/tactical/player.nut
+++ b/mod_reforged/hooks/entity/tactical/player.nut
@@ -24,12 +24,9 @@
 		{
 			if (attribute == ::Const.Attributes.COUNT) continue;
 
-			local attributeMin = ::Const.AttributesLevelUp[attribute].Min;
-			if (this.m.Talents[attribute] >= 1) attributeMin++;
-			if (this.m.Talents[attribute] == 3) attributeMin++;
-
+			local attributeMin = ::Const.AttributesLevelUp[attribute].Min + ::Math.min(this.m.Talents[attribute], 2);
 			local attributeMax = ::Const.AttributesLevelUp[attribute].Max;
-			if (this.m.Talents[attribute] >= 2) attributeMax++;
+			if (this.m.Talents[attribute] == 2) attributeMax++;
 
 			local levelUpsRemaining = ::Math.max(::Const.XP.MaxLevelWithPerkpoints - this.getLevel() + this.getLevelUps(), 0);
 			local attributeValue = attributeName == "Fatigue" ? baseProperties["Stamina"] : baseProperties[attributeName]; // Thank you Overhype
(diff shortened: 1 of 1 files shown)


Uberbagel (approved):

mod_reforged/hooks/entity/tactical/player.nut
```diff
@@ -24,12 +24,9 @@
...
 			if (attribute == ::Const.Attributes.COUNT) continue;
 
-			local attributeMin = ::Const.AttributesLevelUp[attribute].Min;
-			if (this.m.Talents[attribute] >= 1) attributeMin++;
-			if (this.m.Talents[attribute] == 3) attributeMin++;
-
+			local attributeMin = ::Const.AttributesLevelUp[attribute].Min + ::Math.min(this.m.Talents[attribute], 2);
 			local attributeMax = ::Const.AttributesLevelUp[attribute].Max;
-			if (this.m.Talents[attribute] >= 2) attributeMax++;
+			if (this.m.Talents[attribute] == 2) attributeMax++;
```
Darxo, line 29: What about 3 Stars? They can roll up to +1 maximum attribute. You seem to have removed that part now

LordMidas: Ah my bad, that's supposed to be `== 3`. I'll fix.

LordMidas:
Unfortunately, with this change the projected attributes won't be correct if you skip a low/high rolled level-up on a 2-star attribute. So I'll have to adjust the projected attributes calculation.

mod_reforged/hooks/entity/tactical/player.nut
```diff
@@ -129,16 +134,26 @@
...
+				}
+				local numIndicesToRandomize = ::Math.rand(2, _amount);
+				if (numIndicesToRandomize % 2 != 0) numIndicesToRandomize--; // Ensure that we have an even number to randomize
+
+				local change = 1;
+				for (local j = 0; j < numIndicesToRandomize; j++)
 				{
-					this.m.Attributes[i][j] += ::Math.rand(-1, 1);
+					this.m.Attributes[i][indices.remove(::Math.rand(0, indices.len() - 1))] += change;
+					change *= 1;
```
Darxo, line 156: You are not inverting the `change` variable correctly here

mod_reforged/hooks/entity/tactical/player.nut
```diff
@@ -129,16 +134,26 @@
...
 		{
 			if (this.m.Talents[i] == 2)
 			{
-				for (local j = 0; j < _amount; j++)
+				local indices = array(_amount);
+				foreach (j, _ in indices)
+				{
+					indices[j] = j;
+				}
+				local numIndicesToRandomize = ::Math.rand(2, _amount);
```
Darxo, line 149: This means that a brother can never have only +3 level-ups.
But on the flip-side a brother can have __no__ +3 level-ups.

You probably want to have some hardcoded fluff to have a small chance to roll all +3.

mod_reforged/hooks/entity/tactical/player.nut
```diff
@@ -24,16 +24,21 @@
...
+			if (this.m.Talents[attribute] == 3) attributeMax++;
 
 			local levelUpsRemaining = ::Math.max(::Const.XP.MaxLevelWithPerkpoints - this.getLevel() + this.getLevelUps(), 0);
 			local attributeValue = attributeName == "Fatigue" ? baseProperties["Stamina"] : baseProperties[attributeName]; // Thank you Overhype
 
+			if (this.m.Talents[attribute] == 2)
+			{
+				foreach (value in this.m.Attributes[attribute])
+				{
+					attributeValue += value - attributeMax;
```
Darxo, line 38: I feel like this will sometimes spoil a player. At the 10th level-up their projected tooltip will tell them exactly what roll they are gonna get with the 11th level-up.
I don't know a good solution for this. 

But on the other hand, this system is already somewhat deterministic, so if the player counted correctly or kept notes about all level-ups they would also know what the last level-up must be.

Darxo (changes requested):
Needs some adjustments

LordMidas:
Closing as we have implemented different design of randomized rolls for 2 star talents in #177

PR #178, opened 2023-09-25
feat: colorize stun sprites when duration is 2+ turns

Darxo:


mod_reforged/hooks/skills/effects/stunned_effect.nut
```diff
@@ -0,0 +1,56 @@
...
+	o.m.DoubleStunColor <- ::createColor("#ff6600");	// Orange
+
+	local setTurns = o.setTurns;
+	o.setTurns = function( _turns )
+	{
+		setTurns(_turns);
+		this.updateSprite();
+	}
+
+	o.onUpdate = function( _properties )
```
LordMidas, line 12: Is a complete overwrite of `onUpdate` really necessary? Maybe you let the original `onUpdate` run first and then just append `this.updateSprite()` to it? And in `updateSprite` you check if the sprite is visible, and if yes, then update color, otherwise return.

Darxo: I adjusted that now. Overwriting was not necessary.

scripts/skills/perks/perk_rf_concussive_strikes.nut
```diff
@@ -58,7 +58,7 @@ this.perk_rf_concussive_strikes <- ::inherit("scripts/skills/skill", {
 			{
 				local effect = ::new("scripts/skills/effects/dazed_effect");
 				_targetEntity.getSkills().add(effect);
-				effect.m.TurnsLeft = ::Math.max(1, 2 + this.getContainer().getActor().getCurrentProperties().NegativeStatusEffectDuration);
+				effect.setTurnsLeft(2);
```
LordMidas, line 61: `setTurns`

scripts/skills/perks/perk_rf_concussive_strikes.nut
```diff
@@ -69,7 +69,7 @@ this.perk_rf_concussive_strikes <- ::inherit("scripts/skills/skill", {
 		{
 			local effect = ::new("scripts/skills/effects/dazed_effect");
 			_targetEntity.getSkills().add(effect);
-			effect.m.TurnsLeft = ::Math.max(1, 1 + this.getContainer().getActor().getCurrentProperties().NegativeStatusEffectDuration);
+			effect.setTurnsLeft(1);
```
LordMidas, line 72: `setTurns`

mod_reforged/hooks/skills/effects/dazed_effect.nut
```diff
@@ -0,0 +1,51 @@
...
+	}
+
+// New Functions
+	// Initialize the TurnsLeft with a new value. Will always set the Turns to atleast 1. This function already exists for stunned_effect
+	o.setTurns <- function( _amount )
+	{
+		this.m.TurnsLeft = _amount;
+		if (this.getContainer() != null)
+		{
+			this.m.TurnsLeft += this.getContainer().getActor().getCurrentProperties().NegativeStatusEffectDuration;
```
LordMidas, line 18: This is a simple operation, but the code is too long. I suggest the following two options (I prefer the ternary):
```squirrel
if (this.getContainer() == null)
{
   this.m.TurnsLeft = _amount
}
else
{
   `this.m.TurnsLeft = ::Math.max(1, _amount + this.getContainer().getActor().getCurrentProperties().NegativeStatusEffectDuration)`
}
this.updateSprite();
```
I would've used a ternary, but it's up to you:
```squirrel
this.m.TurnsLeft = this.getContainer() == null ? _amount : ::Math.max(1, _amount + this.getContainer().getActor().getCurrentProperties().NegativeStatusEffectDuration);
this.updateSprite();
```

LordMidas (changes requested):

mod_reforged/hooks/skills/effects/dazed_effect.nut
Darxo: Your solution has less but longer lines. I don't see where this would be a clear improvement. Just seems like a side-grade depending on preference.
In this case the function body is already very small so I prefer using a few extra lines for each step.

Your suggestion is missing a `::Math.max(1, _amount)` call for the path where getContainer is null

LordMidas (approved):

LordMidas:
Please adapt the code to Modern Hooks style and then merge.

Darxo:
I don't have a setup to test a modern hooks implementation currently.
And I don't want to switch until I know more about Modern Hooks issue with the scrambled mod hook order

So I will wait with the rewrite until then.

PR #184, opened 2023-10-02
Bandit icons

Uberbagel:


mod_reforged/hooks/items/accessory/accessory.nut
```diff
@@ -0,0 +1,30 @@
+::mods_hookExactClass("items/accessory/accessory", function(o) {
```
Darxo, line 1: Please drop this commit. It should already have been merged into development

It somehow sneaked into your PR but has nothing to do with it.

Darxo (changes requested):
Your code looks good.
Just my commit being in there for something does not

mod_reforged/hooks/items/accessory/accessory.nut
Uberbagel: fixed

Darxo (approved):

LordMidas (changes requested):
For each unit block in bandit_blocks.nut you must also add `DeterminesFigure = true`.

Uberbagel:
Hopefully I have done it correctly.

LordMidas (approved):
Looks good. Please squash the two figure-related commits into one.

LordMidas (approved):

scripts/mods/mod_reforged/dynamic_spawns/units/bandit_units.nut
```diff
@@ -7,6 +7,18 @@ figure_bandit_04 - Bandit Leader
 
 // We add our new bandit units
 local units = [
+	{
+		ID = "Bandit.Thug",
```
LordMidas, line 11: No, this is not the way to change figures of vanilla units. You should change them the same way you change their Cost i.e. `::DynamicSpawns.Units.findById("xyz").Figure = "abc"`.

The reason the current implementation doesn't work is because this is a NEW unit being registered with DSF. And it is missing the `Troop` key.

LordMidas: Unfortunately I merged the PR first and then realized this mistake. So I deleted the merged commits and restored this branch. This is gonna cause GitHub issues whereby we can't merge this PR again. But that's ok, we will cherry pick the commits onto the development branch.

LordMidas: Basically, I'd like that DSF no longer has its own unit/unitblocks/parties definitions and we port them all to Reforged. Then we won't run into this issue as all the units will be defined by us.

PR #186, opened 2023-10-10
Feat damage based xp system

LordMidas:
I have left many `logInfo` in the code for now so you can test the PR on your own if you want. These should be removed before merging.

- XP is now awarded based on the percentage of total damage dealt to an entity.
- Loot should now drop if the player faction did at least 50% of damage. But this is implemented in a hacky way, and we should really think this through if we want to do it like this, or if there is a better solution.

XP sharing

### DIFF
--- a/mod_reforged/hooks/entity/tactical/actor.nut
+++ b/mod_reforged/hooks/entity/tactical/actor.nut
@@ -1,11 +1,18 @@
 ::Reforged.HooksMod.hook("scripts/entity/tactical/actor", function(q) {
 	q.m.IsPerformingAttackOfOpportunity <- false;
 	q.m.IsWaitingTurn <- false;		// Is only set true when using the new Wait-All button. While true this entity will try to use Wait when its their turn
+	q.m.RF_DamageReceived <- null; // Table with faction number as key and tables as values. These tables have actor ID as key and the damage dealt as their value. Is populated during skill_container.onDamageReceived
 
 	q.isDisarmed <- function()
 	{
 		local handToHand = this.getSkills().getSkillByID("actives.hand_to_hand");
 		return handToHand != null && handToHand.isUsable();
+	}
+
+	q.create = @(__original) function()
+	{
+		__original();
+		this.m.RF_DamageReceived = { Total = 0.0 };
 	}
 
 	q.onInit = @(__original) function()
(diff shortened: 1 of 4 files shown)


mod_reforged/hooks/entity/tactical/actor.nut
```diff
@@ -165,3 +174,73 @@
 	{
 	}
 });
+
+::Reforged.HooksMod.hookTree("scripts/entity/tactical/actor", function(q) {
+	q.onActorKilled = @( __original ) function( _actor, _tile, _skill )
+	{
+		::Reforged.Config.XPOverride = true;
+		__original(_actor, _tile, _skill);
+		::Reforged.Config.XPOverride = false;
```
Darxo, line 183: This could become tricky when certain actors deal damage during this function which might cause additional onDeath functions for other actors to trigger.

The `flying_skull` from vanilla does this.
If they die and kill someone else with their damage then a nested onDeath is called. When that one finishes then the XPOverride is false but the first onDeath is still running and no longer applying our changed XP policy correctly.

On a side note. The Flying Skulls also cause other bugs due to their onDeath effects. For example you can cheese the Sunken Library by having your last brother die to a flying skull. Since the Skull was the last one to die (an enemy) you are attributed the win of this battle.

mod_reforged/hooks/skills/skill_container.nut
```diff
@@ -0,0 +1,9 @@
+::Reforged.HooksMod.hook("scripts/skills/skill_container", function(q) {
+	q.onDamageReceived = @(__original) function( _attacker, _damageHitpoints, _damageArmor )
+	{
+		__original(_attacker, _damageHitpoints, _damageArmor);
+
+		this.getActor().m.RF_DamageReceivedFromFaction.add(_attacker.getFaction(), _damageHitpoints + _damageArmor);
```
Darxo, line 6: We might want to limit these values so that overkill damage is not counted towards exp/loot. Otherwise Kill-Stealing is kind of back on the menue.
If I deal 99% of the damage to an entity and an allied billman kills them for like 30% overkill damage then they would "steal" 25% of the exp from me.

LordMidas: Good point. I've fixed this now.

Darxo (approved):

Darxo:
I haven't tested this out ingame but the code looks good.
If it's up to me you can remove the logs again.

LordMidas:
> I haven't tested this out ingame but the code looks good. If it's up to me you can remove the logs again.

Thanks. I will change it to use tables instead of MSU WeightedContainer, and then merge.

mod_reforged/hooks/entity/tactical/actor.nut
```diff
@@ -152,7 +150,7 @@
 
 	q.isTurnDone = @(__original) function()
 	{
-		if (::Tactical.getNavigator().isTravelling(this)) return false;		// Copy&Paste of check from vanilla
+		if (::Tactical.getNavigator().isTravelling(this)) return false;		// Copy&Paste of check  vanilla
```
Darxo, line 153: weird choice of removing "from" here and leaving 2 spaces in the comment

LordMidas: That was a mistake that somehow got into the commit. I've removed that change from this PR.

mod_reforged/hooks/entity/tactical/actor.nut
```diff
@@ -165,3 +172,62 @@
...
+		// - This is divided into Killer XP and Group XP.
+		// - The brothers who did damage are considered "Killers". They share the Killer XP based on the
+		// percentage of damage done by each bro relative to other bros who did damage.
+		// - The group XP is then shared equally among all bros in the company.
+
+		local bros = ::Tactical.Entities.getInstancesOfFaction(::Const.Faction.Player);
+
+		if (::Const.Faction.Player in this.m.RF_DamageReceived)
+		{
+			local XPavailable = this.getXPValue() * this.m.RF_DamageReceived[::Const.Faction.Player].Total / this.m.RF_DamageReceived.Total;
```
Darxo, line 226: I would add a check whether `this.m.RF_DamageReceived.Total == 0`.

Currently it is possible that that variable is 0.0 when a character receives 0 total damage. And if they are then killed off by something other than damage (console command/future alternative deaths?) the game crashes

Darxo: Nevermind. the problem has to be solved elsewhere

mod_reforged/hooks/skills/skill_container.nut
```diff
@@ -0,0 +1,22 @@
+::Reforged.HooksMod.hook("scripts/skills/skill_container", function(q) {
+	q.onDamageReceived = @(__original) function( _attacker, _damageHitpoints, _damageArmor )
+	{
+		local damage = _damageArmor + ::Math.min(_damageHitpoints, this.getActor().getHitpoints());
+
+		__original(_attacker, _damageHitpoints, _damageArmor);
+
+		if (_attacker != null)
```
Darxo, line 8: We should also make sure that `damage != 0` here. Otherwise we write `0.0` in various variables which could later cause a divided by 0

Darxo (commented):
I really like the implementation and the staying close to the vanilla calculation

mod_reforged/hooks/skills/skill_container.nut
LordMidas: Good catch. The PR was already merged by this point. I fixed this in the commit on development branch and force pushed that.

PR #187, opened 2023-10-10
Fix missing icons for vanilla skills

LordMidas:


Darxo (approved):
We should also provide a default Icon for all skills (+ sw version) in the skill.nut.
Modder will not know that suddenly enemy skills need an icon because we display them. So that would be to get ahead of those cases.

I imagine something like `status_effect_107.png`
status_effect_107

I leave it to you whether this addition should be part of this PR.

PR #188, opened 2023-10-14
feat: allow leveling up indefinitely

LordMidas:
In vanilla the `::Const.LevelXP` array has a length of 33 and the `addXP` function does an early return if the character's level is equal to or greater than its length. This means that characters stop gaining XP at level 33. In this hook we expand the array as necessary when the level value reaches its length. The calculation used for adding the XP requirement is taken from the vanilla calculation.


Darxo (approved):
Looks good

PR #189, opened 2023-10-14
refactor: add an offset for location banner to prevent an overlap

Darxo:
This change was meant to affect all locations but during my tests only settlements were affected. But that is not even bad because the Settlements were the most important ones to fix anyways.

I haven't tested this exact implementation yet. 
The test I showed you didn't utilize a central `adjustBannerOffset` and wasn't based on Modern Hooks.

### DIFF
--- a/mod_reforged/hooks/entity/world/location.nut
+++ b/mod_reforged/hooks/entity/world/location.nut
@@ -1,7 +1,28 @@
 ::Reforged.HooksMod.hook("scripts/entity/world/location", function(q) {
+	q.setBanner = @(__original) function( _banner )
+	{
+		__original(_banner);
+		this.adjustBannerOffset();
+	}
+
+	q.onAfterInit = @(__original) function()
+	{
+		__original();
+		this.adjustBannerOffset();
+	}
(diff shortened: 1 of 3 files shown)


mod_reforged/hooks/entity/world/location.nut
```diff
@@ -1,7 +1,29 @@
 ::Reforged.HooksMod.hook("scripts/entity/world/location", function(q) {
+
```
LordMidas, line 2: We don't put extra lines at the start of hooks. So please remove this.

mod_reforged/hooks/entity/world/settlement.nut
```diff
@@ -1,4 +1,17 @@
 ::Reforged.HooksMod.hook("scripts/entity/world/settlement", function(q) {
+
```
LordMidas, line 2: Same with this extra line.

LordMidas (approved):
Looks good to me. Just some minor style changes to extra lines. If you can test it and it works like you showed me on Discord then consider it approved.

PR #190, opened 2023-10-16
fix: add description and tooltip for sneak attack perk's effect part

LordMidas:


scripts/skills/perks/perk_rf_sneak_attack.nut
```diff
@@ -17,6 +17,78 @@ this.perk_rf_sneak_attack <- ::inherit("scripts/skills/skill", {
...
+					id = 10,
+					type = "text",
+					icon = "ui/icons/direct_damage.png",
+					text = ::MSU.Text.colorGreen("+10%") + " armor penetration"
+				}
+			]);
+		}
+		else
+		{
+			tooltip.extend([
```
Darxo, line 49: I would check the main weapon at this point and print a warning if its reach is more than 4 instead of the following tooltip

LordMidas: The effect is hidden unless a valid weapon is equipped. Which is why I didn't do that check here. It is done in `isEnabled`.

Darxo (approved):

PR #191, opened 2023-10-16
refactor: improve description of nimble and poise

Darxo:
The line about Brawny is already flawed when we also have Wears it Well that does the same effect. It's an artifact from Vanilla.
And with the "Weight Rework" it will be more redundant.

### This is how the perk descriptions look now:

image

image

### DIFF
--- a/mod_reforged/hooks/config/strings.nut
+++ b/mod_reforged/hooks/config/strings.nut
@@ -525,12 +525,10 @@ local vanillaDescriptions = [
  				Description = [
  					"Damage to [Hitpoints|Concept.Hitpoints] is reduced by " + ::MSU.Text.colorGreen("50%") + " and that to armor by " + ::MSU.Text.colorGreen("25%") + ".",
  					"The bonus drops exponentially when wearing head and body armor with a total penalty to [Maximum Fatigue|Concept.MaximumFatigue] above 15. The lighter your armor and helmet, the more you benefit.",
- 					"Does not affect damage from mental attacks or status effects, but can help to avoid receiving them.",
- 					"[Brawny|Perk+perk_brawny] does not affect this perk.",
- 					"Cannot be picked if you have [Poise|Perk+perk_rf_poise]."
-
+ 					"Does not affect damage from mental attacks or status effects."
  				]
- 			}]
+ 			}],
+			Footer = ::MSU.Text.colorRed("Cannot be picked if you have [Poise|Perk+perk_rf_poise].")
 	 	}),
 	},
 	{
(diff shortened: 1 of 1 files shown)


LordMidas:
From my experience in PTR, always assume that people can't understand even basic stuff, and design tooltips to compensate for this. We had a perk in PTR that worked off of head and armor fatigue penalty and we frequently got someone asking "does it work with Brawny?", that we ended up adding that line to the perk tooltip. So, I think removing these lines from Nimble and Poise is going to result in a similar problem. We can instead improve the tooltip and say that `"This perk is not affected by any perk that reduces the Fatigue Penalty of equipped gear e.g. Brawny"`.

Darxo:
I prefer almost anything over having to enumerate every perk, effect and item combination that a given perk does interact with weirdly.

The moment that you say, having Brawny in that perk text is important, we automatically also have to include Wears it well there.

LordMidas:
> I prefer almost anything over having to enumerate every perk, effect and item combination that a given perk does interact with weirdly.
> 
> The moment that you say, having Brawny in that perk text is important, we automatically also have to include Wears it well there.
> 
> Maybe the solution is adding a `base` or `original` somewhere in the Nimble and Poise tooltip so that it is more clear that they take the unchanged values as reference.

I agree with you that mentioning one perk is a problem because that doesn't cover every other perk My proposed sentence in the previous comment doesn't suffer from the issue whereby we have to mention every perk.

Darxo:
I want to avoid this extra bullet point and those two extra lines in these tooltips.
With the Weight term implemented this should be easily possible by mentioning "Combined Base Weight" and adding a nested tooltip for `Base Weight`.

How about the following for the meantime:
```The bonus drops exponentially when wearing head and body armor with a **combined base** penalty to Maximum Fatigue above 35.```

LordMidas:
> I want to avoid this extra bullet point and those two extra lines in these tooltips. With the Weight term implemented this should be easily possible by mentioning "Combined Base Weight" and adding a nested tooltip for `Base Weight`.
> 
> How about the following for the meantime:
> 
> * The bonus drops exponentially when wearing head and body armor with a **combined base** penalty to Maximum Fatigue above 35.

I understand your concern - the point is, as I mentioned in my first comment, that players don't understand that. I agree that a nested tooltip would be good. I am happy with your suggested line, but that will also require a nested tooltip for `base penalty to Maximum Fatigue`. And that nested tooltip should mention that `it is the penalty before the reduction from any perks e.g. Brawny etc.`

Darxo:
If someone asks how this works with Brawny despite the new term **base penalty** then I will gather some of my patience and tell them that the effect text is correct and that their answer lies in the effect text.

PR #192, opened 2023-10-17
Attribute bleed damage to last attacker when bleed was applied

LordMidas:
This ensures that the "bleeding_effect", no matter applied by which skill (as long as the skill is an attack), always keeps track of the attacker and attributes the bleeding damage to that attacker.

This is important for the damage-based XP system we now use in Reforged.

### DIFF
--- a/mod_reforged/hooks/skills/effects/bleeding_effect.nut
+++ b/mod_reforged/hooks/skills/effects/bleeding_effect.nut
@@ -1,0 +1,44 @@
+::Reforged.HooksMod.hook("scripts/skills/effects/bleeding_effect", function(q) {
+	q.m.Attacker <- null;
+
+	q.onAdded = @(__original) function()
+	{
+		__original();
+		if (!this.isGarbage())
+		{
+			this.m.Attacker = this.getContainer().getActor().getLastAttacker();
+		}
+	}
+
+	q.applyDamage = @() function()
+	{
+		if (this.m.LastRoundApplied != ::Time.getRound())
+		{
+			this.m.LastRoundApplied = ::Time.getRound();
+			local actor = this.getContainer().getActor();
+			this.spawnIcon("status_effect_01", actor.getTile());
+			local hitInfo = clone ::Const.Tactical.HitInfo;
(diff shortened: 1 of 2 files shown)


mod_reforged/hooks/skills/effects/bleeding_effect.nut
```diff
@@ -0,0 +1,44 @@
...
+			if (--this.m.TurnsLeft <= 0)
+			{
+				this.removeSelf();
+			}
+		}
+	}
+
+	q.getAttacker <- function()
+	{
+		if (!::MSU.isNull(this.m.Attacker) && this.m.Attacker.isAlive() && this.m.Attacker.isPlacedOnMap())
```
Darxo, line 37: Is `this.m.Attacker.isAlive()` important to prevent any crashes?

Right now:
If one of my brothers dies after applying a bleed their bleed wouldn't contribute towards my team gaining experience or potentially the loot of that enemy.

Darxo (changes requested):
I believe the `isAlive` check should be removed, though I'm not that deep into the consequences of that.

Also I was totally expecting you to also implement this new feature for spider poison when I saw this branch. What are your plans regarding that? While it is not as common the player still has access to it via the item.

mod_reforged/hooks/skills/effects/bleeding_effect.nut
LordMidas: I believe the `isAlive()` check is important because if this actor dies as as result of this damage, then it will trigger a chain of events such as `onActorKilled` and others for the killer. When I was working on Legends I saw that there were comments in the `bleeding_effect` code file saying that crashes were happening unless you add both `isAlive()` and `isPlacedOnMap()` checks.

> If one of my brothers dies after applying a bleed their bleed wouldn't contribute towards my team gaining experience or potentially the loot of that enemy.
This is a valid concern. I will try to think if we can solve this.

LordMidas:
> Also I was totally expecting you to also implement this new feature for spider poison when I saw this branch. What are your plans regarding that? While it is not as common the player still has access to it via the item.

I didn't even think of that. Yeah we should implement a similar solution for that - after we've figured out what we want to do with dead attacker.

LordMidas:
With the rework of the bleeding effect that is coming in version `0.5.0` this PR will need a rethink / re-implementation. So I am closing it for now.

PR #194, opened 2023-10-21
refactor(remove): all recover perks from human enemies

Darxo:
This skill is now granted to all humans by default so the perk is redundant.


mod_reforged/hooks/entity/tactical/humans/gunner.nut
Uberbagel, line 1: Why remove Power Shot on gunner?

Darxo: I don't know.

It didn't belong in this PR, I removed it now

Uberbagel (approved):

PR #195, opened 2023-10-22
fix: add veteran perks and hyperlinks to Experience and Level tooltips

LordMidas:
Fixes [this](https://discord.com/channels/1006908336991645757/1163905659834732594) bug.

### DIFF
--- a/mod_reforged/hooks/ui/screens/tooltip/tooltip_events.nut
+++ b/mod_reforged/hooks/ui/screens/tooltip/tooltip_events.nut
@@ -1,0 +1,37 @@
+::Reforged.HooksMod.hook("scripts/ui/screens/tooltip/tooltip_events", function(q) {
+	q.general_queryUIElementTooltipData = @(__original) function( _entityId, _elementId, _elementOwner )
+	{
+		switch (_elementId)
+		{
+			case "character-screen.left-panel-header-module.Experience":
+				return [
+					{
+						id = 1,
+						type = "title",
+						text = "Experience"
+					},
+					{
+						id = 2,
+						type = "description",
+						text = ::Reforged.Mod.Tooltips.parseString("Characters gain experience as they or their allies slay enemies in battles. If a character has accumulated sufficient experience, he\'ll [level up|Concept.Level].\n\nExperience gained from dying enemies is proportional to the fraction of the total damage dealt to that enemy by your company. Of this experience, " + (::Const.XP.XPForKillerPct * 100) + "% is shared proportionally between the brothers who did damage. The other " + (100 - ::Const.XP.XPForKillerPct * 100) + "% is shared equally among all members of the company.")
+					}
+				];
+
+			case "character-screen.left-panel-header-module.Level":
+				return [
+					{
+						id = 1,
+						type = "title",
+						text = "Level"
+					},
+					{
+						id = 2,
+						type = "description",
+						text = ::Reforged.Mod.Tooltips.parseString("The character\'s level measures [experience|Concept.Experience] in battle. Characters rise in levels as they gain experience and are able to increase their [attributes|Concept.CharacterAttribute] and gain [perks|Concept.Perk] that make them better at the mercenary profession.\n\nBeyond the " + ::Const.XP.MaxLevelWithPerkpoints + "th character level, characters are veterans and gain perk points only every 4th level. They can still improve their attributes but the attribute gain per level is small.")
+					}
+				];
+		}
+
+		return __original(_entityId, _elementId, _elementOwner);
+	}
+});


mod_reforged/hooks/ui/screens/tooltip/tooltip_events.nut
```diff
@@ -0,0 +1,37 @@
...
+				return [
+					{
+						id = 1,
+						type = "title",
+						text = "Experience"
+					},
+					{
+						id = 2,
+						type = "description",
+						text = ::Reforged.Mod.Tooltips.parseString("Characters gain experience as they or their allies slay enemies in battles. If a character has accumulated sufficient experience, he\'ll [level up|Concept.Level] and be able to increase his [attributes|Concept.CharacterAttribute] and pick a [perk|Concept.Perk] that grants a unique bonus.\n\nBeyond the 11th character level, characters are veterans and gain perk points only every 4th level. They can still improve their attributes but the attribute gain per level is small.")
```
Darxo, line 16: Now that we are rewriting this tooltip we should use the existing vanilla variable `::Const.XP.MaxLevelWithPerkpoints` instead of writing out the number 11.

This is one step closer to having a smooth transition for our mod or submods when they want to change this value.

mod_reforged/hooks/ui/screens/tooltip/tooltip_events.nut
```diff
@@ -0,0 +1,37 @@
...
+				return [
+					{
+						id = 1,
+						type = "title",
+						text = "Level"
+					},
+					{
+						id = 2,
+						type = "description",
+						text = ::Reforged.Mod.Tooltips.parseString("The character\'s level measures [Experience|Concept.Experience] in battle. Characters rise in levels as they gain experience and are able to increase their [attributes|Concept.CharacterAttribute] and gain [perks|Concept.Perk] that make them better at the mercenary profession.\n\nBeyond the 11th character level, characters are veterans and gain perk points only every 4th level. They can still improve their attributes but the attribute gain per level is small.")
```
Darxo, line 30: Same as above: 11 should be replaced with `::Const.XP.MaxLevelWithPerkpoints`

Darxo (changes requested):
We might aswell replace the hardcoded mentions of 11 with the existing ::Const.XP.MaxLevelWithPerkpoints if we are at it

mod_reforged/hooks/ui/screens/tooltip/tooltip_events.nut
LordMidas: This is a good point. I will adjust the code accordingly.

Darxo (approved):
Looks Good

mod_reforged/hooks/ui/screens/tooltip/tooltip_events.nut
```diff
@@ -13,7 +13,7 @@
...
+						text = ::Reforged.Mod.Tooltips.parseString("Characters gain experience as they or their allies slay enemies in battles. If a character has accumulated sufficient experience, he\'ll [level up|Concept.Level]. Experience for killing an entity is shared among all the characters who did damage to that entity, proportional to the damage dealt by each character. Some experience is also awarded to all members of a faction who did damage to an entity, proportional to the damage dealt by that faction. \n\nBeyond the " + ::Const.XP.MaxLevelWithPerkpoints + "th  character level, characters are veterans and gain perk points only every 4th level. They can still improve their attributes but the attribute gain per level is small.")
```
Darxo: That new section is two mouthful to parse.
Also it should have the global constant about the exact percentage of the shares built-in.

I'm not happy with the following variant but maybe it can get us closer to an acceptable end-result.

The amount of experience that is gained from deaths is proportional to the amount of damage your company dealt to it. `::MSU.Text.colorizePercentage(::Const.XP.XPForKillerPct * 100.0)` of that experience is shared proportionally to the contributor. The other `::MSU.Text.colorizePercentage(100 - ::Const.XP.XPForKillerPct * 100.0)` are shared equally among all characters of that faction.

mod_reforged/hooks/ui/screens/tooltip/tooltip_events.nut
```diff
@@ -13,7 +13,7 @@
...
+						text = ::Reforged.Mod.Tooltips.parseString("Characters gain experience as they or their allies slay enemies in battles. If a character has accumulated sufficient experience, he\'ll [level up|Concept.Level].\n\nExperience gained from dying enemies is proportional to the fraction of the total damage dealt to that enemy by your company. Of this experience, " + ::Const.XP.XPForKillerPct + "% is shared proportionally between the brothers who did damage. The other " + (100 - ::Const.XP.XPForKillerPct) + "% is shared equally among all members of the company.\n\nBeyond the " + ::Const.XP.MaxLevelWithPerkpoints + "th  character level, characters are veterans and gain perk points only every 4th level. They can still improve their attributes but the attribute gain per level is small.")
```
Darxo: I suggest we add a global config variable for Veteran Perks and use that in this tooltip aswell as in the `rf_veteran_levels.nut`. Currently that variables value would be `4`
That doesn't to be part of this PR.

Darxo (approved):
My remaining comment is optional.

PR #196, opened 2023-10-23
feat: implement new stun threshold system

Darxo:
# Current Summary

### Terms
- **StunPower** : New stat on weapons and some active skills that determines how strong stunning attacks with them are
- **ComposureMax**: New character stat on entities that determines how well they can resist stunning attacks
- **Composure**: New stat on entities that is lost upon receiving stunning attacks and reset when it hits 0 or at the start of an entities turn

## Overview
- All Maces and all two-handed flails have a fixed StunPower. 
  - 2H Flails: **40**
  - 1H Human Maces: **80**
  - 1H Orc Maces: **120**
  - 2H Polemace/Goedendag; **120** 
  - remaining 2H Maces: **160**
- Most Races have a set ComposureMax (everyone else is treated as Human). E.g. Goblins have 60, Humans have 100 and Orcs have 150
- All stunning attacks now attack the Composure of the target
- Reducing the Composure below 0 with a stunning attack will stun them
- Reducing the enemy Composure to a value equal or below their negative max Composure will apply a double stun
- Headshots grant 50% increased StunPower
- Enemies that are immune to stuns (eg BaseProperties) are also immune to stuns from Composure
- Enemies that gain their stun immunity from effects (like Battering Ram on Orcs) are now susceptible to stuns from composure loss

### Perk Adjustments
- **Resilient** grants 50% increased Composure 
- **Battering Ram** grants 50% increased Composure 
- **Steelbrow** negates the Headshot bonus on StunPower
- **Mace Mastery** grants 50% increased StunPower while a Mace is equipped
- **Vigilant** grants +10 ComposureMax for every current Action Point

### Skill Adjustments
- All skills that had a Stun Chance (!= 0) before now use the Composure system
- **Shieldwall** grants 25% increased Composure (and no longer ignores the first stun)
- **Pound** grants 100% increased StunPower
- **Charge** inflicts 100 StunPower and Dazes the target if no stun was applied
- **Unstoppable Charge** inflicts 150 StunPower (if stun was rolled as opposed to knock back) and staggers the target if no stun was applied

## Todo
- Icon for new Composure effect
- Icon for Composure stat lines
- Ingame explanations about this system
- Nested Tooltips for StunPower and Composure atleast
- Smarter enemy AI that can predict stuns
- Some casual playtesting
- Review

### Maybe Todo
- Balance tweaks of certain weapon/skill values related to stun threshold

### DIFF
--- a/mod_reforged/hooks/config/character.nut
+++ b/mod_reforged/hooks/config/character.nut
@@ -25,6 +25,21 @@ local getClone = ::Const.CharacterProperties.getClone;
 ::Const.CharacterProperties.OffensiveReachIgnore <- 0;
 ::Const.CharacterProperties.BonusPerReachAdvantage <- 0;
 
+::Const.CharacterProperties.IsImmuneToStunFromPoise <- false;
+::Const.CharacterProperties.PoiseMax <- 100;
+::Const.CharacterProperties.PoiseMult <- 1.0;
+::Const.CharacterProperties.getPoiseMax <- function()
+{
+	return ::Math.floor(this.PoiseMax * this.PoiseMult);
+}
+::Const.CharacterProperties.PoiseDamage <- 0.0;		// Only weapons are meant to provide and set this value
+::Const.CharacterProperties.PoiseDamageMult <- 1.0;
+::Const.CharacterProperties.getPoiseDamage <- function()
+{
+	return ::Math.floor(this.PoiseDamage * this.PoiseDamageMult);
+}
+
+
 ::Const.ProjectileType.FlamingArrow <- ::Const.ProjectileType.COUNT;
 ::Const.ProjectileType.COUNT += 1;
 ::Const.ProjectileDecals.push(clone ::Const.ProjectileDecals[::Const.ProjectileType.Arrow]);
--- a/mod_reforged/hooks/config/character.nut
+++ b/mod_reforged/hooks/config/character.nut
@@ -26,7 +26,7 @@ local getClone = ::Const.CharacterProperties.getClone;
 ::Const.CharacterProperties.BonusPerReachAdvantage <- 0;
 
 ::Const.CharacterProperties.IsImmuneToStunFromPoise <- false;
-::Const.CharacterProperties.PoiseMax <- 100;
+::Const.CharacterProperties.PoiseMax <- ::Reforged.Poise.Default.Default;
 ::Const.CharacterProperties.PoiseMult <- 1.0;
 ::Const.CharacterProperties.getPoiseMax <- function()
 {
--- a/mod_reforged/hooks/skills/skill.nut
+++ b/mod_reforged/hooks/skills/skill.nut
@@ -116,7 +116,7 @@
 
 			if (this.m.IsStunningFromPoise)
 			{
-				if ("StunChance" in this.m)
+				if (::MSU.isIn("StunChance", this.m))
 				{
 					this.m.StunChance = 0;	// Skills that now use the Composure system should no longer use the StunChance
 				}
@@ -162,13 +162,15 @@
 
 	q.onUse = @(__original) function( _user, _targetTile )
 	{
-		local actor = this.getContainer().getActor();
 		if (this.m.IsStunningFromPoise)
 		{
+			local actor = this.getContainer().getActor();
+
 			local oldMaceSpec = actor.getCurrentProperties().IsSpecializedInMaces;
 			actor.getCurrentProperties().IsSpecializedInMaces = false;	// This will disable all vanilla StunChance based stuns
 			local ret = __original(_user, _targetTile);
 			actor.getCurrentProperties().IsSpecializedInMaces = oldMaceSpec;
+
 			return ret;
 		}
 		else
--- a/mod_reforged/hooks/skills/skill.nut
+++ b/mod_reforged/hooks/skills/skill.nut
@@ -85,8 +85,8 @@
 			if (!targetEntity.getCurrentProperties().IsImmuneToStun && this.m.IsStunningFromPoise)
 			{
 				local targetThresholdSkill = targetEntity.getSkills().getSkillByID("effects.poise");
-				local turnsStunnedBody = targetThresholdSkill.wouldStun(this.getContainer().getActor(), this, false);
-				local turnsStunnedHead = targetThresholdSkill.wouldStun(this.getContainer().getActor(), this, true);
+				local turnsStunnedBody = targetThresholdSkill.wouldStun(this.getContainer().getActor(), this, ::Const.BodyPart.Body);
+				local turnsStunnedHead = targetThresholdSkill.wouldStun(this.getContainer().getActor(), this, ::Const.BodyPart.Head);
 
 				ret.push({
 					icon = "ui/tooltips/positive.png",
--- a/mod_reforged/hooks/skills/skill.nut
+++ b/mod_reforged/hooks/skills/skill.nut
@@ -66,15 +66,31 @@
 		}
 
 		// New Entries
-		if (!this.m.IsShieldRelevant && _targetTile.IsOccupiedByActor && _targetTile.getEntity().isArmedWithShield())
+		if (_targetTile.IsOccupiedByActor)
 		{
-			local shield = _targetTile.getEntity().getItems().getItemAtSlot(::Const.ItemSlot.Offhand);
-			local bonus = (this.isRanged()) ? shield.getRangedDefenseBonus() : shield.getMeleeDefenseBonus();
-			if (bonus > 0)
+			local targetEntity = _targetTile.getEntity();
+			if (!this.m.IsShieldRelevant && targetEntity.isArmedWithShield())
 			{
+				local shield = targetEntity.getItems().getItemAtSlot(::Const.ItemSlot.Offhand);
+				local bonus = (this.isRanged()) ? shield.getRangedDefenseBonus() : shield.getMeleeDefenseBonus();
+				if (bonus > 0)
+				{
+					ret.push({
+						icon = "ui/tooltips/positive.png",
+						text = ::MSU.Text.colorGreen(bonus + "%") + " Ignores Shield",
+					});
+				}
+			}
+
+			if (!targetEntity.getCurrentProperties().IsImmuneToStun && this.m.IsStunningFromPoise)
+			{
+				local targetThresholdSkill = targetEntity.getSkills().getSkillByID("effects.poise");
+				local turnsStunnedBody = targetThresholdSkill.wouldStun(this.getContainer().getActor(), this, false);
+				local turnsStunnedHead = targetThresholdSkill.wouldStun(this.getContainer().getActor(), this, true);
+
 				ret.push({
 					icon = "ui/tooltips/positive.png",
-					text = ::MSU.Text.colorGreen(bonus + "%") + " Ignores Shield"
+					text = "Will stun for " + ::MSU.Text.colorGreen(turnsStunnedBody) + "/" + ::MSU.Text.colorGreen(turnsStunnedHead) + " turns",
 				});
 			}
 		}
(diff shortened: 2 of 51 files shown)


mod_reforged/hooks/config/character.nut
```diff
@@ -25,6 +25,21 @@ local getClone = ::Const.CharacterProperties.getClone;
 ::Const.CharacterProperties.OffensiveReachIgnore <- 0;
 ::Const.CharacterProperties.BonusPerReachAdvantage <- 0;
 
+::Const.CharacterProperties.IsImmuneToStunFromPoise <- false;
```
LordMidas, line 28: Probably don't need this boolean. Entities which are `IsImmuneToStun` should not get stunned by Poise. For entities which are by default immune to stun but who we want to be stunnable by Poise in our design, we can manually set their `IsImmuneToStun` to false.

mod_reforged/hooks/skills/skill.nut
```diff
@@ -63,15 +66,31 @@
...
+			if (!targetEntity.getCurrentProperties().IsImmuneToStun && this.m.IsStunningFromPoise)
 			{
+				local targetThresholdSkill = targetEntity.getSkills().getSkillByID("effects.poise");
+				local turnsStunnedBody = targetThresholdSkill.wouldStun(this.getContainer().getActor(), this, false);
+				local turnsStunnedHead = targetThresholdSkill.wouldStun(this.getContainer().getActor(), this, true);
+
 				ret.push({
 					icon = "ui/tooltips/positive.png",
-					text = ::MSU.Text.colorGreen(bonus + "%") + " Ignores Shield"
+					text = "Will stun for " + ::MSU.Text.colorGreen(turnsStunnedBody) + "/" + ::MSU.Text.colorGreen(turnsStunnedHead) + " turns",
```
LordMidas, line 93: This tooltip should instead by added by the `rf_poise_effect` via `onQueryTooltip`.

mod_reforged/hooks/skills/skill.nut
```diff
@@ -94,6 +113,68 @@
 			{
 				this.m.IconDisabled = ::String.replace(this.m.Icon, ".png", "_sw.png");
 			}
+
+			if (this.m.IsStunningFromPoise)
+			{
+				if ("StunChance" in this.m)
```
LordMidas, line 119: Should use `::MSU.isIn` with `true` for going up the chain of inheritance.

mod_reforged/hooks/skills/skill.nut
```diff
@@ -1,4 +1,7 @@
 ::Reforged.HooksMod.hook("scripts/skills/skill", function(q) {
+	q.m.IsDamagingPoise <- false;
+	q.m.IsStunningFromPoise <- false;
```
LordMidas, line 3: `IsStunningFromPoise` should not be a property of a skill, but rather what happens when a character's Poise reaches 0 should be a property of the Poise attribute itself. Just as an example this could, in theory, then be modified by perks/traits/effects etc. which would say "when this character's poise reaches 0, he gets Dazed instead of Stunned".

scripts/skills/special/rf_poise_effect.nut
```diff
@@ -0,0 +1,142 @@
...
+Our intended behavior is that a change in PoiseMax also changes the Poise of that character by the same amount:
+
+The PoiseMax is purely handled in the Properties while the current Poise is handled in this skill.
+It is much cleaner when a change to PoiseMax can solely focus on changing variables in Properties.
+*/
+
+this.rf_poise_effect <- ::inherit("scripts/skills/skill", {
+	m = {
+		// Config
+		HeadshotMult = 1.5,	// multiplier for incoming poise damage on a hit to the head (unless steelbrow was learned)
```
LordMidas, line 12: Perhaps it would be better if this is a character property similar to `BodyPartDamageMult` as an array so that the body/headshot mult can be defined on a per-character basis in their properties and by default the body can be 1.0 and head can be 1.5.

scripts/skills/special/rf_poise_effect.nut
```diff
@@ -0,0 +1,142 @@
...
+
+		// Private
+		InversePoise = 0,	// This is the inverse of Poise and counts upwards instead of downwards as Poise damaged is received
```
LordMidas, line 15: I think having Poise as a character property would be better. Then this skill can use the incoming damage to reduce the Poise in the CurrentProperties of this character. This handles Poise similar on the backend to other attributes e.g. Resolve and that keeps things cleaner and keeps the character properties in one place.

scripts/skills/special/rf_poise_effect.nut
```diff
@@ -0,0 +1,142 @@
...
+		this.m.IsHidden = false;
+		this.m.IsSerialized = false;
+		this.m.IsRemovedAfterBattle = false;
+	}
+
+	function getTooltip()
+	{
+		local tooltip = this.skill.getTooltip();
+		return tooltip;
+	}
```
LordMidas, line 36: Seems redundant.

scripts/skills/special/rf_poise_effect.nut
```diff
@@ -0,0 +1,142 @@
...
+		if (poiseDamage == 0) return;
+
+		if (_isHeadshot && !this.getContainer().hasSkill("perk.steel_brow"))
+		{
+			poiseDamage *= this.m.HeadshotMult;
+		}
+		poiseDamage = ::Math.floor(poiseDamage);
+
+		return poiseDamage;
+	}
```
LordMidas, line 131: Imo this should be directly from the attacker's perspective in this effect's` onAnySkillUsed` e.g. similar to how we do in `rf_reach_effect`. Then apply the damage in `onTargetHit`.

scripts/skills/special/rf_poise_effect.nut
```diff
@@ -0,0 +1,142 @@
...
+
+	// Returns, for how long we are being stunned right now (1 parameter) or would be, if given a simulated attack power
+	function calculateTurnsStunned( _PoiseMax, _simulatedAttackPower = 0 )	// _simulatedAttackPower is only used here when simulating whether something would stun
+	{
+		if (this.m.ReversePoise + _simulatedAttackPower < _PoiseMax) return 0;
+		if (this.m.ReversePoise + _simulatedAttackPower < (_PoiseMax * 2)) return 1;
+		return 2;
+	}
+
+	function calculatePoiseDamage( _attacker, _skill, _isHeadshot)
```
LordMidas, line 117: Instead of `_isHeadShot` just pass the body part.

scripts/skills/special/rf_poise_effect.nut
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	function onBeforeDamageReceived( _attacker, _skill, _hitInfo, _properties )
+	{
+		if (_skill == null || !_skill.m.IsDamagingPoise) return;
+
+		local attackPower = this.calculatePoiseDamage(_attacker, _skill, _hitInfo.BodyPart == ::Const.BodyPart.Head);
+		if (attackPower == 0) return;
+
+		this.onPoiseDamageReceived(_attacker, _skill, attackPower, true, _properties);
```
LordMidas, line 60: If we refactor it to apply poise damage from the attacker's perspective then this logic needs to be rewritten and moved to `onTargetHit`.

scripts/skills/special/rf_poise_effect.nut
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+// New Functions
+	function getPoise( _properties = null )
+	{
+		if (_properties == null) _properties = this.getContainer().getActor().getCurrentProperties();
+
+		return _properties.getPoiseMax() - this.m.InversePoise;
```
LordMidas, line 68: If we make Poise into a character property then `getPoise` should become a function inside character properties and we should fetch it from there.

LordMidas (changes requested):
Overall nice work.
- There are several tech-side improvements that can be made to the structure of the system. I'd be happy to make some changes directly to the branch to bring this into light.
- I would like to discuss the idea of making PoiseDamage of a weapon be automatically calculated from its StaminaModifier. This already has two immediate applications:
  - Our Reverse Grip stance in swordmaster perks that allows you to use a sword as a mace. Unfortunately swords do not have a default PoiseDamage defined, and therefore, that perk will need adjustment to work with this.
  - The point above is a proof of concept that this is going to be an issue with compatibility with mods that add weapons and do not consider the Poise system. I am working on a prototype setup which scales Poise based on StaminaModifier of the weapon so that it works out of the box with everything - and while the numbers it produces may not be 100% ideal for all weapons/cases, I'd argue that it is "mostly good" and should result in a fine baseline design for most cases.

We should discuss on VC in detail to go over my thoughts on the system and its implementation.

mod_reforged/hooks/config/character.nut
Darxo: But then they get stunned by non-poise stuns like Horrified Scream, Throwing Mastery bolas, Headshots from maces with the mace perk etc. Random mods might add more skills that guarantee a stun on someone.

scripts/skills/special/rf_poise_effect.nut
Darxo: I'm unsure about putting the built-up poise in the character properties when similar stats like Hitpoints, Fatigue and Actionpoints are located in the actor.nut
Is that really the right place for a value that does not reset to a base value during any update call?

scripts/skills/special/rf_poise_effect.nut
Darxo: I'm unsure about putting the built-up poise in the character properties when similar stats like Hitpoints, Fatigue and Actionpoints are located in the actor.nut
Is that really the right place for a value that does not reset to a base value during any update call?

mod_reforged/hooks/config/character.nut
LordMidas: I see. Yes that's a fair point. This creates a potential compatibility issue with other mods which add entities (e.g. Davkul Rising mod) which may have stun immune entities but which should be stunnable by Poise in our design. I don't think there's a way to create a compatibility layer unless we somehow utilize vanilla flags only or we create some sort of translation between them.

There are two possible compatibility layers we can make, each with its own pros and cons:
**Option 1:**
- Use the vanilla `IsImmuneToStun` flag only.
- Set that flag to `false` always.
- Entities which had that flag as true in their base properties, we give them a large PoiseMax e.g. 300 or 500 making them effectively "immune" to stun from poise.
- Advantage: Entities added by other mods are automatically assigned an appropriate Poise value and are subject to interaction with our system.
- Disadvantage: These entities are now suddenly stunnable by things which ignore poise e.g. Horrific Scream and Concussive Strikes.

**Option 2:**
- We add our own `IsImmuneToStunFromPoise` flag.
- This flag is always false. So, by default we assume all entities are stunnable from Poise.
- Entities which had `IsImmuneToStun` as true in their base properties, we give them a large PoiseMax e.g. 300 or 500 making them effectively "immune" to stun from poise.
- Advantage: Keeps the vanilla `IsImmuneToStun` flag so abilities which don't interact with the Poise system e.g. Horrific Scream, Concussive Strikes work as intended and don't need to check for a different flag.
- Disadvantage: It may cause some entities added by other mods to become "stunnable" by Poise but have a very high Poise to compensate for that.

I prefer Option 2. It is closer in its approach to what you have already implemented and I think the advantage to disadvantage weight is much better than Option 1.

scripts/skills/special/rf_poise_effect.nut
LordMidas: I see. In that case it makes more sense to have it in `actor` instead of in `properties`, similar to Fatigue and Hitpoints.

wangminqi:
Would you consider biasing Composure by resolve/hitpoint/fatigue etc?

Uberbagel:
> Would you consider biasing Composure by resolve/hitpoint/fatigue etc?

We have. I believe we decided something like that adds too much variability to the system making it too difficult for players to properly plan engagements.

Darxo:
This is an old PR. By now this version is put on ice but available as a submod (see the Submods channel in Discord)
I will close this

PR #197, opened 2023-10-23
refactor: replace ranged skill debuff with hitchance per distance debuff

Darxo:


mod_reforged/hooks/skills/special/night_effect.nut
```diff
@@ -0,0 +1,43 @@
...
+	q.m.HitChancePerTile <- -10;
+
+	q.getTooltip = @(__original) function()
+	{
+		local ret = __original();
+		foreach (entry in ret)
+		{
+			if (entry.id == 12 && entry.icon == "ui/icons/ranged_skill.png")
+			{
+				entry.text = "Has " + ::MSU.Text.colorizeValue(this.m.HitChancePerTile, {AddPercent = true}) + " chance to hit per tile of distance"
```
LordMidas, line 11: use `::MSU.Text.colorizePercentage`

mod_reforged/hooks/skills/special/night_effect.nut
```diff
@@ -0,0 +1,43 @@
...
+			}
+		}
+		return ret;
+	}
+
+	// This should eventually be removed once we add a general line for all kinds of 'HitChancePerTile' for a ranged attack
```
LordMidas, line 17: I'm not sure if I understand this comment. If it means that eventually we want to add a single combined line in hit factors for all HitChancePerTile for ranged attack, I disagree with that. Each effect should show its own effect in the hit factors. Therefore, if I have understood it correctly, then this comment should be removed.

LordMidas (changes requested):

mod_reforged/hooks/skills/special/night_effect.nut
Darxo: This reworked NIght Effect changes the shared vanilla `HitChanceAdditionalWithEachTile` variable in the character properties.
That variable is also touched by other effects like many shooting skills and the perk ballistics.

Because it is a shared variable it should be displayed __only once__ in the combat tooltip. 
The Vanilla representation of this variable is just the following term:
image

Once we adjust that term to display the accurate penalty from distance there is no more need to display the penalty from the night debuff

LordMidas: That's like saying anything that change `properties.MeleeSkill` e.g. Thrust, Unstoppable, Reach Advantage, and a ton of other effects should all be combined into a single `+x% hitchance` entry in hit factors. I completely disagree with this approach.

Anything that modifies `HitChanceAdditionalWithEachTile` ends up modifying the hitchance. Therefore, each such effect should show its own separate entry in the Hit Factors tooltip and show the exact amount that it is changing the hitchance by.

Hence, this comment is suggesting something that I believe we shouldn't do.

Darxo: I will remove the comment.

mod_reforged/hooks/skills/special/night_effect.nut
```diff
@@ -0,0 +1,42 @@
...
+	q.onGetHitFactors = @(__original) function( _skill, _targetTile, _tooltip )
+	{
+		__original(_skill, _targetTile, _tooltip);
+
+		local malus = this.m.HitChancePerTile * (this.getContainer().getActor().getTile().getDistanceTo(_targetTile) - _skill.m.MinRange);
+		if (malus != 0)
+		{
+			_tooltip.push({
+				icon = "ui/tooltips/negative.png",
+				text = ::Reforged.Mod.Tooltips.parseString(::MSU.Text.colorizePercentage(malus) + " [Nighttime|Skill+night_effect]")
```
LordMidas, line 26: Please add `{AddSign = false}` as the second argument to `colorizePercentage` because in hit factors we don't want to add sign (the sign is covered by the `icon`)

LordMidas (approved):

PR #201, opened 2023-10-25
Refactor daily cost rewrite

Darxo:
This rewrite achieves two things:
- Overall Performance improvement by no longer doing a costly daily cost calculation for every brother during every updateProperties call
- Introduction of two new global config variables to control how the DailyCost of brothers increases during regular and paragon levelups

In the second commit I took the liberty to move existing config variables into the new Player namespace.

### DIFF
--- a/mod_reforged/config/config.nut
+++ b/mod_reforged/config/config.nut
@@ -5,6 +5,11 @@
 	XPOverride = false, // Is set to true during actor.onActorKilled to prevent any xp gain via player.getXP function. Then set to false afterwards. This is to do our own override of XP gain system based on damage dealt ratios.
 	VeteranPerksLevelStep = 4, // During veteran character levels, a new perk point is gained every this many levels.
 
+	Player = {
+		DailyCostMultPerLevelRegular = 1.1,		// Exponential increase of daily cost for each level before max level
+		DailyCostMultPerLevelParagon = 1.03		// Exponential increase of daily cost for each level after max level
+	}
+
 	UI = {
 		WorldBannerYOffset = 50
 	}
--- a/mod_reforged/config/config.nut
+++ b/mod_reforged/config/config.nut
@@ -1,11 +1,11 @@
 ::Reforged.Config <- {
 	IsLegendaryDifficulty = true,
-	HiringCostVariance = 0.2,	// Hiring cost for recruits after gear is randomized +- this value. 0.0 = no variance. 0.2 = it varies between 80% and 120% of its original cost
-	HiringCostLuck = 200,	// Luck for rerolling hiring cost. Every 100 luck = 1 reroll. With multiple rerolls present only the closest to the original hiring cost is taken. 0 luck = to linear distribution between min and max,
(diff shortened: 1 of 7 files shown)


mod_reforged/hooks/skills/backgrounds/character_background.nut
```diff
@@ -14,14 +22,25 @@
 		}];
 	}
 
+	// Overwrite to increase overall performance. The Wage for this background is no longer calculated newly during every update loop
+	q.onUpdate = @() function( _properties )
+	{
+		_properties.DailyWage = this.calculateDailyCost();
```
LordMidas, line 28: You are still calculating the daily cost here, the same as vanilla, on every update loop. The only performance upgrade I see is that you no longer check for `trait.player` skill, which isn't enough of a performance upgrade to justify this overwrite of the entire function imo.

mod_reforged/hooks/skills/backgrounds/character_background.nut
```diff
@@ -1,9 +1,17 @@
 ::Reforged.HooksMod.hook("scripts/skills/backgrounds/character_background", function(q) {
+	q.m.CalculatedDailyCost <- 0;
```
LordMidas, line 2: I can't see where this `CalculatedDailyCost` variable actually used e.g. to set the daily wage.

mod_reforged/config/config.nut
```diff
@@ -1,9 +1,14 @@
...
-	VeteranPerksLevelStep = 4, // During veteran character levels, a new perk point is gained every this many levels.
+
+	Player = {
+		HiringCostVariance = 0.2,	// Hiring cost for recruits after gear is randomized +- this value. 0.0 = no variance. 0.2 = it varies between 80% and 120% of its original cost
+		HiringCostLuck = 200,	// Luck for rerolling hiring cost. Every 100 luck = 1 reroll. With multiple rerolls present only the closest to the original hiring cost is taken. 0 luck = to linear distribution between min and max,
+		XPOverride = false, // Is set to true during actor.onActorKilled to prevent any xp gain via player.getXP function. Then set to false afterwards. This is to do our own override of XP gain system based on damage dealt ratios.
+		VeteranPerksLevelStep = 4, // During veteran character levels, a new perk point is gained every this many levels.
+		DailyCostMultPerLevelRegular = 1.1,		// Exponential increase of daily cost for each level before max level
```
LordMidas, line 9: I'd rather just use the `onUpdate( _properties) ` function of `character_background` to use variables defined within the background itself to adjust the `DailyCostMult` or `_properties.DailyWage` directly. That seems more contained and clean.

LordMidas (changes requested):

mod_reforged/hooks/skills/backgrounds/character_background.nut
Darxo: My bad. The calculation was not supposed to happen here, I will fix that

mod_reforged/config/config.nut
Darxo: I don't follow where: Do you not want these two variables to be global? Do you want them to exist inside of each character background?

PR #202, opened 2023-10-25
refactor: change lucky trait into a 5% chance to avoid all damage

Darxo:
image

This is how it looks in the combat log.
I tested it ingame and the tooltip, damage reduction, log and animation seem to work correctly.

All on-hit effects that are not dependant on received damage are still applied (like Stun, Dazed or Dismantled Armor).
But that is completely fine by me.


LordMidas:
I'm not in favor of this tbh. Avoiding all damage is a bit too much. We can discuss separately what we want to do with Lucky. I'd close this PR until we actually have a design first that we want to implement.

Darxo:
The old trait was avoiding the complete attack, not just the damage part.

According to this post: https://www.reddit.com/r/BattleBrothers/comments/cdy4rd/quick_maths_how_good_is_the_lucky_trait/
and my personal eye-test this trait is very bad in vanilla. 
At best its equivalent to 2,5 Melee Defense and 2,5 Ranged Defense.

This rework is now a constant 5 Melee Defense and 5 Ranged Defense worth of avoiding damage but no longer avoids any effects that come with this hit.

Reducing the chance to 4% or 3% is always an option if it is too strong.

LordMidas:
I see. Yes that's a fair point. I still strongly prefer the design where Lucky avoids the attack instead of "receiving the attack but taking no damage". That sounds like what luck is. So I'd be open to a rework that makes the trait better but keeps the concept the same i.e. the attack misses completely.

PR #205, opened 2023-10-30
refactor: replace breastplate art with art from Houman

Uberbagel:


Darxo:
The commit message could be clearer. It sounds like the item in question got a completely new art.
But in reality the existing art was just refined/polished.

I don't know how to feel about crediting the artist in the commit message. This will get old really quickly (as pure artist generally don't commit their own stuff) and doesn't provide good information about what the commit actually did.
Their names should be attached to the art itself. And if that isn't enough then you can add their names in the description lines of the commit message, just not the first one.

LordMidas (approved):

PR #207, opened 2023-11-01
fix: use setWeaponType instead of addWeaponType

LordMidas:
This is a fix because of how the hook in MSU works that breaks Categories when using addWeaponType. Once the MSU behavior is fixed, we can revert to using addWeaponType.


Darxo (approved):
Looks good

PR #208, opened 2023-11-01
fix: add 100ms to onFollow of fling back and line breaker

Darxo:
This bandaid accounts for the mod Swifter that has faster combat speed

I playtested the Line Breaker delay alot and after that I didn't have any issues with my brothers or orcs not moving into the line-broken tile. Though I was also never playing on the highest speed. Only 3/3.5

Fling Back had the same issue. Though I haven't fought enough unholds to remember whether the issue was fixed

### DIFF
--- a/mod_reforged/hooks/skills/actives/fling_back_skill.nut
+++ b/mod_reforged/hooks/skills/actives/fling_back_skill.nut
@@ -1,0 +1,14 @@
+::Reforged.HooksMod.hook("scripts/skills/actives/fling_back_skill", function(q) {
+	// We delay the onFollow of this skill by another 100ms because otherwise the entity doesn't follow through with this function properly when at very high speeds using some faster combat speed mods
+	q.onFollow = @(__original) function( _tag )
+	{
+		if (::Time.getVirtualSpeed() > 2)
+		{
+			::Time.scheduleEvent(::TimeUnit.Virtual, 100, __original, _tag);
+		}
+		else
+		{
+			__original(_tag);
+		}
+	}
+});
--- a/mod_reforged/hooks/skills/actives/line_breaker.nut
+++ b/mod_reforged/hooks/skills/actives/line_breaker.nut
@@ -33,4 +33,17 @@
 
 		return null;
 	}
+
+	// We delay the onFollow of this skill by another 100ms because otherwise the entity doesn't follow through with this function properly when at very high speeds using some faster combat speed mods
+	q.onFollow = @(__original) function( _tag )
+	{
+		if (::Time.getVirtualSpeed() > 2)
+		{
+			::Time.scheduleEvent(::TimeUnit.Virtual, 100, __original, _tag);
+		}
+		else
+		{
+			__original(_tag);
+		}
+	}
 });


mod_reforged/hooks/skills/actives/line_breaker.nut
```diff
@@ -13,7 +13,7 @@
 				}
 			}
 
-			return null;
```
LordMidas, line 16: Please remove these whitespace changes from this PR.

mod_reforged/hooks/skills/actives/line_breaker.nut
```diff
@@ -28,9 +28,15 @@
 
 		altdir = dir + 1 <= 5 ? dir + 1 : 0;
 		knockToTile = getValidTileInDir(altdir);
-		
+
```
LordMidas, line 31: As above

mod_reforged/hooks/skills/actives/fling_back_skill.nut
```diff
@@ -0,0 +1,7 @@
+::Reforged.HooksMod.hook("scripts/skills/actives/fling_back_skill", function(q) {
+	// We delay the onFollow of this skill by another 100ms to account for faster combat speed mods
```
LordMidas, line 2: Please edit the comment to be more comprehensive for those who don't know the bug. Something like: `We delay the onFollow of this skill by another 100ms because otherwise the entity doesn't follow through with this function properly when at very high speeds using some faster combat speed mods`

mod_reforged/hooks/skills/actives/line_breaker.nut
```diff
@@ -28,9 +28,15 @@
...
 		if (knockToTile != null) return knockToTile;
 
 		return null;
 	}
+
+	// We delay the onFollow of this skill by another 100ms to account for faster combat speed mods
```
LordMidas, line 37: Please edit the comment here as above.

LordMidas (changes requested):

mod_reforged/hooks/skills/actives/fling_back_skill.nut
```diff
@@ -0,0 +1,7 @@
...
+	q.onFollow = @(__original) function( _tag )
+	{
+		::Time.scheduleEvent(::TimeUnit.Virtual, 100, __original, _tag);
```
LordMidas, line 5: Add an `if` before this saying `if (::Time.getVirtualSpeed() > 2)` so we only add this delay when the speed is 2x or greater, because we know that up to a speed of 2 it works fine.

mod_reforged/hooks/skills/actives/line_breaker.nut
```diff
@@ -28,9 +28,15 @@
...
+	q.onFollow = @(__original) function( _tag )
+	{
+		::Time.scheduleEvent(::TimeUnit.Virtual, 100, __original, _tag);
```
LordMidas, line 40: Same as above

LordMidas (approved):

LordMidas:
The commit message should be adjusted to something like `fix: add 100ms when at virtual speed greater than 2`. There is no need to mention the skill names in the commit message (that is apparent from the commit's diff).

PR #209, opened 2023-11-01
Fix accessory fatigue

Darxo:


mod_reforged/hooks/items/accessory/antidote_item.nut
```diff
@@ -0,0 +1,10 @@
+::Reforged.HooksMod.hook("scripts/items/accessory/antidote_item", function(q) {
+	// The Vanilla function calls the onEquip base function which only causes problems now that equipped accessory apply their fatigue penalty through the onUpdateProperties
+	// We replace this function to prevent that
+	q.onPutIntoBag = @() function()
```
LordMidas, line 4: Instead of this overwrite manually on accessory items, a more robust solution in my opinion is to change the hook on `onUpdateProperties` of `accessory.nut` to only apply the `StaminaModifier` if the accessory is not in bag. This way you let the vanilla `bag_fatigue` skill handle the `StaminaModifier` application of accessory items in bag.

```squirrel
q.onUpdateProperties = @() function( _properties )
{
	if (this.getCurrentSlotType() != ::Const.SlotType.Bag)
		_properties.Stamina += this.getStaminaModifier();
}
```
In my opinion this is a more robust solution as it covers also other bag-type accessories added by mods in the vanilla way. With this change, the hooks on both `antidote_item` and `bandage_item` can be removed.

LordMidas (changes requested):

LordMidas (approved):
Looks good

PR #214, opened 2023-11-04
feat: add colorizeMult function to MSU modifications

LordMidas:


mod_reforged/hooks/msu.nut
```diff
@@ -80,6 +80,16 @@
 	return this.colorizeValue(_value, _options);
 }
 
+::logInfo("Reforged::MSU -- adding ::MSU.Text.colorizeMult");
+::MSU.Text.colorizeMult <- function( _value, _options = null )
+{
+	if (_options == null) _options = {};
+	_options.AddPercent <- true;
+	_options.AddSign <- false;
```
Darxo, line 88: I think this should behave like a default value rather than overwriting whatever option was passed
So instead write `if (!("AddSign" in _options)) _options.AddSign<- false;
That way this function can be used to turn multipliers into signed percentages if the user chooses so.

LordMidas: Yeah I originally thought that it should never add sign which is why I wrote it like this. But I agree that adding the option to AddSign is probably a good idea. I will adjust it.

Darxo (approved):
Looks good

PR #215, opened 2023-11-04
refactor: reduce target value for melee attack if target has rebuke

LordMidas:


mod_reforged/hooks/ai/tactical/behavior.nut
```diff
@@ -3,6 +3,13 @@
     {
         // Currently Mortars are near invincible and not meant to be killed. But AI still targets them sometimes. This line should make it so they are basically never targeted.
         if (_target.getType() == ::Const.EntityType.Mortar) return 0.01;
-        return __original(_entity, _target, _skill);
+
+        local score = __original(_entity, _target, _skill);
+        if (_skill != null && !_skill.isRanged() && _target.getSkills().hasSkill("effects.rf_rebuke") && _target.getTile().getDistanceTo(_entity.getTile()) == 1)
```
Darxo, line 8: This does not check all conditions whether Rebuke would trigger.
According to the effect it doesn't trigger on stuns or while Riposte is active

Maybe it's time to refactor the rebuke effect to have some sort of `willProcc` function. This would run over all conditions and return true or false. Then it can be used within the rebuke effect aswell as from the outside from the AI code

LordMidas: Good point. I've added it.

mod_reforged/hooks/ai/tactical/behavior.nut
```diff
@@ -5,9 +5,11 @@
         if (_target.getType() == ::Const.EntityType.Mortar) return 0.01;
 
         local score = __original(_entity, _target, _skill);
-        if (_skill != null && !_skill.isRanged() && _target.getSkills().hasSkill("effects.rf_rebuke") && _target.getTile().getDistanceTo(_entity.getTile()) == 1)
+        if (_skill != null && !_skill.isRanged())
```
Darxo, line 8: I get, checking for isRanged here for performance reasons (sometimes saves us a getSkillByID search)
But it is redundant as it is also checked for in `willProc`

LordMidas: Yeah I only added it for performance reason. But I agree it is redundant and I also don't like repeated code - as it makes it more prone to mistakes later when we wanna change something in one place and forget to do it in another place. I think I will remove the `isRanged()` check here and eat the performance hit. That function already has a ton of `hasSkill` checks in the vanilla function so one more check isn't a huge deal.

Darxo (approved):
Looks good

PR #219, opened 2023-11-09
refactor: hide projected attributes when paragon levels are reached

Darxo:


LordMidas (approved):
I understand that `paragon levels` is a term you use to refer to max level with perk points, but this isn't convention. So I'd say please edit the commit message to e.g. `refactor: hide projected attributes after MaxLevelWithPerkpoints`.

Other than that, looks good.

Darxo:
Ill adjust the commit message but I will also leave this screenshot here.
The term "paragon" for this specific concept is not made up by me/diablo. It is used in the vanilla files

image

PR #222, opened 2023-11-09
refactor: require 'no current level-ups' for learning discovered talent

Darxo:
Simplify code of discovered talent skill

This is mostly a refactor to make the code cleaner.

But it can help prevent issues with alternative Gifted effects (like the in vanilla existing `spiritual_reward_item` that are pending on the player. Those are currently not accounted for and would be overwritten by this perk.

I tested this ingame and it works. The GUI even updates immediately once I spend the last level-up. No need to re-open the character screen as I feared at first

### DIFF
--- a/scripts/skills/perks/perk_rf_discovered_talent.nut
+++ b/scripts/skills/perks/perk_rf_discovered_talent.nut
@@ -34,41 +34,10 @@ this.perk_rf_discovered_talent <- ::inherit("scripts/skills/skill", {
 			}
 		}
 
-		local requiredLevelUpsSpent = this.getContainer().hasSkill("perk.gifted") ? 5 : 4;
-
-		if (actor.m.LevelUpsSpent < requiredLevelUpsSpent)
-		{
-			local startIndex = requiredLevelUpsSpent - actor.m.LevelUpsSpent;
-			local attributes = array(actor.m.Attributes.len());
-
-			foreach (i, attributeLevelUps in actor.m.Attributes)
-			{
-				attributes[i] = array(startIndex);
-				for (local j = 0; j < startIndex; j++)
-				{
-					attributes[i][j] = attributeLevelUps[j];
-				}
-			}
-
-			actor.m.Attributes.clear();
-			actor.fillAttributeLevelUpValues(::Const.XP.MaxLevelWithPerkpoints - actor.getLevel() + actor.m.LevelUps);
-
-			foreach (i, attributeLevelUps in attributes)
-			{
-				foreach (j, levelup in attributeLevelUps)
-				{
-					actor.m.Attributes[i][j] = levelup;
-				}
-			}
-		}
-		else
-		{
-			actor.m.Attributes.clear();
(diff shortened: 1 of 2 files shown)


Darxo:
It might be that you have to close and re-open the character menu after spending all your level-ups so that the perk counts as valid for picking now

scripts/skills/perks/perk_rf_discovered_talent.nut
```diff
@@ -34,41 +34,10 @@ this.perk_rf_discovered_talent <- ::inherit("scripts/skills/skill", {
...
-		}
-		else
-		{
-			actor.m.Attributes.clear();
-			actor.fillAttributeLevelUpValues(::Const.XP.MaxLevelWithPerkpoints - actor.getLevel() + actor.m.LevelUps);
-		}
-
+		// Because of its perk requirement we guaranteed that there are no pending past level-ups which would otherwise be overwritten
+		actor.m.Attributes.clear();
+		actor.fillAttributeLevelUpValues(::Const.XP.MaxLevelWithPerkpoints - actor.getLevel() + 1);
```
LordMidas, line 39: Doesn't this also change the behavior of the perk now? Because now you can't pick it before spending all current levelups. So if you gain multiple levels from a fight you might lose out on some benefit from this perk. Let's say you do a big fight and the bro is now level 8 after the fight. Now you pick Discovered Talent, with this new refactor, the perk only affects level-ups of levels 9-11 which isn't intended.

Darxo: That's correct.

LordMidas: We should try to avoid that. The original implementation is safe in the sense that it only re-rolls level-ups past the tier of Discovered Talent. The other attribute rolls are kept as they were before. So it doesn't matter when you pick the perk, it only affects levels 6-11 as intended.

The main problem with the original implementation is that the `requiredLevelUpsSpent` variable is calculated with a hard-coded check on `perk.gifted`. If we can somehow make that more scalable to auto-calculate required levelups that's be great. One way to achieve could be to compare the level-ups spent by a bro vs the level of the bro and the level-ups available to that bro. I have to leave right now but I'll try to think of a formula we could use for that.

Darxo: The perk is complicatedly implemented in the code and a little complicated for a player: 
> You can retroactively affect pending level-ups but you can't affect level-ups before level X

And on top of that I fix the issues that you can 
- overwrite gifted-like effects unless we add hard-coded exceptions for ever one of those in the code
- reroll an existing level-ups that the player already peeked at in the hopes to gain something better

I don't see the gain in taking the player by the hand so that they do not accidentally overlevel their brother and lose out on a few attributes. They will do that maybe a few times and then learn from that and become better player

Darxo:
discovered talent is getting completely reworked. this pr is no longer relevant

PR #223, opened 2023-11-10
refactor: customize beardchance for eunuch, wildman and barbarian

Darxo:
In vanilla these three backgrounds have the same chance of spawning with a beard than anyone else: 60%.
I believe it is more thematic if that chance is adjusted for these backgrounds

### DIFF
--- a/mod_reforged/hooks/skills/backgrounds/barbarian_background.nut
+++ b/mod_reforged/hooks/skills/backgrounds/barbarian_background.nut
@@ -2,6 +2,7 @@
 	q.create = @(__original) function()
 	{
 		__original();
+		this.m.BeardChance = 90;
 		this.m.PerkTreeMultipliers = {
 			"pg.rf_devious": 0,
 			"pg.rf_leadership": 0,
--- a/mod_reforged/hooks/skills/backgrounds/eunuch_background.nut
+++ b/mod_reforged/hooks/skills/backgrounds/eunuch_background.nut
@@ -2,6 +2,7 @@
 	q.create = @(__original) function()
 	{
 		__original();
+		this.m.BeardChance = 0;
 		this.m.PerkTreeMultipliers = {
 			"pg.rf_devious": 1.5,
 			"pg.rf_tactician": 4,


mod_reforged/hooks/skills/backgrounds/eunuch_background.nut
```diff
@@ -2,6 +2,7 @@
 	q.create = @(__original) function()
 	{
 		__original();
+		this.m.BeardChance = 5;
```
LordMidas, line 5: Eunuchs cannot have facial hair in real life. So I think this should be 0.

mod_reforged/hooks/skills/backgrounds/barbarian_background.nut
```diff
@@ -2,6 +2,7 @@
 	q.create = @(__original) function()
 	{
 		__original();
+		this.m.BeardChance = 95;
```
LordMidas, line 5: I think I'd be more comfortable with 85 for barbarian. 95 is a bit too high.

mod_reforged/hooks/skills/backgrounds/wildman_background.nut
```diff
@@ -2,6 +2,7 @@
 	q.create = @(__original) function()
 	{
 		__original();
+		this.m.BeardChance = 95;
```
LordMidas, line 5: I think this should be 100. But 95 is fine too.

LordMidas (changes requested):
These are nice thematic changes. For me, the eunuch and barbarian changes are important and the Wildman I leave to you to decide between 100 and 95.

Darxo:
I really like the 5% and 95% thresholds, alongside the default of 65%.
Those still leave rare/funny situations where you encounter some unusual person but mostly reflect the realism.

We shouldn't discuss for beardchance in between those values (like 85%), that's too much detail for such a minor detail and I don't see how we could ever come to result in a discussion around that except by bringing up real world statistics.

LordMidas:
> I really like the 5% and 95% extremes (in the style of rng in battle brothers), alongside the default of 65%. Those still leave rare/funny situations where you encounter some unusual person but mostly reflect the realism.
> 
> We shouldn't discuss for beardchance in between those values (like 85%), that's too much detail for such a minor detail and I don't see how we could ever come to result in a discussion around that except by bringing up real world statistics.

After your reply, I looked at vanilla numbers for reference. Wildman has 100 BeardChance in vanilla, and I prefer that over 95. Wildmen shouldn't be shaving their faces i.e. they are wild. So, please leave that at 100.

For Barbarian, in vanilla many barbarian enemies spawn without beards, so it makes little sense to me that the barbarian background has such a high chance. Shouldn't the chance match the chance of the barbarian enemies? And if we want to increase it considering that it is a "rare background" that the player can only sometimes hire, so to make it more thematic we increase the beard chance - I think 85 is a nice number for that.

So, I've updated my opinion on this: Wildman I feel strongly about now that it should be 100. Barbarian I'd prefer 85 but if you prefer that is also acceptable to me.

Darxo:
I didn't realize that vanilla already used custom chances here and there. I prefer staying in line with vanilla and using 90% for the barbarian.
Since the bellydancer already uses 0% beardchance I'm ok with doing the same for eunuch.
And wildman does not need any adjustments from its 100% in vanilla

LordMidas (approved):

PR #230, opened 2023-11-13
style: add comment delineating new reforged units from vanilla units

Uberbagel:


scripts/mods/mod_reforged/dynamic_spawns/units/noble_units.nut
```diff
@@ -43,7 +43,9 @@ local units = [
 		Troop = "CaravanDonkey",
 		Cost = 10,      // 0 in Vanilla
 		Figure = "cart_01"
-	}
+	},
+
+	//New in Reforged
```
Darxo, line 48: You should wait with this until there are actually new units for that faction

PR #235, opened 2023-11-15
refactor: add xp ratio scaling to dying and fleeing ally morale check

LordMidas:
Completely overwrites the vanilla actor functions of `onOtherActorDeath` and `onOtherActorFleeing`. The logic is the same as the vanilla functions except:
- The negative morale check from dying or fleeing allies is now scaled with the ratio of the ally's XP value to my XP value. This means that a weaker ally dying will trigger a weaker morale check for more powerful entities (e.g. Thug dying in front of a Bandit Leader. But a Leader dying or fleeing will trigger an even stronger than before morale check for Thug).
- The positive morale check from an enemy dying is only triggered if the killer is null or is allied to my faction.

EDIT: In my opinion this should be a `feat:` commit so I'll update the commit message to reflect that before merging.

### DIFF
--- a/mod_reforged/hooks/entity/tactical/actor.nut
+++ b/mod_reforged/hooks/entity/tactical/actor.nut
@@ -166,20 +166,24 @@
 	}
 
 	// Overwrite vanilla function. The logic is the same as vanilla except the following changes:
+	// Scale the strength of the morale checks with the ratio of the victim's XP value to my XP value.
 	// Dying enemies only trigger positive morale check if the killer is null or belongs to my allied factions.
 	q.onOtherActorDeath = @() function( _killer, _victim, _skill )
 	{
 		if (!this.m.IsAlive || this.m.IsDying || _victim.getXPValue() <= 1)
 			return;
 
+		// In Reforged we add this to scale the morale checks
+		local xpRatioMult = _victim.getXPValue() / ::Math.maxf(1, this.getXPValue());
+
 		if (_victim.getFaction() == this.getFaction() && _victim.getCurrentProperties().TargetAttractionMult >= 0.5 && this.getCurrentProperties().IsAffectedByDyingAllies)
 		{
-			local difficulty = ::Const.Morale.AllyKilledBaseDifficulty - _victim.getXPValue() * ::Const.Morale.AllyKilledXPMult + ::Math.pow(_victim.getTile().getDistanceTo(this.getTile()), ::Const.Morale.AllyKilledDistancePow);
+			local difficulty = ::Const.Morale.AllyKilledBaseDifficulty - _victim.getXPValue() * ::Const.Morale.AllyKilledXPMult * xpRatioMult + ::Math.pow(_victim.getTile().getDistanceTo(this.getTile()), ::Const.Morale.AllyKilledDistancePow);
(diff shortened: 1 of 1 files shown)


Darxo:
I suggest we rewrite this vanilla function entirely because there is bug with those morale checks that I wanted to fix for some time, which requires rewriting:

Currently in three-way fights, when you kill an enemy, then the enemies from the other faction become confident off of that kill even though your people did the last hit.
For example during a barbarian fight the unhold goes wild and changes faction, then it will gain morale and become confident for every kill happening around it.

We need an additional condition that makes sure that
`If the killer is not allied with your faction then your faction should not roll for a positive morale check from that kill.`

LordMidas:
@Uberbagel , @TaroEld , @Darxo Please review again as I have edited the PR with a complete overwrite of the vanilla functions. Now includes the suggestion from Darxo about positive morale check from dying enemies only if the killer is allied to me.

Uberbagel (approved):

mod_reforged/hooks/entity/tactical/actor.nut
```diff
@@ -154,6 +154,43 @@
...
 
+	// Overwrite vanilla function. The logic is the same as vanilla except the following changes:
+	// Scale the strength of the morale checks with the ratio of the victim's XP value to my XP value.
+	// Dying enemies only trigger positive morale check if the killer is null or of my faction.
+	q.onOtherActorDeath = @() function( _killer, _victim, _skill )
+	{
+		if (!this.m.IsAlive || this.m.IsDying || _victim.getXPValue() <= 1)
+			return;
+
+		if (_victim.getFaction() == this.getFaction() && _victim.getCurrentProperties().TargetAttractionMult >= 0.5 && this.getCurrentProperties().IsAffectedByDyingAllies)
```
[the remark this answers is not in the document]
LordMidas: This is a valid point but this seems like a problem that my modular vanilla mod should attempt to alleviate. I don't want to add this right now in Reforged unless it solving an immediate problem in some Reforged design.

Darxo (approved):

mod_reforged/hooks/config/tip_of_the_day.nut
```diff
@@ -19,6 +19,8 @@ for (local index = (::Const.TipOfTheDay.len() - 1); index >= 0; index--)
 }
 
 ::Const.TipOfTheDay.extend([
+	"Reforged: When a character is killed by someone in combat, positive morale checks are triggered only for allies of the killer, instead of for all enemies.",
+	"Reforged: Morale checks from dying/fleeing characters are stronger when the dying/fleeing character is more powerful - and vice versa.",
```
Darxo, line 23: You could write `that` instead of the second `dying/fleeing`. You decide

Darxo (approved):

mod_reforged/hooks/config/tip_of_the_day.nut
LordMidas: I like that change. However, we can go even further and do something like: `Morale checks due to relatively more powerful characters dying or fleeing are proportionately stronger, and vice versa.` I think this is concise and reads better.

Darxo (approved):

PR #236, opened 2023-11-15
fix: tooltips potentially overlapping the cursor by moving them to the left or right in that case

TaroEld:
I put the css classes to change in a dict so it's neater


Darxo (approved):
Looks good

LordMidas:
I don't understand js or css that much - so if Darxo thinks this is good then go ahead for merge.

LordMidas:
The commit message is too long though, needs to be shortened to fit within the character limit.

PR #237, opened 2023-11-17
refactor: combine bandit party ideal size code into separate function

Darxo:


scripts/mods/mod_reforged/dynamic_spawns/parties/bandit_parties.nut
```diff
@@ -15,25 +36,9 @@ local parties = [
 			{ BaseID = "UnitBlock.RF.BanditDog", RatioMin = 0.00, RatioMax = 0.20 }
 		]
 
-		function generateIdealSize( _isLocation = false )
+		generateIdealSize( _isLocation = false )
```
LordMidas, line 39: Missing `function` keyword

LordMidas:
The utility of this move is only so far as until we decide to have different formulas for each party (which I intend to discuss with the Design team) - so I'd hold off on this.

LordMidas (changes requested):

PR #239, opened 2023-11-17
refactor(balance): replace double grip on zombies with +5 melee skill

Darxo:
This will slightly buff all zombies with shield and those with two-handed weapons (even fallen heroes).
And it will slightly nerf all zombies that have a one-handed weapon and no shield.

But most importantly it will make breaking shields against zombies much more useful

### DIFF
--- a/mod_reforged/hooks/entity/tactical/enemies/zombie.nut
+++ b/mod_reforged/hooks/entity/tactical/enemies/zombie.nut
@@ -115,11 +115,12 @@
 		this.addDefaultStatusSprites();
 		this.getSprite("arms_icon").setBrightness(0.85);
 		this.getSprite("status_rooted").Scale = 0.55;
-		this.m.Skills.add(this.new("scripts/skills/special/double_grip"));
+		// this.m.Skills.add(this.new("scripts/skills/special/double_grip"));
 		this.m.Skills.add(this.new("scripts/skills/actives/zombie_bite"));
 		// this.m.Skills.add(this.new("scripts/skills/perks/perk_battle_forged"));
 
 		// Reforged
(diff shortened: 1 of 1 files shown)


LordMidas:
Which regular zombies spawn with shields? I know about Fallen Heroes. Are there others?

Uberbagel:
> Which regular zombies spawn with shields? I know about Fallen Heroes. Are there others?

Just guard zombies I believe

LordMidas:
I like the theme of removing Double Grip from zombies. However, very few of them ever spawn with shields and the +5 Melee Skill is universal so it ends up being an overall buff for the zombie faction. I'd suggest we hold off on this until the balance build so we can reassess this in detail.

Darxo:
This affects Fallen Heroes the most, but currently shields also appear on:
- zombie nomads
- any reanimated humanoid that had a shield

Darxo:
shields are now practically indestructable which removes the reason for this change

LordMidas:
Zombie and Ancient Dead shields have lower durability by design and are breakable. So I think this thing should still be on the docket for discussion. I think the point that zombies should lose double grip is thematic and makes sense. However, I don't know what to do to balance them after removing the double grip. I don't think +5 Melee Skill is an appropriate replacement.

Darxo:
Purely balance reason.
This change is supposed to encourage the usage of shield breaking against zombies because that will now always make them easier and never more dangerous.

Uberbagel:
Well, now that breaking shields will almost exclusively happen against enemies, we can discuss this problem again from a fresh state. I'd like to put on the table two solutions. 

1: Getting your shield broken staggers you until the start of your next turn. 
or
2: Delay the double grip bonus for one round after getting your shield broken.

PR #240, opened 2023-11-17
refactor: slightly adjust the combat difficulties

Darxo:
Add 10% damage nerf to all NPCs while on Easy difficulty 
Increase resource multiplier on Esay to 1.00 (up from 0.85) 
Increase resource multiplier on Hard difficulty to 1.20 (up from 1.15)

The goal is to make fights on easy more fun by having the player encounter more interesting and varied enemy types.
The slight increase for the Hard difficulty is meant to make the 3 difficulties slightly more distinct. I feel that 15% difference is not quite enough

### DIFF
--- a/mod_reforged/hooks/entity/tactical/actor.nut
+++ b/mod_reforged/hooks/entity/tactical/actor.nut
@@ -44,6 +44,15 @@
 		}
 	}
 
+	q.onAfterInit = @(__original) function()
+	{
+		__original();
+		if (this.getType() != ::Const.EntityType.Player)
+		{
+			this.m.BaseProperties.DamageTotalMult *= ::Const.Difficulty.NPCDamageMult[::World.Assets.getCombatDifficulty()];
+		}
+	}
+
 	q.onAttackOfOpportunity = @(__original) function( _entity, _isOnEnter )
 	{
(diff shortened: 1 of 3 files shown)


mod_reforged/hooks/entity/tactical/actor.nut
```diff
@@ -13,6 +13,10 @@
 	{
 		__original();
 		this.m.RF_DamageReceived = { Total = 0.0 };
+		if (::World.Assets.getCombatDifficulty() == ::Const.Difficulty.Easy && this.getType() != ::Const.EntityType.Player)
+		{
+			this.m.BaseProperties.DamageTotalMult = 0.9;
```
LordMidas, line 18: This has to happen in `onInit` because that's where the BaseProperties are initialized.

mod_reforged/hooks/entity/tactical/actor.nut
```diff
@@ -13,6 +13,10 @@
...
+		}
```
LordMidas, line 19: All the changes in this PR also need to be somehow reflected in the tooltips of the difficulty settings.

LordMidas (changes requested):
I'd get @TaroEld's opinion on this particular PR as well.

PR #242, opened 2023-11-19
fix: unkillable hexen when soul link target disappears/dies first

Darxo:


LordMidas (approved):
The commit message is too specific. Fixing the hexen case is a specific example of the bug - the fix itself is fixing a more general issue of the target being potentially null. I'd suggest changing it to:
```
fix: add handling for edge case where transfer target can be null

An example of the edge case is where a Hexe would charm a target making it a valid link. Then the Hexe takes damage, when the link is still valid. Then it dies causing the Charm to expire and the link to no longer be valid and the transfer target being null - causing the Hexe to bug out and become unkillable.
``

PR #243, opened 2023-11-21
refactor(balance): remove setting reachMult to 0 when stunned or fleeing

LordMidas:
This is a major complaint right now and I agree that the situation is bad. Getting stunned or fleeing becomes a death sentence. Stunned/Fleeing already give their own major disadvantages so we can remove the death sentence for now.


Darxo:
As a Band-Aid because of the current balance issues of Reach I'm completely ok with this.

In general I like that debuff (which you now removed) from a mechanical pov

scripts/skills/special/rf_reach.nut
```diff
@@ -61,7 +61,8 @@ this.rf_reach <- ::inherit("scripts/skills/skill", {
 
 	function onUpdate( _properties )
 	{
-		if (_properties.IsStunned || this.getContainer().getActor().getMoraleState() == ::Const.MoraleState.Fleeing || this.getContainer().getAttackOfOpportunity() == null)
+		local actor = this.getContainer().getActor();
+		if (actor.isArmedWithRangedWeapon() && !actor.getMainhandItem().isItemType(::Const.Items.ItemType.MeleeWeapon)) // Ensure that it's not a melee/ranged hybrid weapon
```
Darxo, line 65: The only exception you cover here are ranged weapons that are also melee weapons.
But an entity could have an attack of opportunity for other reasons, like an offhand weapon or just being able to use the offhand as a fist.
Have you considered instead checking for the existence of an attack of opportunity skill?
`if (!actor.getSkills().getAttackOfOpportunity())`

LordMidas: I think that's a good idea. Can check directly for the presence of an attack of opportunity.

LordMidas: There may still be edge cases e.g. where an entity has a ranged weapon equipped (with >0 Reach), and has a melee attack possible from that weapon but which isn't considered an Attack of Opportunity. We don't have to worry about that edge case for now - but just something to keep in mind for later.

Darxo: Those melee skills need to then turn on the flag `IsIgnoredAsAOO = true` that exists in vanilla. Then they are ignored as AOO

Darxo (approved):

Darxo:
You need to remove the last sentence in the Reach nested tooltip now with this change.

PR #246, opened 2023-11-23
Bagel

Uberbagel:


LordMidas (approved):
I'll add a comment in the code to explicitly state that the hook on the footman_armor description is to fix a typo.

PR #251, opened 2023-11-28
feat: add %factionname% in entity name to fetch the name of faction

LordMidas:
Useful for champions to be able to mention their faction in their name or title.


mod_reforged/hooks/config/world_entity_common.nut
```diff
@@ -0,0 +1,26 @@
+// Add functionality to allow using %factionname% in champion names which is to be replaced by that character's faction's name
+::Const.World.Common.RF_generateName <- function( _string, _faction )
+{
+	return ::String.replace(_string, "%factionname%", ::World.FactionManager.getFaction(_faction).getName());
+}
+
```
TaroEld, line 6: This is an awfully specific function with an awfully generic name. It either needs a more specific name, or it needs to be able to process more replaces. In the latter case, make it replace string in-place, and return the result, indicating that more replaces will be added in the future. Of course, the _faction parameter makes that tough. One option is to pass a table of `{"%placeholder", value}`, and iterate through it.

TaroEld (approved):

PR #259, opened 2024-01-01
feat: add new perk Calculated Strikes

Darxo:
### Todo:

image


mod_reforged/hooks/config/strings.nut
```diff
@@ -863,6 +864,15 @@ foreach (vanillaDesc in vanillaDescriptions)
 			]
 		}]
 	}),
+	RF_CalculatedStrikes = ::UPD.getDescription({
+		Effects = [{
+			Type = ::UPD.EffectType.Passive,
+			Description = [
+				"Damage is increased by " + ::MSU.Text.colorGreen("20%") + " against anyone who\'s turn has not started yet."
```
LordMidas, line 871: - `whose`
- Please add hyperlink for `[turn|Concept.Turn]`.
- `not started yet in the current [round|Concept.Round].` to make it clear that it is a per-round thing and not a once per combat thing.

scripts/skills/perks/perk_rf_calculated_strikes.nut
```diff
@@ -0,0 +1,51 @@
...
+	}
+
+	function onUpdate( _properties )
+	{
+		_properties.InitiativeMult *= this.m.InitiativeMult;
+	}
+
+	function onAnySkillUsed( _skill, _targetEntity, _properties )
+	{
+		if (_skill != this) return;
```
LordMidas, line 26: This is wrong. The `_skill` is the skill being used e.g. `Slash`. The `isEnabledFor` check you have is the only one you need.

scripts/skills/perks/perk_rf_calculated_strikes.nut
```diff
@@ -0,0 +1,51 @@
...
+		}
+	}
+
+	function onGetHitFactors( _skill, _targetTile, _tooltip )
+	{
+		if (_targetTile.getEntity() != null && this.isEnabledFor(_targetTile.getEntity()))
+		{
+			_tooltip.push({
+				icon = "ui/icons/damage_dealt.png",
+				text = ::MSU.Text.colorizeMult(this.m.DamageMult) + " increased damage (" + this.getName() + ")"
```
LordMidas, line 40: This is a slightly difficult one to decide about. I don't think "increased damage" should be added to hit factors. You can simply have the name of this perk. Also hit factors should only ever use the `-` and `+` icons. So `icon = ui/icons/positive.png` and then `text = this.getName()`.

scripts/skills/perks/perk_rf_calculated_strikes.nut
```diff
@@ -0,0 +1,51 @@
...
+
+		if (this.isEnabledFor(_targetEntity))
+		{
+			_properties.DamageTotalMult *= this.m.DamageMult;
+		}
+	}
+
+	function onGetHitFactors( _skill, _targetTile, _tooltip )
+	{
+		if (_targetTile.getEntity() != null && this.isEnabledFor(_targetTile.getEntity()))
```
LordMidas, line 36: I just realized that here also we should start being safe and use `_targetTile.IsOccupiedByActor` instead of checking for entity being null. Because what if a skill is targeting some non-actor entity e.g. tree or obstacle, then this might crash when it checks for `isTurnStarted()` etc. So:
```squirrel
if (_targetTile.IsOccupiedByActor && this.isEnabledFor(_targetTile.getEntity()))
```

LordMidas (changes requested):

mod_reforged/hooks/config/strings.nut
```diff
@@ -860,6 +861,15 @@ foreach (vanillaDesc in vanillaDescriptions)
 			]
 		}]
 	}),
+	RF_CalculatedStrikes = ::UPD.getDescription({
+		Effects = [{
+			Type = ::UPD.EffectType.Passive,
+			Description = [
+				"Deal " + ::MSU.Text.colorGreen("20%") + " more Damage against anyone who has not started their [turn|Concept.Turn] yet in the current [round|Concept.Round]."
```
LordMidas, line 868: `damage` instead of `Damage`

LordMidas, line 868: Missing comma at the end

mod_reforged/hooks/config/strings.nut
```diff
@@ -860,6 +861,15 @@ foreach (vanillaDesc in vanillaDescriptions)
...
+				::MSU.Text.colorRed("20%") + " less [Initiative|Concept.Initiative]."
```
LordMidas, line 869: I don't think this looks good in perk descriptions. We should instead say `Initiative is reduced by 20%`.  This `20% less initiative` style should be used in status effect tooltips instead.

scripts/skills/perks/perk_rf_calculated_strikes.nut
```diff
@@ -0,0 +1,49 @@
...
+				icon = "ui/tooltips/positive.png",
+				text = this.getName()
+			});
+		}
+	}
+
+// New Functions
+	function isEnabledFor( _targetEntity )
+	{
+		return (!_targetEntity.isTurnStarted() && !_targetEntity.isTurnDone());	// This is the same condition as vanilla overwhelm has
```
LordMidas, line 46: Please remove parantheses from return statement.

LordMidas (changes requested):

mod_reforged/hooks/config/strings.nut
Darxo: How about `You have 20% less Initiative`?

LordMidas: Not good. That sounds like a descriptive statement of state rather than description of what would happen as a result of this perk i.e. it sounds like I am telling someone what they already have rather than what would happen to them. I don't think we need to shoehorn the less/more wording into perk descriptions. The original way of saying it reads much better and is more in line with the game's style in my opinion.

mod_reforged/hooks/config/strings.nut
```diff
@@ -71,6 +71,7 @@
 	RF_Angler = "Angler",
 	RF_BattleFervor = "Battle Fervor",
 	RF_BackToBasics = "Back to Basics",
+	RF_CalculatedStrikes = "Calculated Strikes"
```
LordMidas, line 74: Missing comma

mod_reforged/config/perk_defs.nut
```diff
@@ -6,7 +6,15 @@
...
 		IconDisabled = "ui/perks/rf_heft_bw.png",
-	}
+	},
+	{
+		ID = "perk.rf_calculated_strikes",
+		Script = "scripts/skills/perks/perk_rf_calculated_strikes",
+		Name = ::Const.Strings.PerkName.RF_CalculatedStrikes,
+		Tooltip = ::Const.Strings.PerkDescription.RF_CalculatedStrikes,
+		Icon = "ui/perks/rf_calculated_strikes.png",
+		IconDisabled = "ui/perks/rf_calculated_strikes_bw.png",
```
LordMidas, line 16: should be `_sw`. Please change the name of the `.png` file accordingly.

LordMidas (approved):
Good to merge after that name change of the sw image.

PR #261, opened 2024-01-01
feat: add new perk Fluid Fighter

Darxo:
### Issues
- This perk might be a bit on the nose/auto-pick for crossbow/gun users because they use two different skills each turn by design.

### Todo:
- Rename into combo and apply our recent adjustments
- Testing
- Artwork for Perk
- Add it into the correct perk group


Darxo:
This perks effect has changed a bit after I implemented this version (my original pitch).

Darxo:
Midas did the implementation now so this PR is no longer required

PR #262, opened 2024-01-01
feat: add new perk Second Wind

Darxo:
### Screenshots

image

image

### Issues
- The last entity during each round is not able to use **Wait**. This is a Vanilla limitation but it will sometimes make this perk awkward to use


mod_reforged/hooks/config/strings.nut
```diff
@@ -1654,7 +1655,15 @@ foreach (vanillaDesc in vanillaDescriptions)
 				"The [Fatigue|Concept.Fatigue] Cost of weapon skills is reduced by " + ::MSU.Text.colorRed("25%") + "."
 			]
 		}]
- 	}),
+	}),
+	RF_SecondWind = ::UPD.getDescription({
+		Effects = [{
```
LordMidas, line 1660: We need Fluff for this perk. Ideas:
- `Just one more turn!` which is a reference to turn based games where you keep telling yourself that you will turn off the game after one more turn.
- `There\'s more where that came from."`
- `I\'m not done yet.`

mod_reforged/hooks/config/strings.nut
```diff
@@ -1654,7 +1655,15 @@ foreach (vanillaDesc in vanillaDescriptions)
...
+			Type = ::UPD.EffectType.Passive,
+			Description = [
+				"When you use \'Wait\', recover Action Points until you have 4 Action Points"
```
LordMidas, line 1663: Please add hyperlinks for the concepts.

mod_reforged/config/perk_defs.nut
```diff
@@ -127,6 +127,14 @@
 		Icon = "ui/perks/rf_small_target.png",
 		IconDisabled = "ui/perks/rf_small_target_bw.png",
 	},
+	{
+		ID = "perk.rf_second_wind",
+		Script = "scripts/skills/perks/perk_rf_second_wind",
+		Name = ::Const.Strings.PerkName.RF_SecondWind,
+		Tooltip = ::Const.Strings.PerkDescription.RF_SecondWind,
+		Icon = "ui/perks/rf_second_wind.png",
+		IconDisabled = "ui/perks/rf_second_wind_bw.png",
```
LordMidas, line 136: We decided to use `_sw` for all perk icons. The `_bw` is from older icons before we decided that and will be eventually changed to `_sw`.

scripts/mods/mod_reforged/perk_groups/pg_rf_large.nut
```diff
@@ -21,7 +21,7 @@ this.pg_rf_large <- ::inherit(::DynamicPerks.Class.PerkGroup, {
 				[],
 				[],
 				[],
-				[],
+				["perk.rf_second_wind"],
```
LordMidas, line 24: Don't add the perk to the perk group in the perk's implementation PR. We'll make a separate PR later where each perk group is fully updated with all the perks it needs.

scripts/skills/perks/perk_rf_second_wind.nut
```diff
@@ -0,0 +1,31 @@
+this.perk_rf_second_wind <- ::inherit("scripts/skills/skill", {
+	m = {
+		RecoverActionPointTarget = 4
```
LordMidas, line 3: `ActionPoints` is always plural so I'd rewrite this as `RecoverActionPointsTarget` and in fact you could make it even shorter and just write `ActionPointsTarget` and that's enough because we know from the perk itself that it's gonna recover it to that target.

scripts/skills/perks/perk_rf_second_wind.nut
```diff
@@ -0,0 +1,31 @@
...
+
+	function onWaitTurn()
+	{
+		local actor = this.getContainer().getActor();
+
+		local recoveredActionPoints = this.m.RecoverActionPointTarget - actor.getActionPoints();
+		if (recoveredActionPoints > 0)
+		{
+			actor.setActionPoints(actor.getActionPoints() + recoveredActionPoints);
+			::Tactical.EventLog.log(::Const.UI.getColorizedEntityName(actor) + " recovers " + ::MSU.Text.colorGreen(recoveredActionPoints) + " Action Point(s)");
```
LordMidas, line 26: This is minor but it would be nice to do `Action " + (recoveredActionPoints == 1 ? "Point" : "Points")` instead of `(s)`.

scripts/skills/perks/perk_rf_second_wind.nut
```diff
@@ -0,0 +1,31 @@
...
+	}
+
+	function onWaitTurn()
+	{
+		local actor = this.getContainer().getActor();
+
+		local recoveredActionPoints = this.m.RecoverActionPointTarget - actor.getActionPoints();
+		if (recoveredActionPoints > 0)
+		{
+			actor.setActionPoints(actor.getActionPoints() + recoveredActionPoints);
```
LordMidas, line 25: You need to make sure you don't exceed the maximum action points (in case for some weird reason this actor's max action points are fewer than 4).

LordMidas (changes requested):

scripts/skills/perks/perk_rf_second_wind.nut
Darxo: I prefer the readability of this code (40 less character in that line) over the ever so slight improvement of the resulting text, and I prefer if this decision remains coders preference.

scripts/skills/perks/perk_rf_second_wind.nut
```diff
@@ -0,0 +1,32 @@
...
+	m = {
+		ActionPointsTarget = 4	// set the current Action Points to this value if it's lower
+	},
+	function create()
+	{
+		this.m.ID = "perk.rf_second_wind";
+		this.m.Name = ::Const.Strings.PerkName.RF_SecondWind;
+		this.m.Description = ::Const.Strings.PerkDescription.RF_SecondWind;
+		this.m.Icon = "ui/perks/rf_second_wind.png";
+		this.m.Overlay = "rf_second_wind_effect";
```
LordMidas, line 11: 1. We use `_effect` for brush icons only for actual status effects. For perks we use the same name as the perk's icon with `perk_` prefixed to it. So the sprite's name in this case should be `perk_rf_second_wind`. As a guideline, keep it the same as the filename of the skill.
2. Do not add this as `this.m.Overlay` because `this.m.Overlay` is triggered when a skill is added to an entity, or when an active skill is used by an entity. We don't want the icon to automatically spawn if e.g. this perk is added to an entity during battle somehow.

scripts/skills/perks/perk_rf_second_wind.nut
```diff
@@ -0,0 +1,32 @@
...
+		this.m.Type = ::Const.SkillType.Perk;
+		this.m.Order = ::Const.SkillOrder.Perk - 5;
```
LordMidas, line 13: - Why `-5`?
- On that note, a small difference like this doesn't really achieve anything in a predictable way because whenever a skill is added to an actor's `skill_container` via the `container.add` function it increases the `Order` of that skill object by `+1`. So, I'd say that this order should be `::Const.SkillOrder.Perk` or if you want it to trigger before all other perks (why?) then set it to an earlier perk order type declared in the vanilla SkillOrder table.

LordMidas (changes requested):

scripts/skills/perks/perk_rf_second_wind.nut
LordMidas: There are ways to write it in a readable way in my opinion but honestly it's a minor thing. I see that vanilla also does this `(s)` thing in tooltips so I guess there is precedent for that already. So you can disregard my comment.

scripts/skills/perks/perk_rf_second_wind.nut
Darxo: I don't remember anymore. I will remove that modifier and we have to see if there was ever an issue. And next time I put a comment there explaining it.

LordMidas (approved):

PR #265, opened 2024-02-08
feat: increase town interaction range while on a caravan mission

Darxo:
This is done by replicating the important conditions for this kind of click. 

At the very end, we skip checking for whether player and target are on the same tile and also check for the increased range

### DIFF
--- a/mod_reforged/hooks/config/tip_of_the_day.nut
+++ b/mod_reforged/hooks/config/tip_of_the_day.nut
@@ -19,7 +19,7 @@ for (local index = (::Const.TipOfTheDay.len() - 1); index >= 0; index--)
 }
 
 ::Const.TipOfTheDay.extend([
-    "You can interact with towns while on a caravan mission.",
+    "You can enter towns while on a caravan mission; even if they are 2 tiles away.",
     "The Mod Options have a lot of customizations regarding quality of life.",
     "Morale Checks are harder for every adjacent enemy and easier for every adjacent ally.",
     "Click on an active contract to focus on its target on the world map, if it is known to you.",
--- a/mod_reforged/hooks/config/tip_of_the_day.nut
+++ b/mod_reforged/hooks/config/tip_of_the_day.nut
@@ -19,7 +19,7 @@ for (local index = (::Const.TipOfTheDay.len() - 1); index >= 0; index--)
 }
 
 ::Const.TipOfTheDay.extend([
-    "You can enter towns while on a caravan mission; even if they are 2 tiles away.",
+	"You can enter towns while on a caravan mission, even if they are 2 tiles away.",
     "The Mod Options have a lot of customizations regarding quality of life.",
     "Morale Checks are harder for every adjacent enemy and easier for every adjacent ally.",
     "Click on an active contract to focus on its target on the world map, if it is known to you.",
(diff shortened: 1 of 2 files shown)


mod_reforged/hooks/config/tip_of_the_day.nut
```diff
@@ -19,7 +19,7 @@ for (local index = (::Const.TipOfTheDay.len() - 1); index >= 0; index--)
 }
 
 ::Const.TipOfTheDay.extend([
-    "You can interact with towns while on a caravan mission.",
+    "You can enter towns while on a caravan mission; even if they are 2 tiles away.",
```
LordMidas, line 22: The semi-colon is wrong here. You can simply remove it and the sentence would be ok.

mod_reforged/hooks/states/world_state.nut
```diff
@@ -13,4 +13,30 @@
 
     	return __original(_properties, _isPlayerInitiated, _isCombatantsVisible, _allowFormationPicking);
     }
+
+	q.onMouseInput = @(__original) function( _mouse )
+	{
+		local ret = __original(_mouse);
+		if (ret == false)	// If the original function wasn't able to process the mouseclick we try to do that with increased interaction range
+		{
+			local isEscorting = this.m.EscortedEntity != null && !this.m.EscortedEntity.isNull();
```
LordMidas, line 22: Just use `!::MSU.isNull(this.m.EscortedEntity);`

mod_reforged/hooks/states/world_state.nut
```diff
@@ -13,4 +13,30 @@
...
+			if (isEscorting && _mouse.getState() == 1 && !this.isInCameraMovementMode() && !this.m.WasInCameraMovementMode)
```
LordMidas, line 23: The `isEscorting` local variable seems unnecessary. You can just put here `if (!::MSU.isNull(this.m.EscortedEntity) && `

mod_reforged/hooks/states/world_state.nut
```diff
@@ -13,4 +13,30 @@
...
+			{
+				foreach (entity in ::World.getAllEntitiesAndOneLocationAtPos(::World.getCamera().screenToWorld(_mouse.getX(), _mouse.getY()), 1.0))
+				{
+					if (entity.isParty()) continue;
```
LordMidas, line 27: I'd also limit the `entity` here to be type of `settlement.nut` so that only actual towns are enterable.

mod_reforged/hooks/states/world_state.nut
```diff
@@ -13,4 +13,30 @@
...
+					if (!entity.isEnterable()) continue;
+					if (!entity.isAlliedWithPlayer()) continue;
```
LordMidas, line 29: I really think the following looks and reads much better:
```squirrel
if (entity.isParty() || !entity.isEnterable() || !entity.isAlliedWithPlayer())
   continue;
```
than these 3 split very small checks. But it's up to you.

LordMidas (changes requested):
I am not an expert in this, but it looks ok to me. I'd suggest to also request review from @TaroEld.

TaroEld:
LGTM, should work.

LordMidas (approved):

PR #266, opened 2024-02-09
refactor: improve structure and moddability of encumbrance_effect

Darxo:
This PR is purely style and does not change how the effect works ingame.

It will make it easier for us change and swap values around as we only need to adjust them in one place now.
This will also make it much easier for submods to adjust the effects of the different levels.

### DIFF
--- a/scripts/skills/effects/rf_encumbrance_effect.nut
+++ b/scripts/skills/effects/rf_encumbrance_effect.nut
@@ -31,70 +31,29 @@ this.rf_encumbrance_effect <- ::inherit("scripts/skills/skill", {
 			text = "Current Encumbrance Level: [color=" + ::Const.UI.Color.NegativeValue + "]" + level + "[/color]"
 		});
 
-		switch (level)
+		local fatigueBuildUp += this.getFatigueOnTurnStart(level);
+		if (fatigueBuildUp)
 		{
-			case 1:
-				tooltip.push({
-					id = 10,
-					type = "text",
-					icon = "ui/icons/fatigue.png",
-					text = "[color=" + ::Const.UI.Color.NegativeValue + "]+1[/color] Fatigue at the start of every turn"
-				});
-				break;
-
-			case 2:
-				tooltip.extend([
-					{
-						id = 10,
-						type = "text",
-						icon = "ui/icons/fatigue.png",
-						text = "[color=" + ::Const.UI.Color.NegativeValue + "]+2[/color] Fatigue at the start of every turn"
-					},
-					{
-						id = 10,
-						type = "text",
-						icon = "ui/icons/fatigue.png",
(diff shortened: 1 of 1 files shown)


scripts/skills/effects/rf_encumbrance_effect.nut
```diff
@@ -31,70 +31,29 @@ this.rf_encumbrance_effect <- ::inherit("scripts/skills/skill", {
 			text = "Current Encumbrance Level: [color=" + ::Const.UI.Color.NegativeValue + "]" + level + "[/color]"
 		});
 
-		switch (level)
+		local fatigueBuildUp = -this.getFatigueRecoveryModifier(level);
```
LordMidas, line 34: I'd call this `getFatigueOnTurnStart(level)`. That way it's the function's job to return the correct value and sign and we don't have to do this inversion. It also makes it clear that it's not a modifier to the `FatigueRecovery` property but rather is actually modifying Fatigue at turn start.

scripts/skills/effects/rf_encumbrance_effect.nut
```diff
@@ -125,51 +84,53 @@ this.rf_encumbrance_effect <- ::inherit("scripts/skills/skill", {
...
-	function onUpdate( _properties )
+	function getMovementFatigueCostModifier( _encumbranceLevel )
 	{
-		switch (this.getEncumbranceLevel())
+		switch (_encumbranceLevel)
 		{
-			case 0:
-			case 1:
-				return;
+			case 0:	// Not a valid level
```
LordMidas, line 91: Minor comment: 
If by `not valid` you mean that no penalties are applied at this level then in my opinion it would be better to say `Not applicable at this level`. Because `not valid` seems to imply that it's wrong or unacceptable. Or you could say `Not a valid level for this effect` to be clear that it's only for this effect that it's not valid.

scripts/skills/effects/rf_encumbrance_effect.nut
```diff
@@ -125,51 +84,53 @@ this.rf_encumbrance_effect <- ::inherit("scripts/skills/skill", {
...
+	}
 
+	function onUpdate( _properties )
+	{
+		_properties.MovementFatigueCostAdditional += this.getMovementFatigueCostModifier(this.getEncumbranceLevel());
+	}
+
+	function onTurnStart()
+	{
+		local fatigueBuildUp = -this.getFatigueRecoveryModifier(this.getEncumbranceLevel());
```
LordMidas, line 132: When you change to `getFatigueOnTurnStart` then you can remove this `-` here.

LordMidas (changes requested):
I'd also remove the word `style` from the commit because it's a pure refactor in my opinion. The commit message is clear that it's only a change of structure/formatting.

LordMidas (approved):
Looks good. I pushed a commit updating a comment that you missed in your commit.

PR #270, opened 2024-02-16
feat: add new decisive perk

Darxo:
### Screenshot
image

image


scripts/skills/perks/perk_rf_decisive.nut
```diff
@@ -0,0 +1,160 @@
+this.perk_rf_decisive <- ::inherit("scripts/skills/skill", {
+	m = {
+		// Config
+		FatigueMult = 0.85,
+		ResolveModifier = 15,
+		DamageMult = 1.15,
+		AttractionMult = 1.1,
+
+		// Const
+		Stacks = 0,
```
LordMidas, line 10: `Stacks` really isn't "Const", it is variable. I can't think of an alternative name though. I also think that commenting these fields like this is overkill and doesn't add any value. By some logic `MaxStacks` is also "config" just like `FatigueMult`. And `FatigueMult` is a "const" as much as it is "config". I understand the desire to separate the mults/modifiers from the "stacks" type fields, in which case a simple empty line between the two would be sufficient e.g.
```squirrel
FatigueMult = 0.85,
ResolveModifier = 15,
DamageMult = 1.15,
AttractionMult = 1.1,

Stacks = 0,
MaxStacks = 3
```

mod_reforged/hooks/config/strings.nut
```diff
@@ -967,6 +968,19 @@ foreach (vanillaDesc in vanillaDescriptions)
 			]
 		}]
  	}),
+	RF_Decisive = ::UPD.getDescription({
+		Fluff = "There is no time to wait!",
+		Effects = [{
+			Type = ::UPD.EffectType.Passive,
+			Description = [
+				"Gain a stack of decisiveness whenever you end your turn without having used \'Wait\', up to a maximum of 3.",
```
LordMidas, line 976: Minor point but I think `maximum of 3 stacks.` reads better in this case than ending at `maximum of 3.` Not a huge deal either way.

scripts/skills/perks/perk_rf_decisive.nut
```diff
@@ -0,0 +1,160 @@
...
+				type = "text",
+				icon = "ui/icons/damage_dealt.png",
+				text = ::MSU.Text.colorizeMult(damageMult, {AddSign = true}) + " Damage"
+			});
+		}
+
+		return tooltip;
+	}
+
+	function onCombatStarted()
```
LordMidas, line 77: I used to add an `onCombatStarted` function like this ~~but then I took an arrow to the knee~~ but then I realized that it's redundant. Basically setting it to 0 in `onCombatFinished` is enough and it covers all cases.

LordMidas (changes requested):

scripts/skills/perks/perk_rf_decisive.nut
```diff
@@ -0,0 +1,160 @@
+this.perk_rf_decisive <- ::inherit("scripts/skills/skill", {
+	m = {
+		// Config
+		FatigueMult = 0.85,
+		ResolveModifier = 15,
```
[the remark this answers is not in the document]
Darxo: I much prefer `Modifier` for anything that is added, if there is even the slightest chance it could be made negative (e.g. by some submod). It's a more neutral term and should give the guarantee that negative values will work with it. `Add` is a bit too close to `positive`

For values which must be positive or where it doesn't make sense if they are negative (weight, wage), we can use other terms like `Bonus`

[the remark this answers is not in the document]
LordMidas: Ok I am fine with that. I will also use `Modifier` for things which are added and `Mult` for things which are multiplied.

mod_reforged/hooks/config/strings.nut
```diff
@@ -967,6 +968,19 @@ foreach (vanillaDesc in vanillaDescriptions)
...
+				::MSU.Text.colorGreen("+15") + " Resolve while you have atleast 1 stack.",
+				::MSU.Text.colorGreen("-15%") + " Fatigue Cost for skills while you have atleast 2 stacks.",
```
LordMidas, line 978: `15% reduced Fatigue Cost`

mod_reforged/hooks/config/strings.nut
```diff
@@ -967,6 +968,19 @@ foreach (vanillaDesc in vanillaDescriptions)
...
+				::MSU.Text.colorGreen("+15%") + " Damage while you have atleast 3 stacks.",
```
LordMidas, line 979: - `15% increased damage`
- The `at least` part in this line can be removed because 3 is the max.

LordMidas (changes requested):

scripts/skills/perks/perk_rf_decisive.nut
Darxo: I improved the formatting and removed the `// Const` category. 
In general I'm happy with the use of comments to separate the member variables.

scripts/skills/perks/perk_rf_decisive.nut
```diff
@@ -0,0 +1,160 @@
...
+		this.m.Type = ::Const.SkillType.Perk | ::Const.SkillType.StatusEffect;
+		this.m.Order = ::Const.SkillOrder.Perk;
+		this.m.IsActive = false;
+		this.m.IsStacking = false;
+		this.m.IsHidden = true;
+	}
+
+	function getName()
+	{
+		return (this.m.Stacks == 0) ? this.m.Name : this.m.Name + " (x" + this.m.Stacks + ")";
```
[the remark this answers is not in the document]
Darxo: It's a personal choice for readability.
If you don't want brackets used in such cases as per coders choice, then I'd like the opinion of the rest tech team and a definition on if/when they are to be used for readability and when - not.

mod_reforged/hooks/config/strings.nut
```diff
@@ -1097,6 +1098,19 @@ foreach (vanillaDesc in vanillaDescriptions)
 			]
 		}]
  	}),
+	RF_Decisive = ::UPD.getDescription({
+		Fluff = "There is no time to wait!",
+		Effects = [{
+			Type = ::UPD.EffectType.Passive,
+			Description = [
+				"Gain a stack of decisiveness whenever you end your turn without having used \'Wait\', up to a maximum of 3 stacks.",
```
LordMidas, line 1106: I think `of decisiveness` can be removed.

mod_reforged/hooks/config/strings.nut
```diff
@@ -1097,6 +1098,19 @@ foreach (vanillaDesc in vanillaDescriptions)
...
+				::MSU.Text.colorGreen("+15") + " [Resolve|Concept.Bravery] while you have atleast 1 stack.",
```
LordMidas, line 1107: `at least`

mod_reforged/hooks/config/strings.nut
```diff
@@ -1097,6 +1098,19 @@ foreach (vanillaDesc in vanillaDescriptions)
...
+				"Skills build up " + ::MSU.Text.colorGreen("15%") + " less [Fatigue|Concept.Fatigue] while you have atleast 2 stacks.",
```
LordMidas, line 1108: `at least`

mod_reforged/hooks/config/strings.nut
```diff
@@ -1097,6 +1098,19 @@ foreach (vanillaDesc in vanillaDescriptions)
...
+				"Deal " + ::MSU.Text.colorGreen("15%") + " more Damage while you have 3 stacks.",
```
LordMidas, line 1109: `damage` (don't capitalize D)

mod_reforged/hooks/config/strings.nut
```diff
@@ -1097,6 +1098,19 @@ foreach (vanillaDesc in vanillaDescriptions)
...
+				"You lose all stacks if you use \'Wait\'",
```
LordMidas, line 1110: Missing full stop at the end.

scripts/skills/perks/perk_rf_decisive.nut
```diff
@@ -0,0 +1,155 @@
+this.perk_rf_decisive <- ::inherit("scripts/skills/skill", {
+	m = {
+		// Config
+		ResolveModifier = 15,	// Gained with 1+ stacks
+		FatigueMult = 0.85,		// Gained with 2+ stacks
```
LordMidas, line 5: Would be better to rename this to `FatigueCostMult`

scripts/skills/perks/perk_rf_the_rush_of_battle.nut
LordMidas, line 1: This change to The Rush of Battle seems to have snuck into this PR by mistake.

LordMidas (changes requested):
Just a few minor changes requested. However, a few of my earlier comments are also still unresolved.

scripts/skills/perks/perk_rf_decisive.nut
```diff
@@ -0,0 +1,160 @@
...
+	function onWaitTurn()
+	{
+		this.m.Stacks = 0;
+	}
+
+	function onUpdate( _properties )
+	{
+		_properties.Bravery += this.getResolveModifier();
+		_properties.DamageTotalMult *= this.getDamageMult();
+		_properties.TargetAttractionMult *= this.getAttractionMult();	// This is on par with other vanilla affects which increase the damage output. All of them make this character more likely to be targeted
```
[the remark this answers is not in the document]
Darxo: I would not add that right now. But if someone feels strong about it they can eventually work out of concept on how to explain and show that effect to the player and retroactively add that to all skills which do that.

mod_reforged/hooks/config/strings.nut
Darxo: when do I capitalize terms and when don't I do that?

Damage is a very common battle brothers stat/term. If the word was fluff and used to describe something like an injury or a trigger condition ("when you deal any damage"), then I understand writing it lower case.

scripts/skills/perks/perk_rf_decisive.nut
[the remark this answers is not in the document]
LordMidas: This is still unresolved and the new ternaries you've added have the same issue. It is rather unorthodox to use parantheses in ternary statements like this. We already talked about this in Discord - so please remove them. I'd expect parantheses to be for some reason and looking for that reason consumes time. See example below. When I see a paranthesis opening I subconsciously expect a reason for it and that consumes effort and time when they are used without reason.
```squirrel
local myVar = (someBool == true ? val1 : val2) * 2;
local myVar = someBool == true ? val1 : val2 * 2;
```
The above code shows a good case of using parantheses because the order of operations changes with them and `myVar` ends up being different. Although the first line I'd personally write as 
```squirrel
local myVar = 2 * (someBool == true ? val1: val2);
```
but that's besides the point.

If you are having issues with readability without the parantheses it may be because of your code editor's color coding, perhaps? Because to me the version without parantheses looks quite readable. It's also a case of being used to a particular style I guess.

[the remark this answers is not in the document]
Darxo: If you think this is not a matter of personal preference but instead always must be done a certain way then you should formulate that as a hard/soft rule and we can discuss over at discord whether to add that to the SOPs.

scripts/skills/perks/perk_rf_decisive.nut
```diff
@@ -0,0 +1,135 @@
...
+
+	function getDamageMult()
+	{
+		return (this.m.Stacks >= 3) ? this.m.DamageMult : 1.0;
+	}
+
+	function getAttractionMult()
+	{
+		return (this.m.Stacks >= 3) ? this.m.AttractionMult : 1.0;
+	}
```
LordMidas, line 134: Remove parantheses

mod_reforged/hooks/config/strings.nut
```diff
@@ -1106,6 +1107,19 @@ foreach (vanillaDesc in vanillaDescriptions)
 			]
 		}]
  	}),
+	RF_Decisive = ::UPD.getDescription({
+		Fluff = "There is no time to wait!",
+		Effects = [{
+			Type = ::UPD.EffectType.Passive,
+			Description = [
+				"Gain a stack whenever you end your turn without having used \'Wait\', up to a maximum of 3 stacks.",
```
LordMidas, line 1115: `[turn|Concept.Turn]`

scripts/skills/perks/perk_rf_decisive.nut
```diff
@@ -0,0 +1,135 @@
...
+		if (damageMult != 1.0)
+		{
+			ret.push({
+				id = 12,
+				type = "text",
+				icon = "ui/icons/damage_dealt.png",
+				text = "Deal " + ::MSU.Text.colorizeMult(damageMult) + " more Damage"
+			});
+		}
+
```
LordMidas, line 73: I think we should add a warning entry here as well saying that you will lose all stacks if you use 'Wait'.

LordMidas (changes requested):

PR #276, opened 2024-02-21
refactor: always unlock drill sergeant and buff its xp gain by +20%

Darxo:


mod_reforged/hooks/entity/tactical/player.nut
```diff
@@ -24,6 +24,11 @@
 			::Const.LevelXP.push(::Const.LevelXP.top() + 4000 + 1000 * (::Const.LevelXP.len() - 11));
 		}
 
+		if (::World.Retinue.hasFollower("follower.drill_sergeant"))		// Temporary buff to vanilla drill sergeant until our Retinue Rework
```
LordMidas, line 27: Should add check for `isScenarioMode` and `_scale` just like vanilla does.

mod_reforged/hooks/entity/tactical/player.nut
```diff
@@ -24,6 +24,11 @@
...
+		{
+			_xp = _xp * ::Math.maxf(1.0, 1.1666 - 0.0166 * (this.m.Level - 1));		// 1.166666 is the exact multiplier to make a 1.2 multiplier into 1.4
```
LordMidas, line 29: Minor point: just use `_xp *= ` instead of `_xp = xp *`

LordMidas (changes requested):

mod_reforged/hooks/entity/tactical/player.nut
```diff
@@ -24,6 +24,12 @@
 			::Const.LevelXP.push(::Const.LevelXP.top() + 4000 + 1000 * (::Const.LevelXP.len() - 11));
 		}
 
+		// Temporary buff to vanilla drill sergeant until our Retinue Rework
+		if (!isScenarioMode && _scale && ::World.Retinue.hasFollower("follower.drill_sergeant"))
```
LordMidas, line 28: `isScenarioMode` isn't defined in this function because it's a hook. So you have to look at the vanilla function for the conditions which define that boolean and use those conditions manually here.

LordMidas (changes requested):

LordMidas (approved):

PR #278, opened 2024-02-23
refactor: require atleast 1 hitpoint damage for cull to trigger

Darxo:
This will prevent all kinds of weird interactions where non-damaging attacks were able to procc cull.
Not only can that be a balancing concern with cheap attacks but it can also lead to crashes with skills that never expect the target to suddenly be dead after being used on it

### DIFF
--- a/mod_reforged/hooks/config/strings.nut
+++ b/mod_reforged/hooks/config/strings.nut
@@ -951,7 +951,7 @@ foreach (vanillaDesc in vanillaDescriptions)
 			Type = ::UPD.EffectType.Passive,
 			Description = [
 				"Maximum Damage is increased by " + ::MSU.Text.colorGreen("10%") + " of the Maximum Damage of the currently equipped axe.",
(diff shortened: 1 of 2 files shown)


mod_reforged/hooks/config/strings.nut
```diff
@@ -951,7 +951,7 @@ foreach (vanillaDesc in vanillaDescriptions)
 			Type = ::UPD.EffectType.Passive,
 			Description = [
 				"Maximum Damage is increased by " + ::MSU.Text.colorGreen("10%") + " of the Maximum Damage of the currently equipped axe.",
-				"Hits to the head will instantly kill a target that has less than " + ::MSU.Text.colorRed("33%") + " [Hitpoints|Concept.Hitpoints] remaining after the hit.",
+				"Hits to the head which inflict atleast 1 damage to [Hitpoints|Concept.Hitpoints] will instantly kill a target that has less than " + ::MSU.Text.colorRed("33%") + " [Hitpoints|Concept.Hitpoints] remaining after the hit.",
```
LordMidas, line 954: `at least`

LordMidas (approved):
Looks good after that minor spelling correction.

PR #283, opened 2024-03-10
refactor: buff adrenaline effect to ignore effects of injuries

Darxo:
I tested this ingame and it seems to work

### Screenshots
image

image

image

image

### DIFF
--- a/mod_reforged/hooks/config/strings.nut
+++ b/mod_reforged/hooks/config/strings.nut
@@ -293,9 +293,10 @@ local vanillaDescriptions = [
 	 		Effects = [{
  				Type = ::UPD.EffectType.Active,
  				Description = [
-					"Unlocks the [Adrenaline|Skill+adrenaline_skill] skill which puts you first in the [turn|Concept.Turn] order for the next [round|Concept.Round], to have another [turn|Concept.Turn] before your enemies do."
+					"Unlocks the [Adrenaline|Skill+adrenaline_skill] skill which puts you first in the [turn|Concept.Turn] order for the next [round|Concept.Round], to have another [turn|Concept.Turn] before your enemies do.",
+					"Under the effects of [adrenaline|Skill+adrenaline_effect] you cannot receive and are not affected by [temporary injuries|Concept.InjuryTemporary]."
 				]
- 			}]
+ 			}],
 	 	}),
 	},
 	{
--- a/mod_reforged/hooks/skills/actives/adrenaline_skill.nut
+++ b/mod_reforged/hooks/skills/actives/adrenaline_skill.nut
@@ -2,6 +2,21 @@
 	q.create = @(__original) function()
 	{
 		__original();
+		this.m.FatigueCost = 10;	// In vanilla this is 20
 		this.m.AIBehaviorID = ::Const.AI.Behavior.ID.Adrenaline;
+	}
+
+	q.getTooltip = @(__original) function()
+	{
+		local ret = __original();
+
+		ret.push({
+			id = 11,
+			type = "text",
+			icon = "ui/icons/special.png",
+			text = ::Reforged.Mod.Tooltips.parseString("Not affected by, and cannot receive, [temporary injuries|Concept.InjuryTemporary] while active")
+		});
+
+		return ret;
 	}
 });
--- a/mod_reforged/hooks/skills/effects/adrenaline_effect.nut
+++ b/mod_reforged/hooks/skills/effects/adrenaline_effect.nut
@@ -1,0 +1,37 @@
+::Reforged.HooksMod.hook("scripts/skills/effects/adrenaline_effect", function(q) {
+	q.m.GrantsInjuryImmunity <- true;
+
+	q.create = @(__original) function()
+	{
+		__original();
+
+		this.m.Order = ::Const.SkillOrder.TemporaryInjury - 100;	// This effect needs to flip the IsAffectedByInjuries flag before any of the injuries had the time to inflict their effects
+	}
+
+	q.getTooltip = @(__original) function()
+	{
+		local ret = __original();
+
+		if (this.m.GrantsInjuryImmunity)
+		{
+			ret.push({
+				id = 11,
+				type = "text",
+				icon = "ui/icons/special.png",
+				text = ::Reforged.Mod.Tooltips.parseString("Not affected by, and cannot receive, [temporary injuries|Concept.InjuryTemporary]"),
+			});
+		}
+
+		return ret;
+	}
+
+	q.onUpdate = @(__original) function( _properties )
+	{
+		__original(_properties);
+
+		if (this.m.TurnsLeft != 0 && this.m.GrantsInjuryImmunity)
+		{
+			_properties.IsAffectedByInjuries = false;
+		}
+	}
+});


mod_reforged/hooks/skills/effects/adrenaline_effect.nut
```diff
@@ -0,0 +1,37 @@
+::Reforged.HooksMod.hook("scripts/skills/effects/adrenaline_effect", function(q) {
+	q.m.GrantsInjuryImmunity <- true;
+
+	q.create = @(__original) function()
+	{
+		__original();
+
+		this.m.Order = ::Const.SkillOrder.PermanentInjury - 100;	// This effect needs to flip the IsAffectedByInjuries flag before any of the injuries had the time to inflict their effects
```
LordMidas, line 8: Why the `-100`? Setting it to PermanentInjury seems good enough.

mod_reforged/hooks/skills/effects/adrenaline_effect.nut
```diff
@@ -0,0 +1,37 @@
...
+		return ret;
+	}
+
+	q.onUpdate = @(__original) function( _properties )
+	{
+		__original(_properties);
+
+		if (this.m.TurnsLeft != 0 && this.m.GrantsInjuryImmunity)
+		{
+			_properties.IsAffectedByInjuries = false;
```
LordMidas, line 34: This will also make it impossible for the character to receive injuries (actor.nut `onDamageReceived` function checks this flag in the target's `CurrentProperties` and if false then it doesn't apply an injury). Adding a check for whether this entity is the currently active entity is also not enough because that will mean he can't receive injury from enemy's riposte.

So you need to do it differently i.e. reset this flag to its original value before taking damage. For ideas consider looking at my implementation of the `rf_undead_injury_receiver_effect` which does the opposite i.e. makes undead be affected by injuries but only receive them if attacker has crippling strikes perk.

mod_reforged/hooks/skills/effects/adrenaline_effect.nut
```diff
@@ -0,0 +1,37 @@
...
+	{
+		local ret = __original();
+
+		if (this.m.GrantsInjuryImmunity)
+		{
+			ret.push({
+				id = 11,
+				type = "text",
+				icon = "ui/icons/special.png",
+				text = "Ignore the effects of non-permanent injuries"
```
LordMidas, line 21: `::Reforged.Mod.Tooltips.parseString("Ignore the effects of [temporary injuries|Concept.InjuryTemporary]")`

mod_reforged/hooks/config/strings.nut
```diff
@@ -270,7 +270,7 @@ local vanillaDescriptions = [
 	 		Effects = [{
  				Type = ::UPD.EffectType.Active,
  				Description = [
-					"Unlocks the [Adrenaline|Skill+adrenaline_skill] skill which puts you first in the [turn|Concept.Turn] order for the next [round|Concept.Round], to have another [turn|Concept.Turn] before your enemies do."
+					"Unlocks the [Adrenaline|Skill+adrenaline_skill] skill which puts you first in the [turn|Concept.Turn] order for the next [round|Concept.Round], to have another [turn|Concept.Turn] before your enemies do and ignore the effects of non-permanent injuries."
```
LordMidas, line 273: `ignore the effects of [temporary injuries|Concept.InjuryTemporary].`

LordMidas (changes requested):

mod_reforged/hooks/skills/effects/adrenaline_effect.nut
Darxo: I felt putting it slightly before permanent injuries was the safest option just in case we want it to trigger before permanent injuries at some point.

mod_reforged/hooks/skills/actives/adrenaline_skill.nut
```diff
@@ -2,6 +2,21 @@
 	q.create = @(__original) function()
 	{
 		__original();
+		this.m.FatigueCost = 10;	// In vanilla this is 20
```
LordMidas, line 5: This change should also be mentioned in the commit message. An example could be: `refactor: buff adrenaline by reducing fatigue cost and ignoring injuries` that fits within the character limit for commit messages.

mod_reforged/hooks/skills/effects/adrenaline_effect.nut
LordMidas: Over at MSU I've been trying to figure out a clean solution for modularizing the receiving of injuries from injuries applying their effects which also is compatible with vanilla and so on. But there are certain edge cases so it will need some more time/effort to find the best solution.

For now, I suggest we make this a feature of Indomitable i.e. under the effects of indomitable you cannot receive and are not affected by already present temporary injuries (so basically the same as the vanilla iron will potion).

mod_reforged/hooks/skills/effects/adrenaline_effect.nut
LordMidas: Permanent Injuries are not affected by `IsAffectedByInjuries` flag. But it's fine to set the order as you have done for future proofing.

mod_reforged/hooks/skills/effects/adrenaline_effect.nut
LordMidas: In Discord we approved this "feature". So you can update the strings and tooltips in this PR to say that it also makes you immune to receiving injuries.

mod_reforged/hooks/skills/effects/adrenaline_effect.nut
Darxo: Thinking more about it I changed it from permanentInjury to temporaryInjury.

Adrenaline should NEVER affect permanent injuries, so I can't think of any scenario where that order should ever be relevant

LordMidas: Yes that's fine by me.

mod_reforged/hooks/config/strings.nut
```diff
@@ -293,7 +293,7 @@ local vanillaDescriptions = [
 	 		Effects = [{
  				Type = ::UPD.EffectType.Active,
  				Description = [
-					"Unlocks the [Adrenaline|Skill+adrenaline_skill] skill which puts you first in the [turn|Concept.Turn] order for the next [round|Concept.Round], to have another [turn|Concept.Turn] before your enemies do."
+					"Unlocks the [Adrenaline|Skill+adrenaline_skill] skill which puts you first in the [turn|Concept.Turn] order for the next [round|Concept.Round], to have another [turn|Concept.Turn] before your enemies do and ignore [temporary injuries|Concept.InjuryTemporary]."
```
LordMidas, line 296: We need to some how also convey the information here that it makes you immune to receiving injuries. Either that has to be done explicitly in words, or via a nested tooltip of the adrenaline effect here. So we could do:
```squirrel
"Unlocks the [Adrenaline|Skill+adrenaline_skill] skill which puts you first in the [turn|Concept.Turn] order for the next [round|Concept.Round], to have another [turn|Concept.Turn] before your enemies do.",
"Under the effects of [adrenaline|Skill+adrenaline_effect] you cannot receive and are not affected by [temporary injuries|Concept.InjuryTemporary]."
```
Or
```squirrel
"Unlocks the [Adrenaline|Skill+adrenaline_skill] skill which puts you first in the [turn|Concept.Turn] order for the next [round|Concept.Round], and grants the [Adrenaline|Skill+adrenaline_effect] effect which makes you immune to [injuries|Concept.InjuryTemporary].",
```

LordMidas (approved):

PR #284, opened 2024-03-10
refactor: remove cap and handicap from fearsome's calculation

Darxo:
image

### DIFF
--- a/mod_reforged/hooks/config/strings.nut
+++ b/mod_reforged/hooks/config/strings.nut
@@ -238,6 +238,20 @@ local vanillaDescriptions = [
 	 	}),
 	},
 	{
+		ID = "perk.fearsome",
+		Key = "Fearsome",
+		Description = ::UPD.getDescription({
+			Fluff = "Make them scatter and flee!",
+	 		Effects = [{
+				Type = ::UPD.EffectType.Passive,
+				Description = [
+					"Reduce the [Hitpoint|Concept.Hitpoints] damage threshold for triggering [Morale Checks|Concept.Morale] with your attacks to 1.",
+					"These morale checks now have an additional penalty to [Resolve|Concept.Bravery] of " + ::MSU.Text.colorGreen("20%") + " of your [Resolve|Concept.Resolve].",
+				]
+ 			}],
+	 	}),
+	},
+	{
 		ID = "perk.footwork",
 		Key = "Footwork",
 		Description = ::UPD.getDescription({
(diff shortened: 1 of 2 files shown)


mod_reforged/hooks/config/strings.nut
```diff
@@ -237,6 +237,20 @@ local vanillaDescriptions = [
...
+	{
+		ID = "perk.fearsome",
+		Key = "Fearsome",
+		Description = ::UPD.getDescription({
+			Fluff = "Make them scatter and flee!",
+	 		Effects = [{
+				Type = ::UPD.EffectType.Passive,
+				Description = [
+					"Reduce the hitpoint threshold for triggering [Morale Checks|Concept.Morale] with your attacks to 1.",
+					"Increase the penalty of those checks by " + ::MSU.Text.colorGreen("20%") + " of your [Resolve|Concept.Resolve]"
```
LordMidas, line 249: I think "penalty of those checks" isn't a good way to phrase this. Something like this would be better:

The target has a penalty to Resolve equal to 20% of your Resolve for these morale checks.

LordMidas (changes requested):

mod_reforged/hooks/config/strings.nut
Darxo: But that is the exact phrasing of the vanilla fearsome perk.

LordMidas: I mean that "penalty of those checks" is a very awkward way of phrasing it. It sounds out of place with the standard of the rest of the language of this game. Secondly, "penalty of those checks" is a rather ambiguous statement. "Increase the penalty" also makes it sound like there was already some penalty which is further increased by 20% of your Resolve. The players are not aware of any such penalty even though we know that it exists. So I'd just say the one I suggested above is more in line with the style and is clear about what the penalty is about. The vanilla description is also similar to that. Just say something like that these morale checks have a penalty to resolve of 20% of your resolve.

Darxo: How about this?
`Those checks have an additional penalty equal to ::MSU.Text.colorGreen("20%") + " of your [Resolve|Concept.Resolve]`
Main difference is swapping out "increase" for "additional". I think this is more vague about whether there is a penalty already existing and a newer player will not put too much thought into that.

LordMidas: This is better. But it is important to mention what the penalty is to, so I would word it as: `additional penalty to Resolve equal to`.

Also the first line should be `Reduce the [hitpoints|Concept.Hitpoints] damage threshold` instead of `hitpoints threshold` i.e. add the word `damage`.

LordMidas: So my suggestion is:
```squirrel
"Reduce the [Hitpoints|Concept.Hitpoints] damage threshold for triggering [Morale Checks|Concept.Morale] with your attacks to 1.",
"These morale checks now have an additional penalty to [Resolve|Concept.Bravery] of " + ::MSU.Text.colorGreen("20%") + " of your [Resolve|Concept.Resolve]"
```

LordMidas (approved):
I pushed one commit to add hyperlink to morale check in the string. Looks good.

PR #285, opened 2024-03-10
Refactor rework taunt effect

Darxo:
I tested this ingame. It seems to work correctly.

**Screenshots:**
image

image

image

### DIFF
--- a/mod_reforged/hooks/config/strings.nut
+++ b/mod_reforged/hooks/config/strings.nut
@@ -305,7 +305,8 @@ local vanillaDescriptions = [
 	 		Effects = [{
  				Type = ::UPD.EffectType.Active,
  				Description = [
-					"Unlocks the [Taunt|Skill+taunt] skill which makes the targeted opponent take offensive actions instead of defensive ones, and attack the taunting character over another, potentially more vulnerable one."
+					"Unlocks the [Taunt|Skill+taunt] skill which makes the targeted opponent take offensive actions instead of defensive ones, and attack the taunting character over another, potentially more vulnerable one.",
+					"When used on an adjacent target, reduces its [Melee Defense|Concept.MeleeDefense] and [Ranged Defense|Concept.RangeDefense] by " + ::MSU.Text.colorGreen("20%") + " of the user's [Resolve|Concept.Bravery]."
 				]
  			}]
 	 	}),
--- a/mod_reforged/hooks/msu.nut
+++ b/mod_reforged/hooks/msu.nut
@@ -81,12 +81,26 @@
 }
 
 ::logInfo("Reforged::MSU -- adding ::MSU.Text.colorizeMult");
-::MSU.Text.colorizeMult <- function( _value, _options = null )
+::MSU.Text.colorizeMult <- function( _value, _options = null )		// will turn values like 1.45 into a formatted and colored 45%
 {
 	if (_options == null) _options = {};
 	if (!("AddSign" in _options)) _options.AddSign <- false;
 	_options.AddPercent <- true;
 	return this.colorizeValue((_value - 1.0) * 100, _options);
+}
+
+::logInfo("Reforged::MSU -- adding ::MSU.Text.colorizeFraction");
+::MSU.Text.colorizeFraction <- function( _value, _options = null )	// will turn values like 0.2 into a formatted and colored 20%
+{
+	if (_value < 0)
+	{
+		throw ::MSU.Exception.InvalidValue(_value);
+	}
+
+	if (_options == null) _options = {};
+	if (!("AddSign" in _options)) _options.AddSign <- false;
+	_options.AddPercent <- true;
+	return this.colorizeValue(_value * 100, _options);
 }
 
 ::logInfo("Reforged::MSU -- adding ::MSU.Tile.getNeighbors");
--- a/mod_reforged/hooks/skills/actives/taunt.nut
+++ b/mod_reforged/hooks/skills/actives/taunt.nut
@@ -1,8 +1,73 @@
 ::Reforged.HooksMod.hook("scripts/skills/actives/taunt", function(q) {
+	q.m.DefenseModifierFraction <- 0.2;	// This percentage of the users resolve is removed from the targets defense
+	q.m.MaxRangeForDefenseDebuff <- 1;	// The target of your taunt must be at most this many tiles from you to receive the defense debuff
+
+	q.getTooltip = @(__original) function()
+	{
+		local ret = __original();
+
+		if (this.m.DefenseModifierFraction != 0)
+		{
+			ret.push({
+				id = 10,
+				type = "text",
+				icon = "ui/icons/special.png",
+				text = ::Reforged.Mod.Tooltips.parseString("When used on an adjacent target, reduces its [Melee Defense|Concept.MeleeDefense] and [Ranged Defense|Concept.RangeDefense] by " + ::MSU.Text.colorizeValue(this.calculateDefenseModifier()) + " (" + ::MSU.Text.colorizeFraction(this.m.DefenseModifierFraction) + " of your current [Resolve|Concept.Bravery])")
+			});
+		}
+
+		return ret;
+	}
+
 	q.onVerifyTarget = @(__original) function( _originTile, _targetTile )
 	{
 		local ret = __original(_originTile, _targetTile);
 		if (ret && _targetTile.getEntity().getSkills().hasSkill("effects.taunted"))	return false;
 		return ret;
+	}
+
+	q.onUse = @(__original) function( _user, _targetTile )
+	{
+		local ret = __original(_user, _targetTile);
+		if (!ret) return ret;
+
+		local target = _targetTile.getEntity();
+		if (this.getContainer().getActor().getTile().getDistanceTo(_targetTile) <= this.m.MaxRangeForDefenseDebuff)
+		{
+			local tauntEffect = target.getSkills().getSkillByID("effects.taunted");
+			tauntEffect.m.DefenseModifier = this.calculateDefenseModifier();
+		}
+		target.getSkills().update();	// Otherwise the defense debuff will not be immediately visible in the combat tooltip
+
+		return true;
+	}
+
+// MSU Functions
+	q.getNestedTooltip = @() function()
+	{
+		if (this.getContainer().getActor().getID() != ::MSU.getDummyPlayer().getID())
+			return this.getTooltip();
+
+		// Replicate the getTooltip method of class in a way that works with dummy players
+		local ret = this.skill.getTooltip();
+
+		if (this.m.DefenseModifierFraction != 0)
+		{
+			ret.push({
+				id = 10,
+				type = "text",
+				icon = "ui/icons/special.png",
+				text = ::Reforged.Mod.Tooltips.parseString("When used on an adjacent target, reduces its [Melee Defense|Concept.MeleeDefense] and [Ranged Defense|Concept.RangeDefense] by " + ::MSU.Text.colorizeFraction(this.m.DefenseModifierFraction) + " of your current [Resolve|Concept.Bravery]")
+			});
+		}
+
+		return ret;
+	}
+
+// New Functions
+	q.calculateDefenseModifier <- function()
+	{
+		local defenseModifier = -1.0 * this.m.DefenseModifierFraction * this.getContainer().getActor().getCurrentProperties().getBravery();
+		return ::Math.min(0, defenseModifier);	// This modifier may never be positive
 	}
 });
--- a/mod_reforged/hooks/skills/effects/taunted_effect.nut
+++ b/mod_reforged/hooks/skills/effects/taunted_effect.nut
@@ -1,4 +1,6 @@
 ::Reforged.HooksMod.hook("scripts/skills/effects/taunted_effect", function(q) {
+	q.m.DefenseModifier <- 0;
+
 	q.create = @(__original) function()
 	{
 		__original();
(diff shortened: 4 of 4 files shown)


mod_reforged/hooks/skills/actives/taunt.nut
```diff
@@ -1,8 +1,48 @@
...
+	{
+		__original(_user, _targetTile);
+
+		local target = _targetTile.getEntity();
+		if (this.getContainer().getActor().getTile().getDistanceTo(_targetTile) <= this.m.MaxRangeDefenseDebuff)
+		{
+			local tauntEffect = target.getSkills().getSkillByID("effects.taunted");
+			tauntEffect.m.DefenseModifier = this.calculateDefenseModifier();
+		}
+		target.getSkills().update();	// Otherwise the defense debuff will not be immediately visible in the combat tooltip
```
LordMidas, line 39: Must return `true` here. In fact the "fully" correct way of doing this is to first do `local ret = __original(_user, _targetTile)` then `if (!ret)` you immediately `return ret` and get out of this function. Then below that do all this logic and then at the end return `true`.

mod_reforged/hooks/skills/actives/taunt.nut
```diff
@@ -1,8 +1,48 @@
...
+	}
+
+// New Functions
+	q.calculateDefenseModifier <- function()
+	{
+		local defenseModifier = -1.0 * this.m.DefenseModifierFraction * this.getContainer().getActor().getCurrentProperties().Bravery;
```
LordMidas, line 45: `properties.getBravery()` instead of `properties.Bravery`

mod_reforged/hooks/skills/effects/taunted_effect.nut
```diff
@@ -1,13 +1,15 @@
 ::Reforged.HooksMod.hook("scripts/skills/effects/taunted_effect", function(q) {
+	q.m.DefenseModifier <- 0;
+
 	q.create = @(__original) function()
 	{
 		__original();
-        this.m.Description = "This character is taunted by another character and is much more likely to engage and attack them.";
+		this.m.Description = "This character is taunted by another character and is much more likely to engage and attack them.";
```
LordMidas, line 7: For cleanliness I'd remove this whitespace change from this commit. It makes going through commit history later especially with "blame file" more polluted. We should do a full codebase conversion of spaces to tabs in a single commit later.

mod_reforged/hooks/skills/effects/taunted_effect.nut
```diff
@@ -19,11 +21,50 @@
...
+
+		if (this.getRangedDefenseModifier() != 0)
+		{
+			ret.push({
+				id = 14,
+				type = "text",
+				icon = "ui/icons/ranged_defense.png",
+				text = ::MSU.Text.colorizeValue(this.getRangedDefenseModifier()) + " Ranged Defense"
+			});
+		}
```
LordMidas, line 42: These should only apply if the `getTauntedTarget()` is not null.

mod_reforged/hooks/skills/effects/taunted_effect.nut
```diff
@@ -19,11 +21,50 @@
...
 		return ret;
 	}
 
-	q.getTauntedTarget <- function()	// New helper-function to get the taunter
+	q.onUpdate = @(__original) function( _properties )
+	{
+		__original(_properties);
+
+		_properties.MeleeDefense += this.getMeleeDefenseModifier();
+		_properties.RangedDefense += this.getRangedDefenseModifier();
```
LordMidas, line 52: These should only apply if the `getTauntedTarget()` is not null.

mod_reforged/hooks/skills/effects/taunted_effect.nut
```diff
@@ -19,11 +21,50 @@
...
+	}
+
+// New Functions
+	q.getTauntedTarget <- function()
```
LordMidas, line 56: I'd say it should be `getTaunter` because `TauntedTarget` is this guy who has been taunted.

mod_reforged/hooks/skills/effects/taunted_effect.nut
```diff
@@ -19,11 +21,50 @@
 			});
 		}
 
+		if (this.getMeleeDefenseModifier() != 0)
+		{
+			ret.push({
+				id = 13,
+				type = "text",
+				icon = "ui/icons/melee_defense.png",
+				text = ::MSU.Text.colorizeValue(this.getMeleeDefenseModifier()) + " Melee Defense"
```
LordMidas, line 30: Please add hyperlinks for the concepts such as melee defense and ranged defense.

mod_reforged/hooks/skills/actives/taunt.nut
```diff
@@ -1,8 +1,48 @@
...
+	{
+		local ret = __original();
+
+		if (this.m.DefenseModifierFraction != 0)
+		{
+			ret.push({
+				id = 12,
+				type = "text",
+				icon = "ui/icons/special.png",
+				text = "When used on an adjacent target, reduce its Melee Defense and Ranged Defense by " + ::MSU.Text.colorizeFraction(this.m.DefenseModifierFraction) + " of your current Resolve (" + ::MSU.Text.colorizeValue(this.calculateDefenseModifier()) + ")"
```
LordMidas, line 15: Please add hyperlinks for the defense concepts and Resolve

LordMidas, line 15: I like that you show the actual modifier that will be applied. But it is slightly confusing to place it at the end in parantheses after `current Resolve` as it can be taken to mean that this value represents your current resolve. Maybe to clarify you could add the word `by` before it e.g. `(by x)`. Or you can say it like: "When used on an adjacent target, reduce its Melee Defense and Ranged Defense by X (20% of your current Resolve)." so the actual value is shown first and its explanation that it is 20% of your resolve is given in the parantheses. What are your thoughts?

mod_reforged/hooks/msu.nut
```diff
@@ -81,14 +81,23 @@
...
+::MSU.Text.colorizeMult <- function( _value, _options = null )		// will turn values like 1.45 into a formatted and colored 45%
 {
 	if (_options == null) _options = {};
 	if (!("AddSign" in _options)) _options.AddSign <- false;
 	_options.AddPercent <- true;
 	return this.colorizeValue((_value - 1.0) * 100, _options);
 }
 
+::logInfo("Reforged::MSU -- adding ::MSU.Text.colorizeFraction");
+::MSU.Text.colorizeFraction <- function( _value, _options = null )	// will turn values like 0.2 into a formatted and colored 20%
```
LordMidas, line 93: `_value` here must be between 0 and 1, so for completeness sake you could add a requirement for that in the beginning of the function i.e.
```squirrel
if (_value < 0 || _value > 1)
   throw ::MSU.Exception.InvalidValue(_value);
```

mod_reforged/hooks/config/strings.nut
```diff
@@ -297,7 +297,8 @@ local vanillaDescriptions = [
 	 		Effects = [{
  				Type = ::UPD.EffectType.Active,
  				Description = [
-					"Unlocks the [Taunt|Skill+taunt] skill which makes the targeted opponent take offensive actions instead of defensive ones, and attack the taunting character over another, potentially more vulnerable one."
+					"Unlocks the [Taunt|Skill+taunt] skill which makes the targeted opponent take offensive actions instead of defensive ones, and attack the taunting character over another, potentially more vulnerable one.",
+					"When used on an adjacent target, reduce its [Melee Defense|Concept.MeleeDefense] and [Ranged Defense|Concept.RangeDefense] by " + ::MSU.Text.colorGreen("20%") + " of the users [Resolve|Concept.Resolve]."
```
LordMidas, line 301: - `user's`
- also `reduces` instead of `reduce` would be more appropriate to further describe the effect of the skill like the first paragraph which says it `makes` the target do something.

mod_reforged/hooks/skills/actives/taunt.nut
LordMidas, line 15: Same comment as the perk description i.e. `user's` and `reduces`

LordMidas (changes requested):

mod_reforged/hooks/skills/actives/taunt.nut
LordMidas, line 15: This will also need special handling for nested tooltip of this effect now, because you put actor dependent value of the defense modifier in the tooltip. I'd suggest adding a `getNestedTooltip` function and check if the actor is the MSU dummy player and if yes then return a tooltip entry without that information. See for example my implementation of nested tooltip of the `rf_between_the_eyes` skill.

mod_reforged/hooks/msu.nut
Darxo: The Fraction concept I have in mind is allowed to go past 1.0.
A Fraction of 1.3 is equal to 130%. E.g. Fearsome would now can 130% of Resolve as a Penalty which is absolutely allowed

I put for now a restriction for not going lower than 0.0 like you suggested

mod_reforged/hooks/skills/effects/taunted_effect.nut
Darxo: Note that I put this specific change in a commit named `refactor(style): improve formatting of taunted_effect` because there were a few other things that bugged me about that class. In this context I find it reasonable to replace whitespaces.

mod_reforged/hooks/skills/actives/taunt.nut
Darxo: I will use your last suggestion

mod_reforged/hooks/skills/actives/taunt.nut
```diff
@@ -28,21 +28,46 @@
...
+		// Replicate the getTooltip method of class in a way that works with dummy players
+		local ret = this.skill.getTooltip();
+
+		if (this.m.DefenseModifierFraction != 0)
+		{
+			ret.push({
+				id = 10,
+				type = "text",
+				icon = "ui/icons/special.png",
+				text = ::Reforged.Mod.Tooltips.parseString("When used on an adjacent target, reduces its [Melee Defense|Concept.MeleeDefense] and [Ranged Defense|Concept.RangeDefense] by " + ::MSU.Text.colorizeFraction(this.m.DefenseModifierFraction) + " of your current [Resolve|Concept.Resolve]")
```
LordMidas, line 60: `Concept.Bravery`

mod_reforged/hooks/skills/actives/taunt.nut
```diff
@@ -28,21 +28,46 @@
...
 			local tauntEffect = target.getSkills().getSkillByID("effects.taunted");
 			tauntEffect.m.DefenseModifier = this.calculateDefenseModifier();
 		}
 		target.getSkills().update();	// Otherwise the defense debuff will not be immediately visible in the combat tooltip
+
+		return true;
+	}
+
+// MSU Functions
+	q.getNestedTooltip = @(__original) function()
```
LordMidas, line 46: Should be `@()` because you aren't using the `__original()` in the new function definition.

mod_reforged/hooks/skills/actives/taunt.nut
```diff
@@ -1,8 +1,73 @@
...
+	{
+		local ret = __original();
+
+		if (this.m.DefenseModifierFraction != 0)
+		{
+			ret.push({
+				id = 10,
+				type = "text",
+				icon = "ui/icons/special.png",
+				text = ::Reforged.Mod.Tooltips.parseString("When used on an adjacent target, reduces its [Melee Defense|Concept.MeleeDefense] and [Ranged Defense|Concept.RangeDefense] by " + ::MSU.Text.colorizeValue(this.calculateDefenseModifier()) + " (" + ::MSU.Text.colorizeFraction(this.m.DefenseModifierFraction) + " of your current [Resolve|Concept.Resolve])")
```
LordMidas, line 15: `Concept.Bravery`

mod_reforged/hooks/skills/effects/taunted_effect.nut
```diff
@@ -1,29 +1,73 @@
...
+
+		if (this.getTauntSource() != null)
+		{
+			_properties.MeleeDefense += this.getMeleeDefenseModifier();
+			_properties.RangedDefense += this.getRangedDefenseModifier();
+		}
+	}
+
+// New Functions
+	q.getTauntSource <- function()
```
LordMidas, line 59: With the idea of tracking the `taunt source` for this effect, we could "improve" the vanilla effect's behavior to be more sensible in this case:
1. Bro A taunts the target.
2. Bro A dies before target's turn ends.
3. The target still has the taunt effect (it only gets removed on turn end). But in its tooltip it doesn't show anything except its name and description now.
4. Bro B now taunts the target. Now suddenly the same taunted effect shows the new info and also applies its debuffs.

This doesn't look very clean in vanilla, and will look even more unclean in Reforged with these new debuffs which are important to visualize. So I suggest that we follow the following approach:
1. Add a field in `taunted_effect`: `this.m.Source`.
2. Add a field in `taunt.nut` (the active skill): `this.m.Target`.

Then in the `onUse` of taunt, you set `this.m.Target` to the `taunted_effect` in the form of `::MSU.asWeakTableRef`. Similarly in the `taunted_effect` of the target you set `this.m.Source` to the taunt user as weak table ref.

Then in both the `onRemoved` and `onDeath` functions of both skills you remove the `taunted_effect` from the target if the target is alive, and reset the `this.m.Target` of the the taunt skill back to null. Similarly in `onCombatFinished` you set that back to null.

I'll be happy to make this change directly to this branch if you'd rather prefer that I do it.

mod_reforged/hooks/config/strings.nut
```diff
@@ -305,7 +305,8 @@ local vanillaDescriptions = [
 	 		Effects = [{
  				Type = ::UPD.EffectType.Active,
  				Description = [
-					"Unlocks the [Taunt|Skill+taunt] skill which makes the targeted opponent take offensive actions instead of defensive ones, and attack the taunting character over another, potentially more vulnerable one."
+					"Unlocks the [Taunt|Skill+taunt] skill which makes the targeted opponent take offensive actions instead of defensive ones, and attack the taunting character over another, potentially more vulnerable one.",
+					"When used on an adjacent target, reduces its [Melee Defense|Concept.MeleeDefense] and [Ranged Defense|Concept.RangeDefense] by " + ::MSU.Text.colorGreen("20%") + " of the user's [Resolve|Concept.Resolve]."
```
LordMidas, line 309: `Concept.Bravery`

LordMidas (changes requested):

mod_reforged/hooks/skills/effects/taunted_effect.nut
Darxo: In vanilla one guy can taunt multiple enemies. So the variable should be an array.

This all reminds me on the parent child framework for skills that I suggested a while ago and exemplary implemented in this commit here: https://github.com/Battle-Modders/mod-reforged/commit/1d3d85a6b8ad40c19f38add2af592544864fbc13

I don't want to force such a general system into this specific skill and have to deal with a back and forth of reviews for that.
So if you want it still, please go ahead and do it.

LordMidas: I think a simpler solution can be for now to just add an isHidden function which checks for there being a valid taunt source present. That will hide the effect if the taunt source dies before this guy's turn ends - and that achieves the player-side visibility goal. I've added that in a commit.

I've also pushed a "fix" commit that adds check for taunt source being not null via `::MSU.isNull` which is necessary as it's an instance of WeakTableRef, and also that taunt source is alive.

LordMidas (approved):

PR #286, opened 2024-03-13
refactor: invert coloring of several values in perk description

Darxo:
All of these effects are positive, but were using the negative color.

I did not do a complete sweep here, nor did I fix any instances of values which are green, but should be red.
This is just a first rough correction with more to go over time as we discover more cases.


LordMidas:
I have rebased the branch on `development` and have also included the new perk descriptions added since the original time this PR was opened.

LordMidas (approved):

PR #291, opened 2024-03-17
Feat perk inspire

Darxo:


mod_reforged/hooks/config/strings.nut
```diff
@@ -1290,6 +1291,20 @@ foreach (vanillaDesc in vanillaDescriptions)
...
+		Effects = [{
+			Type = ::UPD.EffectType.Passive,
+			Description = [
+				"Once per turn, after you use a skill targeting an adjacent ally, recover " + ::MSU.Text.colorGreen("2") + " [Action Points|Concept.ActionPoints]."
+			]
+		},
+		{
+			Type = ::UPD.EffectType.Active,
+			Description = [
+				"Gain the [Inspire|Skill+rf_inspire_skill] skill that triggers a positive morale check for an ally within 3 tiles with a bonus of 10% of your [Resolve|Concept.Bravery]. The same ally can only be inspired one time per turn."
```
LordMidas, line 1304: `10%` should be green.

LordMidas, line 1304: The last sentence would be better as: `A character can only be inspired once per turn.`

mod_reforged/hooks/config/strings.nut
```diff
@@ -1290,6 +1291,20 @@ foreach (vanillaDesc in vanillaDescriptions)
 			]
 		}]
  	}),
+	RF_Inspire = ::UPD.getDescription({
+		Effects = [{
```
LordMidas, line 1295: Improve formatting by pushing the `{` to the next line because we have multiple entries in the array.

scripts/skills/actives/rf_inspire_skill.nut
```diff
@@ -0,0 +1,104 @@
...
+
+		// Const
+		InspireFlag = "RF_InspireSkillFlag"
+	},
+
+	function create()
+	{
+		this.m.ID = "actives.rf_inspire";
+		this.m.Name = "Inspire";
+		this.m.Description = "Inspire an ally to raise their current Morale. The same target can only be inspired once per turn. Cannot be used on fleeing or stunned allies.";
```
LordMidas, line 14: `A character can only be inspired once per turn.`

scripts/skills/actives/rf_inspire_skill.nut
```diff
@@ -0,0 +1,104 @@
...
+	function getTooltip()
+	{
+		local tooltip = this.skill.getDefaultUtilityTooltip();
+
+		tooltip.extend([
+			{
+				id = 8,
+				type = "text",
+				icon = "ui/icons/bravery.png",
+				text = "Trigger a positive morale check with a bonus of " + ::MSU.Text.colorizeFraction(this.m.InspireBonusFraction) + " (" + ::MSU.Text.colorizeValue(this.getInspireBonus()) + ") of your Resolve on the target."
```
LordMidas, line 43: Resolve should be hyperlinked

LordMidas, line 43: Remove `.` from the end of the tooltip text

scripts/skills/actives/rf_inspire_skill.nut
```diff
@@ -0,0 +1,104 @@
...
+		this.m.ActionPointCost = 3;
+		this.m.FatigueCost = 15;
+		this.m.MinRange = 1;
+		this.m.MaxRange = 3;
+		// this.m.AIBehaviorID = ::Const.AI.Behavior.ID.RF_Inspire;	// Todo Implement
+	}
+
+	function getTooltip()
+	{
+		local tooltip = this.skill.getDefaultUtilityTooltip();
```
LordMidas, line 36: Would be better to use `ret` as variable name.

scripts/skills/actives/rf_inspire_skill.nut
```diff
@@ -0,0 +1,104 @@
...
+	{
+		if (!this.skill.onVerifyTarget(_originTile, _targetTile)) return false;
+
+		local target = _targetTile.getEntity();
+
+		if (this.getContainer().getActor().getFaction() != target.getFaction()) return false;
+		if (target.getCurrentProperties().IsStunned) return false;
+
+		if (target.getMoraleState() >= ::Const.MoraleState.Confident) return false;
+		if (target.getMoraleState() >= target.m.MaxMoraleState) return false;	// This might be confusing for the player because it's not mentioned in the tooltip but it's just logical
```
LordMidas, line 66: Shouldn't these 3 conditions also be part of `__canBeInspired`?

scripts/skills/actives/rf_inspire_skill.nut
```diff
@@ -0,0 +1,104 @@
...
+		return this.__canBeInspired(target);
+	}
+
+	function onUse( _user, _targetTile )
+	{
+		local target = _targetTile.getEntity();
+
+		target.checkMorale(1, this.getInspireBonus(), ::Const.MoraleCheckType.Default, "status_effect_56");	// TODO replace status effect
+
+		::Tactical.Entities.getFlags().set(target.getID() + this.m.InspireFlag, ::Time.getRound());	// We use entity flags to keep track of this skills usage instead of a dummy skill
```
LordMidas, line 77: `skill's usage`

scripts/skills/actives/rf_inspire_skill.nut
```diff
@@ -0,0 +1,104 @@
...
+
+		// this.spawnIcon("rf_inspire_skill", _targetTile);	// TODO decide on whether to spawn this?
```
LordMidas, line 79: Yes I think it should spawn it. The icon should be the icon of this perk so `perk_rf_inspire`.

scripts/skills/perks/perk_rf_inspire.nut
```diff
@@ -0,0 +1,69 @@
...
+		this.m.Description = ::Const.Strings.PerkDescription.RF_Inspire;
+		this.m.Icon = "ui/perks/rf_inspire.png";	// TODO
+		this.m.Type = ::Const.SkillType.Perk;
+		this.m.Order = ::Const.SkillOrder.Perk;
+		this.m.IsActive = false;
+		this.m.IsStacking = false;
+		this.m.IsHidden = false;
+	}
+
+	function isHidden()
```
LordMidas, line 24: `function isHidden()` is unnecessary because this does not have the `StatusEffect` type.

scripts/skills/perks/perk_rf_inspire.nut
```diff
@@ -0,0 +1,69 @@
...
+	function onTurnStart()
+	{
+		this.m.IsSpent = false;
+	}
+
+// MSU Functions
+	// Implement Action Point recovery when targeting adjacent allies
+	function onAnySkillExecuted( _skill, _targetTile, _targetEntity, _forFree )
+	{
+		if (this.m.IsSpent) return;
```
LordMidas, line 50: `_targetEntity` can be `null`. So you should add a guard clause for `_targetEntity == null` and return early in that case.

scripts/skills/perks/perk_rf_inspire.nut
```diff
@@ -0,0 +1,69 @@
...
+
+		local actor = this.getContainer().getActor();
+		if (actor.getFaction() == _targetEntity.getFaction())
+		{
+			local dist = actor.getTile().getDistanceTo(_targetTile);
```
LordMidas, line 55: The local `dist` is unnecessary here. It is more efficient to just do the calculation inside the `if` condition.

mod_reforged/hooks/config/strings.nut
LordMidas, line 1304: Because of nested tooltips, we don't need to give a full description of what the active skill does here. Just giving a general description of it is fine and the specifics can be found in the nested tooltip. I've been trying to do that in general with perk tooltips that give actives. For example you could write here `Gain the [Inspire|Skill+rf_inspire_skill] that allows you to target allies to increase their morale.` This way when we change the behavior of the inspire active skill we don't have to remember to modify the numbers in the string here as well.

LordMidas (changes requested):

scripts/skills/perks/perk_rf_inspire.nut
Darxo: Ok I remove it. 
In the future we might want to add some marker to show whether I player already cashed in the triggered action point recovery this turn. But not now as I think its very unlikely that they forget that within a turn.

scripts/skills/actives/rf_inspire_skill.nut
Darxo: I removed that line because I just realized that it is redundant.
An Icon is already generated on the receiver of the morale check because I pass a sprite name in that faction.

mod_reforged/hooks/config/strings.nut
```diff
@@ -1325,6 +1326,22 @@ foreach (vanillaDesc in vanillaDescriptions)
...
+			{
+				Type = ::UPD.EffectType.Passive,
+				Description = [
+					"Once per turn, after you use a skill targeting an adjacent ally, recover " + ::MSU.Text.colorGreen("2") + " [Action Points|Concept.ActionPoints].",
+				],
+			},
+			{
+				Type = ::UPD.EffectType.Active,
+				Description = [
+					"Gain the [Inspire|Skill+rf_inspire_skill] skill that allows you to increase the morale of an ally.",
```
LordMidas, line 1340: Please add hyperlinks for `turn` and `morale`

LordMidas (changes requested):
A few of my earlier comments are still unresolved (probably because they were hidden in the GitHub UI), please take a look at them too.

scripts/skills/actives/rf_inspire_skill.nut
```diff
@@ -0,0 +1,99 @@
...
+
+		// Private
+		InspireFlag = "RF_InspireSkillFlag",
+	},
+
+	function create()
+	{
+		this.m.ID = "actives.rf_inspire";
+		this.m.Name = "Inspire";
+		this.m.Description = "Inspire an ally to raise their current Morale. A character can only be inspired once per turn. Cannot be used on fleeing or stunned allies.";
```
LordMidas, line 14: - `Morale` and `fleeing` should be hyperlinked with `Concept.Morale`.
- `turn` should be hyperlinked with `Concept.Turn`.
- `stunned` should be hyperlinked with stunned effect.

scripts/skills/actives/rf_inspire_skill.nut
```diff
@@ -0,0 +1,99 @@
...
+		return ::Math.max(0, inspireBonus);	// The bonus can never be negative
+	}
+
+	function __canBeInspired( _targetEntity )
+	{
+		if (target.getCurrentProperties().IsStunned) return false;
+		if (target.getMoraleState() >= ::Const.MoraleState.Confident) return false;
+		if (target.getMoraleState() >= target.m.MaxMoraleState) return false;	// This might be confusing for the player because it's not mentioned in the tooltip but it's just logical
+
+		return (this.__getLastInspiredRound(_targetEntity) < ::Time.getRound());
```
LordMidas, line 87: Please remove the redundant parantheses

scripts/skills/actives/rf_inspire_skill.nut
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	function __getLastInspiredRound( _targetEntity )
+	{
+		if (::Tactical.Entities.getFlags().has(_targetEntity.getID() + this.m.InspireFlag))
+		{
+			return ::Tactical.Entities.getFlags().get(_targetEntity.getID() + this.m.InspireFlag);
+		}
+
+		return 0;
```
LordMidas, line 97: The design is to be only inspired once per turn, not per round. And we should keep it like that i.e. character dependent skills should generally always be turn-based and not round-based because of the concept of turns and rounds in this game. E.g. I inspire an ally, then that ally starts his turn, does some actions, then ends his turn. This should now qualify that ally again for an inspire.

While I really like the idea of using flags instead of dummy skill to block the usage of inspire on the same target again I don't know if it can achieve the turn-based design, as we need to know when the target has ended their turn. So, we might need to resort to a dummy skill unfortunately.

LordMidas (changes requested):

scripts/skills/actives/rf_inspire_skill.nut
Darxo: I prefer a once-per-round design. That is cleaner to follow by the player and probably also to implement. Since there are slight differences between those two concepts I realize I then must update the tooltips to say round

mod_reforged/config/perk_defs.nut
```diff
@@ -924,6 +924,14 @@
 		Icon = "ui/perks/rf_alert.png",
 		IconDisabled = "ui/perks/rf_alert_bw.png",
 	},
+	{
+		ID = "perk.rf_supporter",
+		Script = "scripts/skills/perks/perk_rf_supporter",
+		Name = ::Const.Strings.PerkName.RF_Supporter,
+		Tooltip = ::Const.Strings.PerkDescription.RF_Supporter,
+		Icon = "ui/perks/rf_supporter.png",
+		IconDisabled = "ui/perks/rf_supporter_sw.png",
```
LordMidas, line 933: Remove comma

mod_reforged/hooks/config/strings.nut
```diff
@@ -1422,6 +1423,22 @@ foreach (vanillaDesc in vanillaDescriptions)
 			]
 		}]
  	}),
+	RF_Supporter = ::UPD.getDescription({
```
LordMidas, line 1426: We need fluff for this. Something like `I\'ve got your back!`

scripts/skills/actives/rf_encourage_skill.nut
```diff
@@ -0,0 +1,104 @@
...
+	m = {
+		// Config
+		EncourageBonusFraction = 0.1		// This percentage of Resolve is added as a bonus for the morale check
+	},
+
+	function create()
+	{
+		this.m.ID = "actives.rf_encourage";
+		this.m.Name = "Encourage";
+		this.m.Description = ::Reforged.Mod.Tooltips.parseString("Encourage an ally to raise their current [Morale|Concept.Morale]. A character can only be encouraged once per round. Cannot be used on fleeing or [stunned|Skill+stunned_effect] allies.");
```
LordMidas, line 11: - `[fleeing|Concept.Morale]`
- `[round|Concept.Round]`

scripts/skills/actives/rf_encourage_skill.nut
```diff
@@ -0,0 +1,104 @@
...
+				type = "text",
+				icon = "ui/icons/bravery.png",
+				text = ::Reforged.Mod.Tooltips.parseString("Trigger a positive [Morale Check|Concept.Morale] with a bonus of " + ::MSU.Text.colorizeFraction(this.m.EncourageBonusFraction) + " (" + ::MSU.Text.colorizeValue(this.getEncourageBonus()) + ") of your [Resolve|Concept.Bravery] on the target")
+			},
+			{
+				id = 15,
+				type = "text",
+				icon = "ui/icons/vision.png",
+				text = "Has a range of " + ::MSU.Text.colorGreen(this.getMaxRange()) + " tiles"
+			},
```
LordMidas, line 47: Remove comma

scripts/skills/actives/rf_encourage_skill.nut
```diff
@@ -0,0 +1,104 @@
...
+		]);
+
+		return ret;
+	}
+
+	function onVerifyTarget( _originTile, _targetTile )
+	{
+		if (!this.skill.onVerifyTarget(_originTile, _targetTile)) return false;
```
LordMidas, line 55: Newline the return

scripts/skills/actives/rf_encourage_skill.nut
```diff
@@ -0,0 +1,104 @@
...
+
+		local target = _targetTile.getEntity();
+		if (this.getContainer().getActor().getFaction() != target.getFaction()) return false;
+		return this.__canBeEncouraged(target);
```
LordMidas, line 59: `return this.getContainer().getActor().getFaction() == target.getFaction() && this.__canbeEncouraged(target);`

scripts/skills/actives/rf_encourage_skill.nut
```diff
@@ -0,0 +1,104 @@
...
+	{
+		local encourageBonus = this.getContainer().getActor().getCurrentProperties().getBravery() * this.m.EncourageBonusFraction;
+		return ::Math.max(0, encourageBonus);	// The bonus can never be negative
+	}
+
+	function __canBeEncouraged( _targetEntity )
+	{
+		if (target.getCurrentProperties().IsStunned) return false;
+		if (target.getMoraleState() >= ::Const.MoraleState.Confident) return false;
+		if (target.getMoraleState() >= target.m.MaxMoraleState) return false;	// This might be confusing for the player because it's not mentioned in the tooltip but it's just logical
```
LordMidas: Please combine into a single `if`.

scripts/skills/actives/rf_encourage_skill.nut
```diff
@@ -0,0 +1,104 @@
...
+// New Functions
+	function getEncourageBonus()
+	{
+		local encourageBonus = this.getContainer().getActor().getCurrentProperties().getBravery() * this.m.EncourageBonusFraction;
+		return ::Math.max(0, encourageBonus);	// The bonus can never be negative
+	}
+
+	function __canBeEncouraged( _targetEntity )
+	{
+		if (target.getCurrentProperties().IsStunned) return false;
```
LordMidas: All 3 of these conditions should be with the variable `_targetEntity`.

LordMidas (changes requested):

scripts/skills/actives/rf_inspire_skill.nut
LordMidas: I see the advantage in simplicity of the round-based design. But I'm averse to the idea of round-based skills because that isn't in line with the concept of what a turn represents and what a round represents. The rest of the skills in this game all care about turns. So I quite strongly feel that this should be a turn-based effect. 

> Not to mention that you can otherwise inspire the same guy twice in the same round if you play with the turn order correctly.

That is fine. With the round-based design you can do something similar i.e. if the other guy acted after you in the current round and then in the next round you inspire him again before he acts. It's just shifted in time.

LordMidas: For the actual 0.5.0 build I'd want the Encourage ability to be turn-based and not round-based. But in favor of speed I'm willing to let the round-based design be merged in for now so that we can get an unstable build out tonight for testers.

mod_reforged/config/perk_defs.nut
Darxo: Every other perk def introduced by reforged also ends with a comma here

LordMidas: Yeah that's a mistake that we are going to fix in a separate overall commit later, but for the new perkdefs we shouldn't add the comma.

LordMidas (approved):
I pushed a couple of commits for minor refactor and style changes.

PR #292, opened 2024-03-18
Feat add three new injuries (black eye, heat stroke, dislocated jaw)

Darxo:
### Black Eye
image

### Heat Stroke
image

### Dislocated Jaw
**Not his turn**
image

**His Turn**
image


scripts/skills/injury/rf_black_eye_injury.nut
```diff
@@ -0,0 +1,75 @@
...
+		}
+
+		if (this.m.RangedSkillMult != 1.0)
+		{
+			tooltip.push({
+				id = 9,
+				type = "text",
+				icon = "ui/icons/ranged_skill.png",
+				text = ::MSU.Text.colorizeMult(this.m.RangedSkillMult, {AddSign = true}) + " Ranged Skill"
+			});
```
LordMidas, line 52: Please add hyperlinks in the tooltips for concepts such as vision, melee skill, ranged skill. Please go through the other files too and do the same where applicable.

scripts/skills/injury/rf_black_eye_injury.nut
```diff
@@ -0,0 +1,75 @@
...
+		}
+
+		if (this.m.RangedSkillMult != 1.0)
+		{
+			tooltip.push({
+				id = 9,
+				type = "text",
+				icon = "ui/icons/ranged_skill.png",
+				text = ::MSU.Text.colorizeMult(this.m.RangedSkillMult, {AddSign = true}) + " Ranged Skill"
```
LordMidas, line 51: Multiplicative changes should be without sign and use the words `increased/decreased` for now. Later we can change them to `more/less` in a single overall commit when we get to that point. Please go through the other files too and do the same where applicable.

scripts/skills/injury/rf_dislocated_jaw_injury.nut
```diff
@@ -0,0 +1,81 @@
...
+	m = {
+		FatigueRecoveryModifier = -3,
+		ResolveMult = 0.7,
+	},
+	function create()
+	{
+		this.injury.create();
+		this.m.ID = "injury.rf_dislocated_jaw";
+		this.m.Name = "Dislocated Jaw";
+		this.m.Description = "The forceful impact to the jaw caused it to become dislocated, making it painful to breath or communicate.";
```
LordMidas, line 11: `A forceful impact` instead of The

scripts/skills/injury/rf_heat_stroke_injury.nut
```diff
@@ -0,0 +1,51 @@
+this.rf_heat_stroke_injury <- this.inherit("scripts/skills/injury/injury", {
+	m = {
+		FatigueCostMult = 1.25,
+	},
+	function create()
+	{
+		this.injury.create();
+		this.m.ID = "injury.rf_heat_stroke";
+		this.m.Name = "Heat Stroke";
+		this.m.Description = "An intense heat caused the player\'s body temperature to soar to dangerous levels.";
```
LordMidas, line 10: `Being exposed to intense heat caused this character\'s body temperature to soar to dangerous levels.`

scripts/skills/injury/rf_heat_stroke_injury.nut
```diff
@@ -0,0 +1,51 @@
+this.rf_heat_stroke_injury <- this.inherit("scripts/skills/injury/injury", {
+	m = {
+		FatigueCostMult = 1.25,
```
LordMidas, line 3: Should be `FatigueEffectMult` because it applies to all fatigue built, not just from skills with fatigue cost.

scripts/skills/injury/rf_heat_stroke_injury.nut
```diff
@@ -0,0 +1,51 @@
...
+		this.injury.onUpdate(_properties);
+
+		if (!_properties.IsAffectedByInjuries || this.m.IsFresh && !_properties.IsAffectedByFreshInjuries)
+		{
+			return;
+		}
+
+		_properties.FatigueEffectMult *= this.m.FatigueCostMult;
+	}
+
```
LordMidas, line 49: This empty line before the last closing brackets should be removed. It's also present in the other files so please remove from them too.

scripts/skills/injury/rf_dislocated_jaw_injury.nut
```diff
@@ -0,0 +1,81 @@
+this.rf_dislocated_jaw_injury <- this.inherit("scripts/skills/injury/injury", {
+	m = {
+		FatigueRecoveryModifier = -3,
+		ResolveMult = 0.7,
```
LordMidas, line 4: Should we instead call this `BraveryMult`? At least that way in the code it is always called the same thing (Bravery instead of Resolve).

scripts/skills/injury/rf_black_eye_injury.nut
```diff
@@ -0,0 +1,75 @@
...
+		this.m.DropIcon = "rf_black_eye_injury";
+		this.m.Icon = "ui/injury/rf_black_eye_injury.png";
+		this.m.IconMini = "rf_black_eye_injury_mini";
+		this.m.HealingTimeMin = 1;
+		this.m.HealingTimeMax = 3;
+	}
+
+	function getTooltip()
+	{
+		local tooltip = this.skill.getTooltip();
```
LordMidas, line 23: I would suggest changing this variable's name to `local ret =`. I've been doing that in my new files too. We do that with virtually all return values. What do you think?

scripts/skills/injury/rf_black_eye_injury.nut
```diff
@@ -0,0 +1,75 @@
...
+
+		if (this.m.VisionModifier != 0)
+		{
+			tooltip.push({
+				id = 7,
```
LordMidas, line 28: I've been starting our ids at 10 (as it gives more room for other mods who may be adding between 7-9). What do you think?

LordMidas (changes requested):

scripts/skills/injury/rf_black_eye_injury.nut
Darxo: I started that using `tooltip` because I saw someone else (you?) do it.
I like `ret`

scripts/skills/injury/rf_dislocated_jaw_injury.nut
Darxo: I don't like that for Bravery/Resolve in particular and would leave it up to the respective dev.

By that logic we would need to start using `Stamina` for some fatigue related variable names. Though on a side note I actually want that example because I still want to partly return to `Stamina`

scripts/skills/injury/rf_heat_stroke_injury.nut
Darxo: FatigueCostMult does not mention `skills`. I find it a very neutral term and always see it as including cost for moving around in it.
Using skills and moving around costs fatigue.

Other sources of fatigue build-up like getting hit, encumbrance or special triggered costs from skills are not influenced (en garde, rebuke)

scripts/skills/injury/rf_dislocated_jaw_injury.nut
LordMidas: In my opinion, when a multiplier applies to a certain property/field in an object it should conform to the naming scheme of THAT object. So, for example. if you want to modify the `_properties.Stamina` then you should use `StaminaMult` and if you want to modify the `actor.setFatigue` then you should use `FatigueModifier` etc.

This is the most stable and uniform way of working with these fields in my opinion and is the path of least resistance. Otherwise, everyone uses their own preferred style and it becomes a mess to browse through the codebase.

scripts/skills/injury/rf_heat_stroke_injury.nut
LordMidas: Same [comment as on the other conversation](https://github.com/Battle-Modders/mod-reforged/pull/292#discussion_r1566774650). If you are changing the `FatigueEffectMult` then this variable should be named similarly to make that clear.

scripts/skills/injury/rf_dislocated_jaw_injury.nut
```diff
@@ -0,0 +1,79 @@
...
+					text = ::Reforged.Mod.Tooltips.parseString(::MSU.Text.colorizeMult(this.m.ResolveMult) + " less [Resolve|Concept.Bravery]")
+				});
+			}
+			else
+			{
+				ret.push({
+					id = 11,
+					type = "text",
+					icon = "ui/icons/special.png",
+					text = ::Reforged.Mod.Tooltips.parseString(::MSU.Text.colorizeMult(this.m.ResolveMult) + " less [Resolve|Concept.Bravery] during this characters turn")
```
LordMidas, line 52: `character\'s`

scripts/skills/injury/rf_heat_stroke_injury.nut
```diff
@@ -0,0 +1,49 @@
...
+	{
+		local ret = this.skill.getTooltip();
+
+		if (this.m.FatigueCostMult != 1.0)
+		{
+			ret.push({
+				id = 10,
+				type = "text",
+				icon = "ui/icons/fatigue.png",
+				text = ::Reforged.Mod.Tooltips.parseString(::MSU.Text.colorizeMult(this.m.FatigueCostMult) + " more [Fatigue|Concept.Fatigue] built")
```
LordMidas, line 29: Need to invert color here to make it red.

scripts/skills/injury/rf_black_eye_injury.nut
```diff
@@ -0,0 +1,73 @@
...
+	{
+		local ret = this.skill.getTooltip();
+
+		if (this.m.VisionModifier != 0)
+		{
+			ret.push({
+				id = 10,
+				type = "text",
+				icon = "ui/icons/vision.png",
+				text = ::MSU.Text.colorRed(this.m.VisionModifier) + " Vision"
```
LordMidas, line 31: We can take this opportunity to add `Vision` as a concept. To do this you can add a commit to the `z_nested_tooltips.nut` file and in the `AutoConcepts` array you can add `"character-stats.SightDistance"`. Then here it can be parsed using `[Vision|Concept.SightDistance]`

mod_reforged/hooks/config/character_injuries.nut
LordMidas, line 1: Need to add these injuries to the `msu_excluded_injuries.nut` file as well. I think all 3 of these injuries should be excluded from all kinds of Undead so you can simply add them to the `RF_Undead` excluded injuries list.

LordMidas (changes requested):

scripts/skills/injury/rf_dislocated_jaw_injury.nut
Darxo: I'm ok with this way of writing those typical member variables now

mod_reforged/config/msu_excluded_injuries.nut
```diff
@@ -34,6 +35,7 @@
 		"injury.pierced_side",
 		"injury.pierced_arm_muscles",
 		"injury.stabbed_guts",
+		"injury.rf_black_eye"
```
LordMidas, line 38: Black Eye should be removed from RF_Undead instead of only from RF_Skeleton. Undead do not have a functioning blood circulatory system and therefore shouldn't be getting bruises or swollen body parts from getting hit.

Dislocated Jaw should also be removed from skeletons imo but we can discuss this in design discussion if you like.

scripts/skills/injury/rf_dislocated_jaw_injury.nut
```diff
@@ -0,0 +1,79 @@
...
+	m = {
+		FatigueRecoveryModifier = -3,
+		BraveryMult = 0.7,
+	},
+	function create()
+	{
+		this.injury.create();
+		this.m.ID = "injury.rf_dislocated_jaw";
+		this.m.Name = "Dislocated Jaw";
+		this.m.Description = "A forceful impact to the jaw caused it to become dislocated, making it painful to breath or communicate.";
```
LordMidas, line 11: just noticed a typo: `breathe`

LordMidas (changes requested):

mod_reforged/hooks/config/character_injuries.nut
```diff
@@ -0,0 +1,74 @@
+// Regular Damage Types
+::Const.Injury.BluntHead.extend([
```
LordMidas, line 2: I'm sorry for missing this in my earlier reviews but I just noticed it now that these injuries should also additionally be added to the following vanilla arrays:
1. All three injuries go in `::Const.Injury.All`.
2. Black Eye and Dislocated Jaw go in `::Const.Injury.BluntAndPiercingHead`
3. Heat Stroke goes in `::Const.Injury.BurningAndPiercingBody` and `::Const.Injury.Burning`.

LordMidas (changes requested):

mod_reforged/hooks/config/character_injuries.nut
```diff
@@ -1,74 +1,126 @@
...
+		{
+			ID = "injury.rf_dislocated_jaw",
+			Threshold = 0.25,
+			Script = "injury/rf_dislocated_jaw_injury"
+		},
+		{
+			ID = "injury.rf_heat_stroke",
+			Threshold = 0.25,
+			Script = "injury/rf_heat_stroke_injury"
+		},
```
LordMidas, line 40: Comma should be removed

mod_reforged/hooks/config/character_injuries.nut
```diff
@@ -1,74 +1,126 @@
...
+		{
+			ID = "injury.rf_black_eye",
+			Threshold = 0.25,
+			Script = "injury/rf_black_eye_injury"
+		},
+		{
+			ID = "injury.rf_dislocated_jaw",
+			Threshold = 0.25,
+			Script = "injury/rf_dislocated_jaw_injury"
+		},
```
LordMidas, line 53: Comma should be removed

mod_reforged/hooks/config/character_injuries.nut
```diff
@@ -1,74 +1,126 @@
...
+			Script = "injury/rf_heat_stroke_injury"
+		},
+	]);
+
+	::Const.Injury.BurningAndPiercingBody.extend([
+		{
+			ID = "injury.rf_heat_stroke",
+			Threshold = 0.25,
+			Script = "injury/rf_heat_stroke_injury"
+		},
```
LordMidas, line 69: Comma should be removed

mod_reforged/hooks/config/character_injuries.nut
```diff
@@ -1,74 +1,126 @@
...
-		Threshold = 0.25,
-		Script = "injury/rf_dislocated_jaw_injury"
-	}
-]);
+	::Const.Injury.Burning.extend([
+		{
+			ID = "injury.rf_heat_stroke",
+			Threshold = 0.25,
+			Script = "injury/rf_heat_stroke_injury"
+		},
```
LordMidas, line 61: Comma should be removed

LordMidas (approved):
Looks good to merge after the requested comma removals.

mod_reforged/hooks/config/character_injuries.nut
LordMidas: I am aware that commas are optional in squirrel (it will work even without commas in the entire array/table and with extra comma at the end). But your point that having commas at the end can lead to cleaner diff in commits is interesting. I don't know what the convention is in this case that we want to go with, so I'd ask Enduriel and if he's ok with this then we can do it.

LordMidas: Based on the discussion in Discord we decided that we like these commas at the end. So you can disregard this comment.

PR #294, opened 2024-03-20
refactor: improve vanilla iron will tooltip to clarify full immunity

Darxo:
New Descriptions:

image

image


LordMidas (changes requested):
Also need to hook the `iron_will_effect.nut` and make the same change to its tooltip.

LordMidas (approved):

PR #295, opened 2024-03-20
Refactor recover action points utility function

Darxo:


mod_reforged/hooks/entity/tactical/actor.nut
LordMidas, line 1: If we are going to add these functions then they should:
- Handle increasing/decreasing the stat.
- Handle all stats.

Therefore, I suggest something like the following:
```squirrel
// Don't like this function name, should think of something better
q.changeStat <- function( _statName, _change, _printLog = true, _canExceedBounds = false )
{
	local setter = this["set" + _statName];
	local getter = this["get" + _statName];
	local getterMax = this["get" + _statName + "Max"];

	local oldValue = getter();
	if (_canExceedBounds)
		setter(oldValue + _amount);
	else
		setter(::Math.max(0, ::Math.min(getterMax(), oldValue + _amount)));
	
	local netChange = getter() - oldValue;
	if (_printLog && netChange != 0)
	{
		local str = netChange > 0 ? " recovers " + ::MSU.Text.colorGreen(netChange) : " loses " + ::MSU.Text.colorRed(netChange);
		::Tactical.EventLog.log(::Const.UI.getColorizedEntityName(this) + str + " " + _statName);
	}
}

q.changeActionPoints <- function( _change, _printLog = true, _canExceedBounds = false )
{
	this.changeStat("ActionPoints", _printLog, _canExceedBounds);
}

q.changeHitpoints <- function( _change, _printLog = true, _canExceedBounds = false )
{
	this.changeStat("Hitpoints", _printLog, _canExceedBounds);
}
```

LordMidas (changes requested):

mod_reforged/hooks/entity/tactical/actor.nut
Darxo: changeStat is too complicated for me. If no one else is gonna implement that generic function and consider all of its edge cases then I prefer implementing the function `recoverActionPoints` as described above and later on we can redirect it to use that generic function

PR #297, opened 2024-03-23
Refactor cull between the eyes

LordMidas:
- Remove Between the Eyes perk and grant the Between the Eyes skill via the Cull perk.
- Reduce the additional Fatigue Cost from Between the Eyes skill from 20 to 10.


Darxo (changes requested):
I don't see where you adjust the perk description for cull

Darxo (changes requested):
You have not removed the 10% damage bonus part of Cull (neither from description nor actual effect).
That effect is no longer present on this skill according to memory and the perk sheet.

Darxo (approved):

PR #299, opened 2024-03-23
Refactor tweak strength in numbers

LordMidas:
- Gain resolve for adjacent allies instead of allies in combat
- Work with allied actors instead of faction only actors


Darxo (commented):
Since you're already touching this file you could aswell replace the vanilla style coloring with `::MSU.Text.colorizeValue`

Darxo (approved):

PR #300, opened 2024-03-23
refactor: rework unstoppable to increase AP and Initiative

LordMidas:
As long as you continue to end your turn with half or less AP remaining.

### DIFF
--- a/scripts/skills/perks/perk_rf_unstoppable.nut
+++ b/scripts/skills/perks/perk_rf_unstoppable.nut
@@ -1,16 +1,14 @@
 this.perk_rf_unstoppable <- ::inherit("scripts/skills/skill", {
 	m = {
 		Stacks = 0,
-		SkillBonusPerStack = 5,
-		MaxStacks = 10,
-		Distance = 0,
-		APBonusBefore = 0
+		MaxStacks = 5,
+		IsGainingStackThisTurn = true // We set it to false after using Recover so we don't gain stack during that turn
 	},
 	function create()
 	{
 		this.m.ID = "perk.rf_unstoppable";
 		this.m.Name = ::Const.Strings.PerkName.RF_Unstoppable;
-		this.m.Description = "This character\'s attacks seem to not miss at all.";
+		this.m.Description = "This character is like a boulder rolling down a hill. Unstoppable!";
 		this.m.Icon = "ui/perks/rf_unstoppable.png";
 		this.m.IconMini = "rf_unstoppable_mini";
 		this.m.Type = ::Const.SkillType.Perk | ::Const.SkillType.StatusEffect;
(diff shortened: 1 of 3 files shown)


scripts/skills/perks/perk_rf_unstoppable.nut
```diff
@@ -30,85 +27,70 @@ this.perk_rf_unstoppable <- ::inherit("scripts/skills/skill", {
...
+	function getInitiativeBonus()
+	{
+		return this.m.Stacks * 10;
+	}
+
 	function getTooltip()
 	{
 		local tooltip = this.skill.getTooltip();
 
 		tooltip.push({
```
Darxo, line 44: This should be wrapped around by an `if (this.getAPBonus())` in case a submod changes how these values are generated.

Darxo (changes requested):

scripts/skills/perks/perk_rf_unstoppable.nut
```diff
@@ -30,95 +27,101 @@ this.perk_rf_unstoppable <- ::inherit("scripts/skills/skill", {
...
-	{
-		local actor = this.getContainer().getActor();
-		if (_skill.isAttack() && ::Tactical.TurnSequenceBar.isActiveEntity(actor) &&!_targetEntity.isAlliedWith(actor))
-		{
-			this.m.Stacks = ::Math.minf(this.m.MaxStacks, this.m.Stacks + (this.m.Distance > 1 ? 0.5 : 1));
-			actor.setActionPoints(actor.getActionPoints() + this.getAPBonus() - this.m.APBonusBefore);
-		}
-	}
+		tooltip.push({
+			id = 11,
```
Darxo, line 65: Now the id 11 is used twice.
This should be a unique id, e.g. 12 (if 12 is indeed unused by the time you implement this change)

Darxo (approved):
After you fixed the remaining id duplication (11) it's ok to be merged

Uberbagel:
Testing:
Unstoppable is not losing all stacks when using wait. I tried it twice. Recover however is removing all stacks as intended. Log is clean.

Also, as it works right now, if I use recover, I then get a stack of Unstoppable for that turn, because it ended with half actions points remaining or fewer. I do not think you should get a stack of Unstoppable on the turn you use Recover.

Uberbagel (changes requested):
Testing:
Unstoppable is not losing all stacks when using wait. I tried it twice. Recover however is removing all stacks as intended. Log is clean.

Also, as it works right now, if I use recover, I then get a stack of Unstoppable for that turn, because it ended with half actions points remaining or fewer. I do not think you should get a stack of Unstoppable on the turn you use Recover.

Darxo (approved):

scripts/skills/perks/perk_rf_unstoppable.nut
```diff
@@ -80,14 +80,8 @@ this.perk_rf_unstoppable <- ::inherit("scripts/skills/skill", {
...
 
-		if (this.m.Stacks >= this.m.MaxStacks)
-			return;
-
 		local actor = this.getContainer().getActor();
-		if (actor.getActionPoints() <= actor.getActionPointsMax() / 2)
-		{
-			this.m.Stacks++;
-		}
+		this.m.Stacks = actor.getActionPoints() <= actor.getActionPointsMax() / 2 ? ::Math.min(this.m.MaxStacks, this.m.Stacks + 1) : 0;
```
Darxo, line 84: I would put `(actor.getActionPoints() <= actor.getActionPointsMax() / 2)` like this in brackets. It's a mouthful of a condition that makes this one-liner harder to parse.
This comment is mainly to archive this line of code as an example in case we discuss the use of optional brackets.

scripts/skills/perks/perk_rf_unstoppable.nut
```diff
@@ -131,7 +125,7 @@ this.perk_rf_unstoppable <- ::inherit("scripts/skills/skill", {
 				id = 15,
 				type = "text",
 				icon = "ui/icons/warning.png",
-				text = ::Reforged.Mod.Tooltips.parseString(::MSU.Text.colorRed("Will cause you to lose all stacks of [Unstoppable|Skill+perk_rf_unstoppable]"))
+				text = ::Reforged.Mod.Tooltips.parseString(::MSU.Text.colorRed(format("Will cause you to lose all stacks of [%s|Perk+%s]", ::Const.Perks.findById(this.getID()).Name, split(::IO.scriptFilenameByHash(this.ClassNameHash), "/").top())))
```
Darxo, line 128: I've seen this way of writing a few times now and it strikes me as lazy and hard to read.

I recommend that you create a global utility function like this:
```
function getPerkHyperlink( _class )
{
   return format([%s|Perk+%s]", ::Const.Perks.findById(_class.getID()).Name, split(::IO.scriptFilenameByHash(_class.ClassNameHash), "/").top())
}

LordMidas: I agree it would be good to have a function like that. I'll set it up.

scripts/skills/perks/perk_rf_unstoppable.nut
LordMidas: I've changed it to an `if/else` to improve readability in this case.

Darxo (approved):

PR #301, opened 2024-03-23
refactor: rework professional gain perks from a random melee perk group

LordMidas:
Grant the first two perks in a random melee perk group that this character has access to.

### DIFF
--- a/mod_reforged/hooks/config/strings.nut
+++ b/mod_reforged/hooks/config/strings.nut
@@ -1554,11 +1554,11 @@ foreach (vanillaDesc in vanillaDescriptions)
 		}]
  	}),
 	RF_Professional = ::UPD.getDescription({
- 		Fluff = "You\'re a professional, experienced fighter, able to wield many weapons in many styles!",
+ 		Fluff = "You\'re a professional, experienced fighter!",
  		Effects = [{
 			Type = ::UPD.EffectType.Passive,
 			Description = [
-				"Gain the [Shield Expert|Perk+perk_shield_expert], [Duelist|Perk+perk_duelist], [Weapon Master|Perk+perk_rf_weapon_master], [Death Dealer|Perk+perk_rf_death_dealer], and [Formidable Approach|Perk+perk_rf_formidable_approach] perks."
+				"Gain the first two [perks|Concept.Perk] in a random melee perk group that you have access to."
 			]
 		}]
  	}),
--- a/mod_reforged/hooks/config/strings.nut
+++ b/mod_reforged/hooks/config/strings.nut
@@ -1556,7 +1556,7 @@ foreach (vanillaDesc in vanillaDescriptions)
 	RF_Professional = ::UPD.getDescription({
  		Fluff = "You\'re a professional, experienced fighter!",
  		Effects = [{
-			Type = ::UPD.EffectType.OneTimeEffect,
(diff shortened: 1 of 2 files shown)


mod_reforged/hooks/config/strings.nut
```diff
@@ -1554,11 +1554,11 @@ foreach (vanillaDesc in vanillaDescriptions)
 		}]
  	}),
 	RF_Professional = ::UPD.getDescription({
- 		Fluff = "You\'re a professional, experienced fighter, able to wield many weapons in many styles!",
+ 		Fluff = "You\'re a professional, experienced fighter!",
  		Effects = [{
 			Type = ::UPD.EffectType.Passive,
```
Darxo, line 1559: This must now be `OneTimeEffect`

scripts/skills/perks/perk_rf_professional.nut
```diff
@@ -15,45 +15,85 @@ this.perk_rf_professional <- ::inherit("scripts/skills/skill", {
...
+		foreach (row in ::DynamicPerks.PerkGroups.findById(chosenPerkGroupID).getTree())
+		{
+			foreach (perkID in row)
+			{
+				perkIDs.push(perkID);
+			}
+		}
+
+		if (perkIDs.len() > 2)
+			perkIDs = perkIDs.slice(0, 2);
```
Darxo, line 92: You always remove the first two perks in that perk group. But those were not necessarily all added by `Professional`, if a player learned those perks before they learned `Professional`.

This is a minor edge case.

Maybe we prevent `Professional` from being picked if the player has learned any perk from any weapon tree.

LordMidas: The perks are added and removed via stack (`mod_stack_based_skills`) so it should be fine. Even if the bro already had the perk before picking Professional, Professional will add a "stack" of that perk, and when removed will only remove that stack. Meanwhile, if the bro "unlearns" that perk, he will unlearn the original perk, but the stack based addition from Professional will remain.

An issue here can happen with regards to refundability of the perks. So the original perk was refundable, but the one added by Professional isn't - however, the way stack based skills mod works is that the data is kept from the serialized version of the skill. So after the original perk is unlearned, the one added by Professional will also be refundable. I think a meaningful solution to this should be that all non-serialized skills should be considered not-refundable by default. I can update Dynamic Perks Framework to account for this.

Darxo: Ok, I made simple issue over there for a later time.
No further adjustments are needed regarding potential edge cases.

Darxo (approved):

mod_reforged/hooks/config/strings.nut
LordMidas: If the perk is refundable, should it still be OneTimeEffect? Or does it make more sense to keep it "Passive" in that case? I think the latter is more sensible because it passively grants you the other 2 perks but you lose those perks when you lose this perk.

Uberbagel:
Tested. Works as intended. Tested oblivion potion, which removes all given perks and refunds Professional.

mod_reforged/hooks/config/strings.nut
Darxo: You decide.
Overall I prefer making this perk non-refundable in which case this question is easy. But in this case were it is redunable, I have no idea and don't care.

PR #302, opened 2024-03-23
Rework phalanx dont lose shieldwall

LordMidas:
Don't lose shieldwall with phalanx if adjacent ally has shieldwall

### DIFF
--- a/scripts/skills/perks/perk_rf_phalanx.nut
+++ b/scripts/skills/perks/perk_rf_phalanx.nut
@@ -26,7 +26,7 @@ this.perk_rf_phalanx <- ::inherit("scripts/skills/skill", {
 			id = 10,
 			type = "text",
 			icon = "ui/icons/rf_reach.png",
-			text = ::MSU.Text.colorizeValue(this.getCount()) + " Reach when attacking or defending in melee"
+			text = ::Reforged.Mod.Tooltips.parseString(::MSU.Text.colorizeValue(this.getCount()) + " [Reach|Concept.Reach] when attacking or defending in melee")
 		});
 
 		if (this.hasAdjacentShieldwall())
--- a/scripts/skills/perks/perk_rf_phalanx.nut
+++ b/scripts/skills/perks/perk_rf_phalanx.nut
@@ -7,7 +7,7 @@ this.perk_rf_phalanx <- ::inherit("scripts/skills/skill", {
 		this.m.Description = "This character is highly skilled in fighting in a shielded formation and gains bonuses when adjacent to allies with shields.";
 		this.m.Icon = "ui/perks/rf_phalanx.png";
 		this.m.Type = ::Const.SkillType.Perk | ::Const.SkillType.StatusEffect;
-		this.m.Order = ::Const.SkillOrder.Perk;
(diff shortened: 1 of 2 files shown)


scripts/skills/perks/perk_rf_phalanx.nut
```diff
@@ -25,8 +25,16 @@ this.perk_rf_phalanx <- ::inherit("scripts/skills/skill", {
 			id = 6,
 			type = "text",
 			icon = "ui/icons/rf_reach.png",
-			text = ::MSU.Text.colorizeValue(this.getCount()) + " Reach when attacking or defending in melee"
+			text = ::Reforged.Mod.Tooltips.parseString(::MSU.Text.colorizeValue(this.getCount()) + " [Reach|Concept.Reach] when attacking or defending in melee")
 		});
+
+		tooltip.push({
```
Darxo, line 31: I would capsule both tooltip lines in their own if to only show up while their effects are actually active.

```
if (this.getCount())
{
// Add Reach Tooltip
}

if (this.hasAdjacentShieldwall())
{
// Add Shieldwall expire tooltip
}
```

And the Shieldwall expire tooltip would then be changed into
`"[Shieldwall|Skill+shieldwall_effect] on this character will not expire"`

scripts/skills/perks/perk_rf_phalanx.nut
```diff
@@ -46,6 +54,30 @@ this.perk_rf_phalanx <- ::inherit("scripts/skills/skill", {
 		}
 	}
 
+	function onTurnStart()
+	{
+		local actor = this.getContainer().getActor();
+		local myTile = actor.getTile();
+		for (local i = 0; i < 6; i++)
```
Darxo, line 61: This is less about your PR:
I would really like to have a `function getAdjacentActors()` for the `actor.nut` class which returns an array of up to 6 actor references.

MSU currently only offers a `getAlliedActors` but that is not efficient when we know that we only need to check the adjacent tiles.

Darxo (changes requested):

scripts/skills/perks/perk_rf_phalanx.nut
LordMidas: I've improved the tooltip a bit based on your suggestion. I think it is important to mention the condition in the tooltip i.e. being next to an ally with shieldwall. A general statement like "shieldwall will not expire" may be misinterpreted to mean that it won't expire unconditionally.

scripts/skills/perks/perk_rf_phalanx.nut
LordMidas: You can propose this idea as an issue over on the MSU repo. We will discuss it. But I want to point out that currently this can be done using the MSU-added `::Tactical.Entities.getActorsByFunction` as:
```squirrel
local myTile = actor.getTile();
local actors = ::Tactical.Entities.getActorsByFunction(@(a) a.isPlacedOnMap() && a.getTile().getDistanceTo(myTile) == 1);
```
This will give all actors even if they are allied or not. If we only want allied actors then we can just add `a.isAlliedWith(actor)` condition in there.

PR #303, opened 2024-03-23
refactor: remove bloodbath and rework sanguinary

LordMidas:
- Sanguinary is now what Bloodbath used to be.
- Sanguinary now requires Cleaver.

Note: I did not test this in game as the changes are mostly copy pasting the Bloodbath working code into the Sanguinary perk with only minor tweaks to the code.


scripts/skills/perks/perk_rf_sanguinary.nut
```diff
@@ -17,78 +18,35 @@ this.perk_rf_sanguinary <- ::inherit("scripts/skills/skill", {
...
-			this.m.IsHitHandlingComplete = true;
-		}
-	}
-
-	// Handles cases where Bleeding is applied by the attack e.g. during cleave skill's onUse function
-	// Note: Is unable to handle bleeding applied by "scheduledEvent" attacks
-	function onAnySkillExecuted( _skill, _targetTile, _targetEntity, _forFree )
-	{
-		if (this.m.IsHitHandlingComplete || !this.isAttackValid(_skill, this.getContainer().getActor(), _targetEntity))
+		if (_skill == null || !_skill.isAttack() || (!this.m.RequiresWeapon && !_skill.m.IsWeaponSkill) || !this.isEnabled())
```
Darxo, line 38: I don't understand this line `(!this.m.RequiresWeapon && !_skill.m.IsWeaponSkill)`

If no weapon is required and the skill is not a weapon skill, then nothing happens?

I feel like the first `!` is wrong and should be removed.

scripts/skills/perks/perk_rf_sanguinary.nut
```diff
@@ -1,8 +1,9 @@
 this.perk_rf_sanguinary <- ::inherit("scripts/skills/skill", {
 	m = {
-		FatigueCostRefundPercentage = 25,
-		IsHitHandlingComplete = true,
-		WasBleeding = false
+		IsForceEnabled = false,
+		RequiresWeapon = true,
```
Darxo, line 4: Why exactly do we need `IsForceEnabled` aswell as `RequiresWeapon`? It seems like they do the same thing.
I would like comments explaining the purpose of both of them.

scripts/skills/perks/perk_rf_sanguinary.nut
Darxo, line 38: Even if you fix the issue with the `!` above.

There is still the issue that `IsForceEnabled` is not strong enough to force-enable this effect if the skill is not a weapon skill. In that case we abort (`return;`) before `IsForceEnabled` inside the `this.isEnabled()` can do its work.

What exactly is `IsForceEnabled` supposed to do (see remark at the top about comment in member declaration), if not circumventing the weapon restriction?

Darxo (changes requested):

scripts/skills/perks/perk_rf_sanguinary.nut
LordMidas: This was the older system and quite confusing tbh to work with too. I've improved it and pushed a commit now using `isSkillValid` instead.

Darxo (approved):

PR #305, opened 2024-03-23
Rework kingfisher to clean up code and grant +2 Reach with usable net

LordMidas:
- Use the throw net skill. So can benefit from perks which enhance throw net skill.
- Properly disable swapping the net, the previous implementation could have been circumvented by other perks that grant reduced AP offhand swapping.
- Lose the net when moving more than 1 tile away from the netted target.

### DIFF
--- a/scripts/skills/perks/perk_rf_kingfisher.nut
+++ b/scripts/skills/perks/perk_rf_kingfisher.nut
@@ -2,6 +2,7 @@ this.perk_rf_kingfisher <- ::inherit("scripts/skills/skill", {
 	m = {
 		IsForceEnabled = false,
 		Chance = 0,
+		NetEffect = null,
 		IsSpent = false
 	},
 	function create()
@@ -21,7 +22,7 @@ this.perk_rf_kingfisher <- ::inherit("scripts/skills/skill", {
 	{
 		if (this.m.IsForceEnabled) return true;
 
-		return this.getContainer().hasSkill("actives.throw_net");
+		return this.getContainer().hasSkill("actives.throw_net"); // Ignores hidden skills
 	}
 
 	function onBeforeTargetHit( _skill, _targetEntity, _hitInfo )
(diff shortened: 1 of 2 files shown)


scripts/skills/perks/perk_rf_kingfisher.nut
```diff
@@ -21,7 +22,7 @@ this.perk_rf_kingfisher <- ::inherit("scripts/skills/skill", {
 	{
 		if (this.m.IsForceEnabled) return true;
 
-		return this.getContainer().hasSkill("actives.throw_net");
+		return this.getContainer().hasSkill("actives.throw_net") != null; // Ignores hidden skills so while this IsSpent, it will be false
```
Darxo, line 25: This is not necessary and misleading.
The Vanilla `hasSkill` function always returns either `true` or `false`.

scripts/skills/perks/perk_rf_kingfisher.nut
```diff
@@ -35,82 +36,92 @@ this.perk_rf_kingfisher <- ::inherit("scripts/skills/skill", {
 
 	function onTargetHit( _skill, _targetEntity, _bodyPart, _damageInflictedHitpoints, _damageInflictedArmor )
 	{
-		if (!_targetEntity.isAlive())
+		if (!_targetEntity.isAlive() || ::Math.rand(1, 100) > this.m.Chance || !this.getContainer().getActor().isPlacedOnMap() || !_targetEntity.isPlacedOnMap())
```
Darxo, line 39: You call `this.getContainer().getActor()` several times in this function.
You should therefor use a `local actor` for it.

scripts/skills/perks/perk_rf_kingfisher.nut
```diff
@@ -35,82 +36,92 @@ this.perk_rf_kingfisher <- ::inherit("scripts/skills/skill", {
...
+			return;
+
+		local netItemScript = ::IO.scriptFilenameByHash(this.getContainer().getActor().getOffhandItem().ClassNameHash);
+
+		throwNetSkill.useForFree(_targetEntity.getTile());
+
+		local replacementNet = ::new(netItemScript)
+
+		this.getContainer().getActor().getItems().equip(replacementNet); // the original net is unequipped during use of throw net skill, so we equip a "new" net of the same type
+		foreach (skill in replacementNet.m.SkillPtrs)
```
Darxo, line 53: Before continuing here you need to double check whether the target is even netted.
In Vanilla it is possible to throw nets on enemies which are immune to the net effect.

scripts/skills/perks/perk_rf_kingfisher.nut
```diff
@@ -35,82 +36,92 @@ this.perk_rf_kingfisher <- ::inherit("scripts/skills/skill", {
...
+			// so we can't get it via container.getSkillByID, instead we find it by iterating over the skills of the item
+			if (skill.getID() == "actives.throw_net")
+			{
+				skill.m.IsHidden = true;
+				break;
+			}
+		}
+
+		local netEffect = _targetEntity.getSkills().getSkillByID("effects.net");
+		netEffect.m.KingfisherPerk <- ::MSU.asWeakTableRef(this);
```
Darxo, line 65: This will probably crash the game when you kingfishered an enemy who is immune to being netted.
You just blindly assume that the net effect is present on the enemy

scripts/skills/perks/perk_rf_kingfisher.nut
```diff
@@ -35,82 +36,92 @@ this.perk_rf_kingfisher <- ::inherit("scripts/skills/skill", {
...
-			local isReinforced = this.getContainer().getSkillByID("actives.throw_net").m.IsReinforced;
-			local effect = ::Tactical.spawnSpriteEffect(isReinforced ? "bust_net_02" : "bust_net", this.createColor("#ffffff"), _targetEntity.getTile(), 0, 10, 1.0, _targetEntity.getSprite("status_rooted").Scale, 100, 100, 0);
-			local flip = !_targetEntity.isAlliedWithPlayer();
-			effect.setHorizontalFlipping(flip);
-			::Time.scheduleEvent(::TimeUnit.Real, 200, this.onNetSpawn.bindenv(this), {
-				TargetEntity = _targetEntity,
-				IsReinforced = isReinforced
-			});
+			onDeath();
+			this.onRemoved();
```
Darxo, line 79: This line does not seem correct. So when the target dies then our kingfisher perk is removed?

You probably want
```
if (!::MSU.isNull(this.m.KingfisherPerk))
{
	this.m.KingfisherPerk.setSpent(false);
}
```
here

scripts/skills/perks/perk_rf_kingfisher.nut
```diff
@@ -35,82 +36,92 @@ this.perk_rf_kingfisher <- ::inherit("scripts/skills/skill", {
...
 		{
-			if (item != null && item.getSlotType() == ::Const.ItemSlot.Offhand)
-				return 99;
+			this.m.NetEffect.m.KingfisherPerk = null;
+			this.setSpent(false);
+			this.getContainer().getActor().getItems().unequip(this.getContainer().getActor().getOffhandItem());
 		}
 	}
 
 	function setSpent( _isSpent )
```
Darxo, line 106: As this is a newly introduced function it should be sorted to the bottom of this script.
I would then also introduce those functions with `// New Functions`

Like so:
```

// New Functions
	function setSpent( _isSpent )
	{
		[...]
	}
```

Darxo (changes requested):

scripts/skills/perks/perk_rf_kingfisher.nut
LordMidas: Ah yes the `!= null` is there by mistake. Will fix.

scripts/skills/perks/perk_rf_kingfisher.nut
LordMidas: I see. I was not aware of this. I will add this check. Nevertheless, should we also make the net skill unusable on such targets?

scripts/skills/perks/perk_rf_kingfisher.nut
LordMidas: `this` here is the `netEffect`. On death of the netted character it calls the `this.onRemoved()` function inside the netEffect which is hooked just a few lines above. That then calls the `setSpent(false)` for the kingfisrher perk.

scripts/skills/perks/perk_rf_kingfisher.nut
LordMidas: I am not convinced for the need for this, especially in a completely new skill script file. If it was a hooked file then I could see _some_ use for that even there I am not fully convinced but it's relevant. For brand new scripts this seems unnecessary. Vanilla doesn't do it either for special functions inside certain skills.

scripts/skills/perks/perk_rf_kingfisher.nut
Darxo: Yes, I suggest we add a `if (!targetEntity.getCurrentProperties().IsImmuneToRoot)` to the vanilla `throw_net` skill `onVerifyTarget` function.

That should fix this issue for the foreseeable future.

scripts/skills/perks/perk_rf_kingfisher.nut
Darxo: I wonder now if this causes potential future issues. `onRemoved` is not normally called during the `onDeath` of a skill.
What if a submod adds an additional effect to when a net_effect is removed which expects the character to still be alive

scripts/skills/perks/perk_rf_kingfisher.nut
Darxo: And I think was bad practice by Vanilla. 
Those new utility functions mixed in between the overwritten functions makes it harder to understand what a skill is actually doing.

scripts/skills/perks/perk_rf_kingfisher.nut
LordMidas: I've edited the code to add a separate function to the skill which is now called from both `onDeath` and `onRemoved`.

scripts/skills/perks/perk_rf_kingfisher.nut
LordMidas: As far as I could understand based on the discussion in Discord, we didn't establish any rule for this. But generally in the future I'll try to put utility/helper functions at the bottom of the files.

scripts/skills/perks/perk_rf_kingfisher.nut
```diff
@@ -35,82 +36,98 @@ this.perk_rf_kingfisher <- ::inherit("scripts/skills/skill", {
...
+		local throwNetSkill = this.getContainer().getSkillByID("actives.throw_net");
+		if (!throwNetSkill.onVerifyTarget(actor.getTile(), _targetEntity.getTile()))
+			return;
+
+		local netItemScript = ::IO.scriptFilenameByHash(actor.getOffhandItem().ClassNameHash);
+
+		throwNetSkill.useForFree(_targetEntity.getTile());
+
+		local netEffect = _targetEntity.getSkills().getSkillByID("effects.net");
+		if (netEffect == null) // Make target is actually netted. In vanilla net can be thrown on targets which are immune to it causing it to miss.
```
Darxo, line 52: in the comment a ` sure ` is missing between first and second word.

Darxo (approved):

PR #307, opened 2024-03-23
Rework tempo remove fluid weapon

LordMidas:
- Remove Fluid Weapon effect from Tempo perk
- Increase turn order initiative by 50%
- Recover action points when attacking opponents who act after you in a round

- Remove Fluid Weapon perk and status effect from Reforged codebase


scripts/skills/perks/perk_rf_tempo.nut
```diff
@@ -22,43 +24,46 @@ this.perk_rf_tempo <- ::inherit("scripts/skills/skill", {
...
+		}
 
 		if (this.m.IsPrimed)
 		{
 			tooltip.push({
 				id = 10,
 				type = "text",
 				icon = "ui/icons/warning.png",
-				text = ::MSU.Text.colorRed("This bonus has been carried over from the previous turn and will expire after using a skill or upon waiting or ending this turn")
+				text = ::MSU.Text.colorRed("The Initiative bonus has been carried over from the previous turn and will expire after using a skill or upon waiting or ending this turn")
```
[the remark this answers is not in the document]
LordMidas: Any skill/attack that might scale off of initiative. The most relevant case is that of fencers, so that they can use their first lunge on the next turn with the built-up initiative. If they lose that initiative then the perk becomes significantly less useful for them.

scripts/skills/perks/perk_rf_tempo.nut
```diff
@@ -1,8 +1,10 @@
 this.perk_rf_tempo <- ::inherit("scripts/skills/skill", {
 	m = {
 		BonusInitiative = 15,
+		InitiativeForTurnOrderMult = 1.5,
 		Stacks = 0,
 		IsPrimed = false,
```
Darxo, line 6: This needs an explanation about what this variable does/controls

mod_reforged/hooks/config/strings.nut
```diff
@@ -1898,18 +1886,14 @@ foreach (vanillaDesc in vanillaDescriptions)
...
-					"The bonus lasts over into your next [turn|Concept.Turn] but only until the first skill used or waiting that [turn|Concept.Turn]."
-				]
-			},
-			{
-				Type = ::UPD.EffectType.Passive,
-				Description = [
-					"When wielding a sword, gain the [Fluid Weapon|NullEntitySkill+rf_fluid_weapon_effect] effect."
+					"Your [Initiative|Concept.Initiative] is considered " + ::MSU.Text.colorGreen("50%") + " higher for calculating your [turn order|Concept.Turn].",
+					"Every hit or miss against a target who acts after you in the current round increases your [Initiative|Concept.Initiative] by " + ::MSU.Text.colorGreen("+15") + ". This bonus lasts over into your next [turn|Concept.Turn] but only until the first skill used or waiting that [turn|Concept.Turn].",
+					"The first two hits against opponents who act after you in the current [round|Concept.Round] recover " + ::MSU.Text.colorGreen("2") + " and " + ::MSU.Text.colorGreen("1") + " [Action Points|Concept.ActionPoints] respectively. Misses increment the counter, but do not recover any [Action Points|Concept.ActionPoints].",
```
Darxo, line 1896: The part where it first recovers 2 AP and then 1 AP is complicated in description and in code for barely any gain as I pointed out to you in discord. I would remove that part and make both attacks restore 2 AP.

I only remember you reply with some calculations but nothing else.
Have you reached a verdict about that?

scripts/skills/perks/perk_rf_tempo.nut
```diff
@@ -95,10 +96,11 @@ this.perk_rf_tempo <- ::inherit("scripts/skills/skill", {
 
 	function onBeforeTargetHit( _skill, _targetEntity, _hitInfo )
 	{
-		if (this.m.APRecovery.len() != 0 && _skill.isAttack() && ::Tactical.TurnSequenceBar.isActiveEntity(this.getContainer().getActor()) && !_targetEntity.isTurnDone() && !_targetEntity.isTurnStarted())
+		if (this.m.AttacksRemaining > 0 && _skill.isAttack() && ::Tactical.TurnSequenceBar.isActiveEntity(this.getContainer().getActor()) && !_targetEntity.isTurnDone() && !_targetEntity.isTurnStarted())
 		{
+			this.m.AttacksRemaining--;
```
Darxo, line 101: If I hit a target, whose turn is already done or whose turn has already startet, then none of my two AttacksRemaining are taken away from me.
The conditions `!_targetEntity.isTurnDone() && !_targetEntity.isTurnStarted()` must be put in an extra `if` after this line, in order to prevent that.

scripts/skills/perks/perk_rf_tempo.nut
```diff
@@ -95,10 +96,11 @@ this.perk_rf_tempo <- ::inherit("scripts/skills/skill", {
 
 	function onBeforeTargetHit( _skill, _targetEntity, _hitInfo )
 	{
-		if (this.m.APRecovery.len() != 0 && _skill.isAttack() && ::Tactical.TurnSequenceBar.isActiveEntity(this.getContainer().getActor()) && !_targetEntity.isTurnDone() && !_targetEntity.isTurnStarted())
+		if (this.m.AttacksRemaining > 0 && _skill.isAttack() && ::Tactical.TurnSequenceBar.isActiveEntity(this.getContainer().getActor()) && !_targetEntity.isTurnDone() && !_targetEntity.isTurnStarted())
```
Darxo, line 99: This perk has **Sword** as a requirement and you decided for this patch that all weapon tree perks are weapon locked, but I don't see anywhere in this file, where you enforce this restriction.

scripts/skills/perks/perk_rf_tempo.nut
```diff
@@ -107,6 +109,8 @@ this.perk_rf_tempo <- ::inherit("scripts/skills/skill", {
 	function onTargetMissed( _skill, _targetEntity )
 	{
 		this.gainStackIfApplicable(_skill, _targetEntity);
+		if (this.m.AttacksRemaining > 0 && ::Tactical.TurnSequenceBar.isActiveEntity(this.getContainer().getActor()))
```
Darxo, line 112: You do not check, whether _skill is an attack. But during the hit-check you check for that. That seems like an oversight, even if there are currently none/few? skills which are not an attack but can miss.

Darxo (changes requested):

scripts/skills/perks/perk_rf_tempo.nut
LordMidas: The `isAttack` check is redundant as both `onBeforeTargetHit` and `onTargetMissed` are called from `onScheduledTargetHit` and `attackEntity` functions of skill.nut which are for attacks only. So I will remove the check.

mod_reforged/hooks/config/strings.nut
```diff
@@ -1893,7 +1893,7 @@ foreach (vanillaDesc in vanillaDescriptions)
 				Description = [
 					"Your [Initiative|Concept.Initiative] is considered " + ::MSU.Text.colorGreen("50%") + " higher for calculating your [turn order|Concept.Turn].",
```
Darxo, line 1894: This effect is currently active all the time, even while not having a sword equipped. Is that intended?

Darxo (changes requested):

mod_reforged/hooks/config/strings.nut
LordMidas: Not intended. Fixed in new commit.

Darxo (approved):

PR #308, opened 2024-03-23
Midas perks rebalance misc

LordMidas:
These are generally small/simple changes so I combined them into a single PR for simplicity. For the review, it might make sense to go through this PR commit by commit because each commit is a complete thing in its own. This will make the review process easier.

Most of these changes are "untested" in-game. But because how simple they are, we can be fairly confident that they will work if the code looks ok.


mod_reforged/hooks/skills/perks/perk_mastery_cleaver.nut
```diff
@@ -0,0 +1,34 @@
...
+			return false;
+
+		local weapon = this.getContainer().getActor().getMainhandItem();
+		return weapon != null && weapon.isWeaponType(::Const.Items.WeaponType.Cleaver);
+	}
+
+	q.onTargetHit = @(__original) function( _skill, _targetEntity, _bodyPart, _damageInflictedHitpoints, _damageInflictedArmor )
+	{
+		__original(_skill, _targetEntity, _bodyPart, _damageInflictedHitpoints, _damageInflictedArmor);
+		if (_targetEntity.isAlive() && _damageInflictedHitpoints >= ::Const.Combat.MinDamageToApplyBleeding && !_targetEntity.getCurrentProperties().isImmuneToBleeding && _skill.isAttack() && _skill.m.IsWeaponSkill && this.isEnabled())
```
Darxo, line 29: This doesn't check whether the skill is actually coming from the equipped cleaver. If someone has a cleaver equipped but attacks with another skill (probably only possible with mods) then those attacks will also inflict a bleed stack. E.g. Hand-To-Hand being always available. Or a kick ability that deals damage.

You could check `_skill.getItem()` of the passed skill.

But even then the cleaver could be a hybrid weapon with a piercing/blunt attack. Should that attack then also inflict bleeds?

mod_reforged/hooks/config/strings.nut
```diff
@@ -460,6 +460,23 @@ local vanillaDescriptions = [
...
+	{
+		ID = "perk.mastery.cleaver",
+		Key = "SpecCleaver",
+		Description = ::UPD.getDescription({
+			Fluff = "Master cleavers to inflict gruesome wounds.",
+			Requirement = "Cleaver",
+	 		Effects = [{
+ 				Type = ::UPD.EffectType.Passive,
+ 				Description = [
+					"Skills build up " + ::MSU.Text.colorRed("25%") + " less [Fatigue|Concept.Fatigue].",
```
Darxo, line 472: this is a positive effect for the character so it should be colorGreen

mod_reforged/hooks/skills/actives/gash_skill.nut
```diff
@@ -35,6 +37,16 @@
 			});
 		}
 
+		if (this.m.BleedStacks != 0)
+		{
+			ret.push({
+				id = 6,
+				type = "text",
+				icon = "ui/icons/special.png",
+				text = ::Reforged.Mod.Tooltips.parseString("Applies " + ::MSU.Text.colorGreen(this.m.BleedStacks) + " stacks of [Bleeding|Skill+bleeding_effect] when inflicting at least " + ::MSU.Text.colorRed(::Const.Combat.MinDamageToApplyBleeding) + " damage")
```
Darxo, line 46: How do you define the coloring for thresholds?
Here you use red for a threshold, which is really neither positive nor negative, it's just not clear whether this number is good or bad for the player. 
If your only concern was to highlight this number then I suggest using using colorNeutral for this.

There is not a clear vanilla convention for these numbers when poison_coat_effect colors is red but fearsome perk description does not use color for these thresholds.

scripts/skills/perks/perk_rf_iron_sights.nut
```diff
@@ -15,16 +15,30 @@ this.perk_rf_iron_sights <- ::inherit("scripts/skills/skill", {
 		this.m.IsHidden = false;
 	}
 
+	function isEnabled()
```
Darxo, line 18: Depending on the outcome of the tech discussion:
- Since this is a newly introduced function it should at least be sorted to the bottom, ideally with a comment introducing that section

mod_reforged/hooks/config/strings.nut
```diff
@@ -1249,9 +1249,8 @@ foreach (vanillaDesc in vanillaDescriptions)
  		Effects = [{
 			Type = ::UPD.EffectType.Passive,
 			Description = [
-				"Successful hits from Flails progressively reduce the target's [Melee Defense|Concept.MeleeDefense] and [Ranged Defense|Concept.RangeDefense] by a stacking " + ::MSU.Text.colorRed(-3) + " for one [turn|Concept.Turn].",
+				"Successful hits from Flails progressively reduce the target's [Melee Defense|Concept.MeleeDefense] and [Ranged Defense|Concept.RangeDefense] by a stacking " + ::MSU.Text.colorRed(-5) + " for one [turn|Concept.Turn].",
```
Darxo, line 1252: I don't see where this change is reflected in the from all sides effect

Darxo (changes requested):

scripts/skills/perks/perk_rf_the_rush_of_battle.nut
```diff
@@ -66,6 +66,14 @@ this.perk_rf_the_rush_of_battle <- ::inherit("scripts/skills/skill", {
 		}
 	}
 
+	function onAnySkillExecuted( _skill, _targetTile, _targetEntity, _forFree )
+	{
+		if (_skill.isAttack())
```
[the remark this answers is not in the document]
LordMidas: Are there any attack skills that can target the actor himself? I think we can and should assume that this cannot be the case because I believe the code of the vanilla game wherever they check for `skill.isAttack()` is also written with this assumption. The check in `onBeforeDamageReceived` is to account for things like bleeding/poison damage in which case the source of the damage is the actor receiving the damage.

mod_reforged/hooks/config/strings.nut
LordMidas: Ah yeah. I just followed the vanilla style here. I think we decided to change this to green, so I'll do that.

mod_reforged/hooks/skills/actives/gash_skill.nut
LordMidas: Anything other than green is acceptable for this here. I think the red works fine as it shows a limit you must overcome for the perk to trigger.

scripts/skills/perks/perk_rf_the_rush_of_battle.nut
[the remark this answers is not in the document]
Darxo: Nevermind I thought this function was called whenever any entity on the battlefield uses any skill, because I remember how other classes who use this function always check `_skill == this`.
This seems to be correct

mod_reforged/hooks/config/strings.nut
LordMidas: Actually doing it just like this will make it inconsistent with the other tooltips. I suggest we make that change in a global commit later in entire codebase once we have finalized which values we want to be green and red.

mod_reforged/hooks/skills/perks/perk_bullseye.nut
```diff
@@ -0,0 +1,13 @@
+::Reforged.HooksMod.hook("scripts/skills/perks/perk_bullseye", function(q) {
+	q.onAnySkillUsed = @(__original) function( _skill, _targetEntity, _properties )
+	{
+		if (!_skill.isRanged() || _targetEntity != null)
```
Darxo, line 4: The second condition must be `_targetEntity == null`.

mod_reforged/hooks/skills/perks/perk_mastery_cleaver.nut
```diff
@@ -0,0 +1,46 @@
...
+	}
+
+	q.onQueryTooltip = @(__original) function( _skill, _tooltip )
+	{
+		__original(_skill, _tooltip);
+
+		if (this.isSkillValid(_skill))
+		{
+			_tooltip.push({
+				id = _tooltip().top().id + 1,
```
Darxo, line 33: this should have a hard-coded id. I don't quite remember where you wanted onQueryTooltips to start though, was it 1000?

We might even want to use the same id for every skill. That way a submod can at least remove them all at once or search in them for a specific name.
A different id for every line is a very hard thing to guarantee even just within reforged as this has to be maintained over every different skill.

scripts/skills/perks/perk_rf_iron_sights.nut
```diff
@@ -15,16 +15,30 @@ this.perk_rf_iron_sights <- ::inherit("scripts/skills/skill", {
...
 	function onUpdate( _properties )
 	{
-		if (this.getContainer().getActor().isDisarmed()) return;
+		if (this.isEnabled())
+			_properties.HitChance[::Const.BodyPart.Head] += this.m.Bonus;
+	}
+
+	function onTargetHit( _skill, _targetEntity, _bodyPart, _damageInflictedHitpoints, _damageInflictedArmor )
+	{
+		if (!_targetEntity.isAlive() || _bodyPart != ::Const.BodyPart.Head || _targetEntity.getMoraleState() == ::Const.MoraleState.Ignore || !_skill.isRanged() || !_skill.isAttack() || !this.isEnabled())
```
Darxo, line 35: How does the morale state `::Const.MoraleState.Ignore` make an enemy immune to being shellshocked?

I think this condition only existed in vanilla because they do a negative morale check alongside applying shellshocked.
And I even consider it a vanilla bug, that mortars can't apply shellshocked on enemies which ignore Morale.

scripts/skills/perks/perk_rf_bolster.nut
```diff
@@ -15,38 +16,33 @@ this.perk_rf_bolster <- ::inherit("scripts/skills/skill", {
...
-					{
-						this.spawnIcon("perk_rf_bolster", ally.getTile());
-					}
-				}
-			}
-		}
+		if (_skill.isRanged() || !_skill.isAttack())
+			return false;
+
+		if (this.m.RequiredWeaponType == null)
```
Darxo, line 42: If this is the only time you use `RequiredWeaponType` you could also change it into a bool and call it `RequiresWeapon`.

Otherwise you probably want to include an additional check for this variable in the big return condition below

Darxo (changes requested):

scripts/skills/perks/perk_rf_iron_sights.nut
LordMidas: I think vanilla did this to prevent undead units from getting shellshocked which kinda makes sense because they don't suffer from mental and physical trauma like that. I just added this condition to be in line with vanilla conditions for applying shellshocked. But we can discuss it and can replace it with a check for the entity being undead instead of checking the morale state.

LordMidas: That's a good idea, I'll add it to the tooltip.

Darxo (approved):

mod_reforged/hooks/skills/perks/perk_mastery_cleaver.nut
Uberbagel, line 1: Bloodlust not properly appling. It shows up in hover-over tooltip without showing how many stacks it has. It does not show up as a mini-icon on the effects bar at the bottom. No effects from Bloodlust are properly applying.

mod_reforged/hooks/skills/perks/perk_bullseye.nut
Uberbagel, line 1: Should the hover-over tooltip when you have bullseye indicate when the armor-pen portion of bullseye will trigger on the target?

mod_reforged/hooks/skills/perks/perk_mastery_cleaver.nut
LordMidas: This branch hasn't been rebased on development so it does not have the reworked Bloodlust perk. That perk has been tested separately in its own branch and has been merged into development.

mod_reforged/hooks/skills/perks/perk_bullseye.nut
Uberbagel: Bear Down has an indicator, just says + Bear Down

LordMidas: We could add it. Can appear as `+ Bullseye`. A concern is that it may get confused with a hit-chance bonus rather than an armor-pen bonus because that line isn't specific. But I wouldn't want to add text that goes beyond a single line.

mod_reforged/hooks/skills/perks/perk_mastery_sword.nut
Uberbagel, line 1: Kata Step not appearing on characters with sword mastery and an applicable sword.

scripts/skills/actives/rf_bestial_vigor_skill.nut
Uberbagel, line 1: Perk and skill work but there's a tooltip error.
image

Uberbagel (changes requested):
Shelving review and leaving this to an unstable build

mod_reforged/hooks/skills/perks/perk_mastery_sword.nut
LordMidas: Pushed a fix for this.

scripts/skills/actives/rf_bestial_vigor_skill.nut
Darxo: This is the PR in question that contains the `colorizeFraction` function https://github.com/Battle-Modders/mod-reforged/pull/291

Uberbagel (approved):
Approved for unstable build

PR #310, opened 2024-03-23
Mentors presence tweaks

LordMidas:
- As mentor: Don't gain bonuses from student's death when you are fleeing or stunned
- As student: Don't gain morale bonus at turn start when mentor is fleeing


Darxo (approved):
Looks good.

PR #311, opened 2024-03-23
feat: add unique double grip bonus for each weapon type

LordMidas:


mod_reforged/hooks/skills/special/double_grip.nut
```diff
@@ -0,0 +1,307 @@
...
+
+		return offhand.getStaminaModifier() < 10 && this.m.CurrWeaponType == "SouthernSword" && this.getContainer().hasSkill("perk.rf_en_garde");
+	}
+
+	q.applyBonusOnUpdate <- function( _properties )
+	{
+		switch (this.m.CurrWeaponType)
+		{
+			case ::Const.Items.WeaponType.Axe:
+				_properties.MeleeDamageMult *= 1.15;
```
Darxo, line 31: All of these values should be made into member variables.

LordMidas: That is too much in this case imo as there's too many values. I instead implemented these separate functions with a switch statement so submodders who want to modify a particular weapon's behavior can hook these functions to easily modify the behavior of a particular weapon.

Darxo (approved):
I don't care about the reuse of IDs in this case because a submod will likely just rewrite a whole batch at once anyways.

PR #312, opened 2024-03-23
refactor: disable shield defense penalty from fatigue during shieldwall

LordMidas:


mod_reforged/hooks/skills/actives/shieldwall.nut
```diff
@@ -18,6 +18,13 @@
 			text = ::Reforged.Mod.Tooltips.parseString("Grants immunity against [Hook Shield|Skill+rf_hook_shield_skill]")
 		});
 
+		tooltip.push({
+			id = 6,
```
Darxo, line 22: should have a unique id to improve submodding. Anything from 7 onwards is unused.

mod_reforged/hooks/skills/actives/shieldwall.nut
```diff
@@ -18,6 +18,13 @@
...
+			type = "text",
+			icon = "ui/icons/special.png",
```
Darxo, line 24: In the shield.nut you give the very similar tooltip entry about this mechanic the icon `icon = "ui/icons/fatigue.png"`. I much prefer that over a generic `special` icon whenever possible

Darxo (changes requested):

mod_reforged/hooks/skills/actives/shieldwall.nut
LordMidas: In this particular case I think special is better than the fatigue icon because in the `shield.nut` it is an effect that brother is gaining because of building fatigue. So the fatigue icon (or maybe a melee defense icon) is more representative in that case. Whereas here this perk is removing the penalty which is already present - so the special icon (or a melee defense icon) is more representative of what this perk does. I avoided using the melee defense icon because that is already used in the tooltip of these things so I didn't want to spam it.

mod_reforged/hooks/items/shields/shield.nut
```diff
@@ -20,7 +20,7 @@
 			id = 8,
 			type = "text",
 			icon = "ui/icons/fatigue.png",
-			text = ::Reforged.Mod.Tooltips.parseString("The defense values drop linearly as you build [Fatigue|Concept.Fatigue] up to " + ::MSU.Text.colorRed((maxReduction * 100) + "%") + " at maximum [Fatigue|Concept.Fatigue]")
+			text = ::Reforged.Mod.Tooltips.parseString("Except when using [Shieldwall|Skill+shieldwall], the defense values drop linearly as you build [Fatigue|Concept.Fatigue] up to " + ::MSU.Text.colorRed((maxReduction * 100) + "%") + " at maximum [Fatigue|Concept.Fatigue]")
```
Darxo, line 23: `::MSU.Text.colorRed((maxReduction * 100) + "%")`
could be turned into
`::MSU.Text.colorizeMult(maxReduction)`

Darxo (approved):
I don't care how you resolve my last comment. Either way works, as the variable in question is not strictly speaking a multiplier

PR #313, opened 2024-03-24
Rework swordmaster perks

LordMidas:
I'd suggest going through this one commit by commit. The commit messages describe the changes.


Darxo (changes requested):
There are a lot of
~~- missing commas at the end of arrays/tables~~
- duplicate ids used for tooltip lines

in this PR as it was made before we officially added rules for those cases.

Please fix those first so I don't have to write so many comments.

Darxo:
I'm overwhelmed by the changes. Because you added commits at the very end which change/fix things from earlier commits I don't feel like going through it commit by commit because many of my findings would be wasted time.

And I'm missing a lot of context when looking at the changes all at once

LordMidas:
> I'm overwhelmed by the changes. You added commits at the very end which change/fix things from earlier commits so I don't feel like going through it commit by commit because many of my findings would be wasted time.
> 
> And I'm missing a lot of context when looking at the changes all at once

The fix commits I've added in this branch fix existing bugs with these abilities i.e. they are not fixes for previous commits in this branch. Each commit is a standalone change and should be gone through one by one for all the changes to individual perks. Nevertheless, I've tested these in-game and they work fine.

PR #315, opened 2024-03-24
refactor: require recover to re-enable fresh and furious

LordMidas:
Now checks fatigue against the threshold at the start of a turn. So going above the threshold during a turn will not disable it (i.e. require recover to re-activate it) as long as you can start your next turn below the threshold.

### DIFF
--- a/mod_reforged/hooks/config/strings.nut
+++ b/mod_reforged/hooks/config/strings.nut
@@ -1203,8 +1203,8 @@ foreach (vanillaDesc in vanillaDescriptions)
  		Effects = [{
 			Type = ::UPD.EffectType.Passive,
 			Description = [
-				"As long as your current [Fatigue|Concept.Fatigue] is below " + ::MSU.Text.colorRed("30%") + " of [Maximum Fatigue|Concept.MaximumFatigue] after gear, the [Action Point|Concept.ActionPoints] cost of the first skill used every [turn|Concept.Turn] is " + ::MSU.Text.colorGreen("halved") + ".",
-				"Does not expire when using a skill with no [Action Point|Concept.ActionPoints] and [Fatigue|Concept.Fatigue] cost."
+				"The [Action Point|Concept.ActionPoints] cost of the first non-free skill used every [turn|Concept.Turn] is " + ::MSU.Text.colorGreen("halved") + ".",
+				"The effect becomes disabled if you start a [turn|Concept.Turn] with " + ::MSU.Text.colorRed("30%") + " or more [Fatigue|Concept.Fatigue] built and remains disabled until you use [Recover|Skill+recover]."
 			]
 		}]
  	}),
--- a/scripts/skills/perks/perk_rf_fresh_and_furious.nut
+++ b/scripts/skills/perks/perk_rf_fresh_and_furious.nut
@@ -64,7 +64,7 @@ this.perk_rf_fresh_and_furious <- ::inherit("scripts/skills/skill", {
 
 	function onAfterUpdate( _properties )
 	{
-		if (!this.m.IsSpent && !this.m.RequiresRecover && this.isEnabled())
+		if (!this.m.IsSpent && !this.m.RequiresRecover)
 		{
 			foreach (skill in this.getContainer().getAllSkillsOfType(::Const.SkillType.Active))
 			{
--- a/scripts/skills/perks/perk_rf_fresh_and_furious.nut
+++ b/scripts/skills/perks/perk_rf_fresh_and_furious.nut
@@ -1,7 +1,9 @@
 this.perk_rf_fresh_and_furious <- ::inherit("scripts/skills/skill", {
 	m = {
 		IsUsingFreeSkill = false,
-		IsSpent = true
+		IsSpent = true,
+		RequiresRecover = false,
+		FatigueThreshold = 0.3
 	},
 	function create()
 	{
@@ -45,7 +47,7 @@ this.perk_rf_fresh_and_furious <- ::inherit("scripts/skills/skill", {
 
 	function isEnabled()
 	{
-		return this.getContainer().getActor().getFatigue() < 0.3 * this.getContainer().getActor().getFatigueMax();
+		return !this.m.RequiresRecover && this.getContainer().getActor().getFatigue() < this.m.FatigueThreshold * this.getContainer().getActor().getFatigueMax();
 	}
 
 	function onAfterUpdate( _properties )
@@ -83,16 +85,23 @@ this.perk_rf_fresh_and_furious <- ::inherit("scripts/skills/skill", {
 	function onAnySkillExecuted( _skill, _targetTile, _targetEntity, _forFree )
 	{
 		this.m.IsSpent = !this.m.IsUsingFreeSkill;
+
+		if (_skill.getID() == "actives.recover")
+			this.m.RequiresRecover = false;
 	}
 
 	function onTurnStart()
 	{
 		this.m.IsSpent = false;
+
+		if (this.getContainer().getActor().getFatigue() >= this.m.FatigueThreshold * this.getContainer().getActor().getFatigueMax())
+			this.m.RequiresRecover = true;
 	}
 
 	function onCombatFinished()
 	{
 		this.skill.onCombatFinished();
 		this.m.IsSpent = true;
+		this.m.RequiresRecover = false;
 	}
 });
--- a/scripts/skills/perks/perk_rf_fresh_and_furious.nut
+++ b/scripts/skills/perks/perk_rf_fresh_and_furious.nut
@@ -28,18 +28,30 @@ this.perk_rf_fresh_and_furious <- ::inherit("scripts/skills/skill", {
 	{
 		local tooltip = this.skill.getTooltip();
 
-		tooltip.push({
-			id = 10,
-			type = "text",
-			icon = "ui/icons/special.png",
-			text = "The next skill used has its Action Point cost [color=" + ::Const.UI.Color.PositiveValue + "]halved[/color]"
-		});
+		if (this.m.RequiresRecover)
+		{
+			tooltip.push({
+				id = 10,
+				type = "text",
+				icon = "ui/icons/warning.png",
+				text = ::MSU.Text.colorRed("Disabled until this character uses [Recover|Skill+recover]")
+			});
+		}
+		else
+		{
+			tooltip.push({
+				id = 11,
+				type = "text",
+				icon = "ui/icons/special.png",
+				text = "The next non-free skill used has its Action Point cost [color=" + ::Const.UI.Color.PositiveValue + "]halved[/color]"
+			});
+		}
 
 		tooltip.push({
-			id = 10,
+			id = 12,
 			type = "text",
 			icon = "ui/icons/warning.png",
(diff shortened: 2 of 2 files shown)


mod_reforged/hooks/config/strings.nut
```diff
@@ -1203,8 +1203,8 @@ foreach (vanillaDesc in vanillaDescriptions)
  		Effects = [{
 			Type = ::UPD.EffectType.Passive,
 			Description = [
-				"As long as your current [Fatigue|Concept.Fatigue] is below " + ::MSU.Text.colorRed("30%") + " of [Maximum Fatigue|Concept.MaximumFatigue] after gear, the [Action Point|Concept.ActionPoints] cost of the first skill used every [turn|Concept.Turn] is " + ::MSU.Text.colorGreen("halved") + ".",
-				"Does not expire when using a skill with no [Action Point|Concept.ActionPoints] and [Fatigue|Concept.Fatigue] cost."
+				"As long as you start your [turn|Concept.Turn] with " + ::MSU.Text.colorRed("30%") + " or less built [Fatigue|Concept.Fatigue], the [Action Point|Concept.ActionPoints] cost of the first skill used is " + ::MSU.Text.colorGreen("halved") + ". Does not expire when using a skill with no [Action Point|Concept.ActionPoints] and [Fatigue|Concept.Fatigue] cost.",
```
Darxo, line 1206: I don't see the part worthwhile to keep where you can enable this effect when you naturally hit 0 Fatigue. 
Because what is the guy doing in their turn which allows them to naturally reach 0 Fatigue? If they are doing nothing they might aswell use Recover. It's just bloat that may become relevant in 1% of situations.

Suggestion for a cleaner effect text with that change in mind:
```
"The [Action Point|Concept.ActionPoints] cost of the first skill used every [turn|Concept.Turn] is " + ::MSU.Text.colorGreen("halved") + "."
"This effect becomes disabled if you start your turn with 30% or more Fatigue built up, until you use Recover."
```

scripts/skills/perks/perk_rf_fresh_and_furious.nut
```diff
@@ -45,11 +48,31 @@ this.perk_rf_fresh_and_furious <- ::inherit("scripts/skills/skill", {
...
+
+			if (this.getContainer().getActor().getFatigue() == 0)
+			{
+				this.m.RequiresRecover = false;
+			}
+			else
+			{
+				if (this.m.RequiresRecover || this.getContainer().getActor().getFatigue() >= this.m.FatigueThreshold * this.getContainer().getActor().getFatigueMax())
+				{
+					this.m.RequiresRecover = true;
```
Darxo, line 70: This must be `this.m.RequiresRecover = false;`

scripts/skills/perks/perk_rf_fresh_and_furious.nut
```diff
@@ -45,11 +48,31 @@ this.perk_rf_fresh_and_furious <- ::inherit("scripts/skills/skill", {
...
+		{
+			this.m.IsStartingTurn = false;
+
+			if (this.getContainer().getActor().getFatigue() == 0)
+			{
+				this.m.RequiresRecover = false;
+			}
+			else
+			{
+				if (this.m.RequiresRecover || this.getContainer().getActor().getFatigue() >= this.m.FatigueThreshold * this.getContainer().getActor().getFatigueMax())
```
Darxo, line 68: `this.m.RequiresRecover ||` is wrong and also redundant unless you are aiming for performance. 
But setting a bool variable to a new value is not worth improving the performance around.

Darxo (changes requested):

scripts/skills/perks/perk_rf_fresh_and_furious.nut
LordMidas: No it is correct. If the fatigue is greater than 30% then we need to set `RequiresRecover` to true.

scripts/skills/perks/perk_rf_fresh_and_furious.nut
LordMidas: The alternative was the following:
```squirrel
if (this.m.RequiresRecover)
   return;

if (this.getContainer().getActor().getFatigue() >= this.m.FatigueThreshold * this.getContainer().getActor().getFatigueMax())
{
     this.m.RequiresRecover = true;
     return;
}
```
I just merged them into a single if because if the bool was already true, setting it again to true doesn't hurt and the performance impact is negligible. But I can change it to this more elaborate code setup if you prefer this.

scripts/skills/perks/perk_rf_fresh_and_furious.nut
```diff
@@ -45,11 +48,31 @@ this.perk_rf_fresh_and_furious <- ::inherit("scripts/skills/skill", {
 
 	function isEnabled()
 	{
-		return this.getContainer().getActor().getFatigue() < 0.3 * this.getContainer().getActor().getFatigueMax();
+		return !this.m.RequiresRecover && this.getContainer().getActor().getFatigue() < this.m.FatigueThreshold * this.getContainer().getActor().getFatigueMax();
 	}
 
 	function onAfterUpdate( _properties )
 	{
+		// We do it like this instead of checking it in onTurnStart directly because some skills e.g. Encumbrance
```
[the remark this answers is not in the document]
LordMidas: Unfortunately it's not just encumbrance i.e. we can't predict what other effects can be added by submods which modify the Fatigue value of the actor in `onTurnStart`. And we cannot predict those effects' skill order. Therefore, this is the safest bet.

scripts/skills/perks/perk_rf_fresh_and_furious.nut
Darxo: I would just remove that condition outright. Currently it is redundant and I don't see that changing any time soon.
So just
```
if (this.getContainer().getActor().getFatigue() >= this.m.FatigueThreshold * this.getContainer().getActor().getFatigueMax())
{
     this.m.RequiresRecover = true;
     return;
}
```

LordMidas: But we have to return from the function if RequiresRecover is already true. We can't remove that condition.

scripts/skills/perks/perk_rf_fresh_and_furious.nut
Darxo: you're right

scripts/skills/perks/perk_rf_fresh_and_furious.nut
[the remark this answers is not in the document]
Darxo: By that same logic we can't predict what other skills might also do this same work-around in order to apply some elaborate fatigue changes, e.g. when certain fatigue thresholds are met during start up.

Someone might implement a skill that recovers 5 Fatigue on turn start if you have at least X fatigue build up. And then they employ the same tech as you just did up there.

---

I much rather prefer lowering the skill order of encumbrance, rather than implementing these work around in every skill which checks any play resource on turn start.

Anyone who implements resource manipulating effects on turn start need to learn the easy/hard way that those skills need a lower skill order or other effects will not recognize those changes.

scripts/skills/perks/perk_rf_fresh_and_furious.nut
Darxo: I don't see the issue in not returning early. The second half of this elabtore `onAfterUpdate` body is gated by `this.isEnabled()` which checks `RequiresRecover`

mod_reforged/hooks/config/strings.nut
LordMidas: I kinda agree we can simplify it like that. Let's do that and see how it goes.

scripts/skills/perks/perk_rf_fresh_and_furious.nut
```diff
@@ -83,16 +97,30 @@ this.perk_rf_fresh_and_furious <- ::inherit("scripts/skills/skill", {
 	function onAnySkillExecuted( _skill, _targetTile, _targetEntity, _forFree )
 	{
 		this.m.IsSpent = !this.m.IsUsingFreeSkill;
+
+		if (_skill.getID() == "actives.recover")
+		{
+			this.m.RequiresRecover = false;
+			if (this.isEnabled())
```
Darxo, line 104: This condition is redundant and not in line with how that effect actually works.

Currently this effect only becomes disabled at the start of a turn, not at any other points in time. And using Recover always activates it, regardless of any other conditions.

scripts/skills/perks/perk_rf_fresh_and_furious.nut
```diff
@@ -26,31 +28,43 @@ this.perk_rf_fresh_and_furious <- ::inherit("scripts/skills/skill", {
...
+				text = ::MSU.Text.colorRed("Disabled until this character uses [Recover|Skill+recover]")
+			});
+		}
+		else
+		{
+			tooltip.push({
+				id = 11,
+				type = "text",
+				icon = "ui/icons/special.png",
+				text = "The next non-free skill used has its Action Point cost [color=" + ::Const.UI.Color.PositiveValue + "]halved[/color]"
```
Darxo, line 46: From the gitlab POV it looks like you are touch all this text at once. At that point you might aswell fix the coloring here by using `::MSU.Text.colorGreen("halved")`

Darxo (changes requested):

Darxo (approved):

scripts/skills/perks/perk_rf_fresh_and_furious.nut
```diff
@@ -26,31 +28,43 @@ this.perk_rf_fresh_and_furious <- ::inherit("scripts/skills/skill", {
...
 	function isEnabled()
 	{
-		return this.getContainer().getActor().getFatigue() < 0.3 * this.getContainer().getActor().getFatigueMax();
+		return this.getContainer().getActor().getFatigue() < this.m.FatigueThreshold * this.getContainer().getActor().getFatigueMax();
 	}
 
 	function onAfterUpdate( _properties )
 	{
-		if (!this.m.IsSpent && this.isEnabled())
+		if (!this.m.IsSpent && !this.m.RequiresRecover && this.isEnabled())
```
Darxo, line 67: The perk description says that this skill only becomes enabled **at the start of your turn**, therefor not at any other point in time.

But here it also stops working once your fatigue goes over 30% at any point in time during an onAfterUpdate.

Darxo (approved):

scripts/skills/perks/perk_rf_fresh_and_furious.nut
```diff
@@ -12,6 +13,7 @@ this.perk_rf_fresh_and_furious <- ::inherit("scripts/skills/skill", {
 		this.m.Description = "This character is exceptionally fast when not fatigued.";
 		this.m.Icon = "ui/perks/rf_fresh_and_furious.png";
 		this.m.IconMini = "rf_fresh_and_furious_mini";
+		this.m.IconMiniBackup = this.m.IconMini;
```
Darxo, line 16: I would sort this line to the bottom of the block, even starting a new block with an empty line in between. That is to emphasize, that this/these variables are non-vanilla ones.

Darxo (approved):
My last comment here is more feedback and less a remark. You can merge either way

PR #316, opened 2024-03-24
feat: new enemy-only headless perk

Darxo:
This branch is currently pure QoL by displaying the player exactly which enemies are "headless" and therefor only damage body-hits. This is done by implementing a new **headless** perk, which re-directs all hits to the body.

Future headless enemies just need this perk instead of implementing the headless effect silently in the onDamageReceived function of that actor.

TODO:
- Icon for this perk


Darxo:
This perk could grant **Immune to Distracted** (pocket sand) by default. You can't put sand in the eyes of someone who has no head.

PR #318, opened 2024-03-25
docs: add entity names as comments to skeleton files

Uberbagel:
Per Darxo's request, which I agree with.

### DIFF
--- a/mod_reforged/hooks/entity/tactical/enemies/skeleton_light.nut
+++ b/mod_reforged/hooks/entity/tactical/enemies/skeleton_light.nut
@@ -1,3 +1,4 @@
+// Ancient Auxiliary
 ::Reforged.HooksMod.hook("scripts/entity/tactical/enemies/skeleton_light", function(q) {
 	q.onInit = @() function()
 	{
--- a/scripts/entity/tactical/enemies/rf_skeleton_heavy_lesser.nut
+++ b/scripts/entity/tactical/enemies/rf_skeleton_heavy_lesser.nut
@@ -1,3 +1,4 @@
+// Ancient Praetorian
 this.rf_skeleton_heavy_lesser <- ::inherit("scripts/entity/tactical/skeleton", {
 	m = {},
 	function create()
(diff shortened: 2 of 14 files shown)


mod_reforged/hooks/entity/tactical/enemies/skeleton_light.nut
```diff
@@ -1,3 +1,4 @@
+//Ancient Auxiliary
```
Darxo, line 1: Using the very first line for that feels a bit odd.
Maybe it should go on the right side of the HooksMod/inherit line? Separated by a single tab or space
I would also add a single space bar after the slashes.

e.g.:
`::Reforged.HooksMod.hook("scripts/entity/tactical/enemies/skeleton_light", function(q) { // Ancient Auxiliary`

scripts/entity/tactical/enemies/rf_skeleton_heavy_lesser.nut
```diff
@@ -1,3 +1,4 @@
+//Ancient Praetorian
```
Darxo, line 1: Using the very first line for that feels a bit odd.
Maybe it should go on the right side of the inherit line? Separated by a single tab or space
I would also add a single space bar after the slashes.

e.g.:
`this.rf_skeleton_heavy_lesser <- ::inherit("scripts/entity/tactical/skeleton", { // Ancient Praetorian`

Uberbagel: I'm neutral as to which is nicer. Will see what Midas thinks

LordMidas: I prefer the comment on the first line, as it is much more visible the moment you open the file. But I agree with Darxo that there should be a space after the slashes so `// Ancient Praetorian`

Darxo: Alright, let's leave it in the first row with an additional space each

Uberbagel: Done

Darxo (approved):

LordMidas (approved):

PR #321, opened 2024-03-25
refactor: rework arrow to the knee skill and effect

LordMidas:
- The active skill now inherits from quick_shot. This ensures that many effects which weren't properly accounted for in the original implementation but are implemented in vanilla e.g. _properties.IsSharpshooter are now automatically applied. Also cleans up the code quite a bit.
- Reduced AP cost to 4.
- Now applies Staggered effect as well.
- Removed defense malus from arrow to the knee debuff.

### DIFF
--- a/scripts/skills/actives/rf_arrow_to_the_knee_skill.nut
+++ b/scripts/skills/actives/rf_arrow_to_the_knee_skill.nut
@@ -1,91 +1,37 @@
-this.rf_arrow_to_the_knee_skill <- ::inherit("scripts/skills/skill", {
-	m = {
-		AdditionalAccuracy = 0,
-		AdditionalHitChance = -2
-	},
+this.rf_arrow_to_the_knee_skill <- ::inherit("scripts/skills/actives/quick_shot", {
+	m = {},
 	function create()
 	{
+		this.quick_shot.create();
 		this.m.ID = "actives.rf_arrow_to_the_knee";
 		this.m.Name = "Arrow to the Knee";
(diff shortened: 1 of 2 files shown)


scripts/skills/actives/rf_arrow_to_the_knee_skill.nut
```diff
@@ -121,94 +67,21 @@ this.rf_arrow_to_the_knee_skill <- ::inherit("scripts/skills/skill", {
...
-	}
-
 	function onAnySkillUsed( _skill, _targetEntity, _properties )
 	{
+		this.quick_shot.onAnySkillUsed(_skill, _targetEntity, _properties);
 		if (_skill == this)
 		{
-			_properties.RangedSkill += this.m.AdditionalAccuracy;
-			_properties.HitChanceAdditionalWithEachTile += this.m.AdditionalHitChance;
-			_properties.RangedDamageMult *= 0.5;
```
Darxo, line 209: The description of the skill still says that this attack deals reduced damage but you removed the damage multiplier now. Something is wrong here.

If the damage multiplier is meant to stay then it should be implemented as a member variable.

Darxo (changes requested):
Looks good, except the thing I pointed out

scripts/skills/actives/rf_arrow_to_the_knee_skill.nut
LordMidas: Yeah removal of the damage multiplier (and the defense malus) is intended. I forgot to update the description.

Darxo (approved):

PR #322, opened 2024-03-25
refactor: rework target practice

LordMidas:
- Require bow
- Remove ability to swap ammo for free
- Change bonus to affect enemies not in cover instead of enemies with no adjacent ally

### DIFF
--- a/scripts/skills/perks/perk_rf_offhand_training.nut
+++ b/scripts/skills/perks/perk_rf_offhand_training.nut
@@ -104,10 +104,7 @@ this.perk_rf_offhand_training <- ::inherit("scripts/skills/skill", {
 
 	function onPayForItemAction(_skill, _items)
 	{
-		if (_skill != null && _skill.getID() != "perk.rf_target_practice")
-		{
-			this.m.IsSpent = true;
-		}
+		this.m.IsSpent = true;
 	}
 
 	function onAnySkillExecuted( _skill, _targetTile, _targetEntity, _forFree )
--- a/scripts/skills/perks/perk_rf_target_practice.nut
+++ b/scripts/skills/perks/perk_rf_target_practice.nut
@@ -16,11 +16,21 @@ this.perk_rf_target_practice <- ::inherit("scripts/skills/skill", {
 		this.m.ItemActionOrder = ::Const.ItemActionOrder.First;
 	}
 
+	function isEnabled()
+	{
+		if (this.getContainer().getActor().isDisarmed())
+			return false;
+
+		local weapon = this.getContainer().getActor().getMainhandItem();
+		return weapon != null && weapon.isWeaponType(::Const.Items.WeaponType.Bow);
+	}
+
 	function onAnySkillUsed( _skill, _targetEntity, _properties )
 	{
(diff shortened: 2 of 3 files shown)


scripts/skills/perks/perk_rf_target_practice.nut
```diff
@@ -16,40 +16,38 @@ this.perk_rf_target_practice <- ::inherit("scripts/skills/skill", {
...
+		if (!_skill.isRanged() || !_skill.m.IsWeaponSkill || !this.isEnabled())
+			return;
+
 		local targetEntity = _targetTile.getEntity();
-		if (_skill.isRanged() && targetEntity != null && (targetEntity.isArmedWithRangedWeapon() || ::Tactical.Entities.getAlliedActors(targetEntity.getFaction(), targetEntity.getTile(), 1, true).len() == 0))
+		if (targetEntity != null && (targetEntity.isArmedWithRangedWeapon() || ::Const.Tactical.Common.getBlockedTiles(this.getContainer().getActor().getTile(), targetEntity.getTile(), this.getContainer().getActor().getFaction()).len() == 0))
 		{
 			_tooltip.push({
 				icon = "ui/tooltips/positive.png",
 				text = "[color=" + ::Const.UI.Color.PositiveValue + "]" + this.m.RangedSkillBonus + "%[/color] " + this.getName()
```
Darxo, line 49: Might aswell improve the tooltip coloring if you are already working on this perk
`text = ::MSU.Text.colorGreen(this.m.RangedSkillBonus + "%") + this.getName()`

scripts/skills/perks/perk_rf_offhand_training.nut
```diff
@@ -104,10 +104,7 @@ this.perk_rf_offhand_training <- ::inherit("scripts/skills/skill", {
 
 	function onPayForItemAction(_skill, _items)
 	{
-		if (_skill != null && _skill.getID() != "perk.rf_target_practice")
-		{
-			this.m.IsSpent = true;
-		}
+		this.m.IsSpent = true;
```
Darxo, line 107: This has nothing to do with this PR and should be moved to another PR

Darxo (changes requested):
The skill implementation itself looks good.
Just two minor things around this PR

scripts/skills/perks/perk_rf_target_practice.nut
LordMidas: I avoided doing these changes in general in this perk rework so as not to muddle the actual perk changes with extra lines of just refactoring the tooltip. I'd rather do these changes separately in a single commit in the entire codebase.

scripts/skills/perks/perk_rf_offhand_training.nut
LordMidas: This perk used to be able to allow swapping ammo for free. This was then accounted for in `rf_offhand_training` where if you are swapping ammo with this perk then that perk didn't become "spent". Now that we removed this ability from the target practice perk, we have to remove that accounting for it in the other perk too. So it is part of this change.

Darxo (approved):

Uberbagel:
Tested. Working. Log clean.

PR #323, opened 2024-03-25
refactor: rework discovered talent to give random stars every levelup

LordMidas:


scripts/skills/perks/perk_rf_discovered_talent.nut
```diff
@@ -1,6 +1,6 @@
 this.perk_rf_discovered_talent <- ::inherit("scripts/skills/skill", {
 	m = {
-		IsApplied = false
+		AttributesRolled = []
```
Darxo, line 3: This could also be just a U8 byte where each bit position from 0 - 7 represents, whether the corresponding attribute from `::Const.Attributes` was already picked.

That will 
- simplify the serialization code
- probably improve performance slightly

But might complicate 
- inclusion of additional attributes in the future
- readability of the non-serialized code

Have you considered this option?

Darxo, line 3: Is the order in this array relevant?
You could prefill it with `false` and have instead the index in the array represent an attribute from `::Const.Attributes`. That should simplify the code overall.

scripts/skills/perks/perk_rf_discovered_talent.nut
```diff
@@ -16,66 +16,53 @@ this.perk_rf_discovered_talent <- ::inherit("scripts/skills/skill", {
 		this.m.IsRefundable = false;
 	}
 
-	function onAdded()
+	// Is called from player.setAttributeLevelUpValues
+	function addStars()
```
Darxo, line 20: You never call this function.

I tested this perk ingame briefly and it doesn't work, which is weird because you marked it as working.
image

Darxo (changes requested):

scripts/skills/perks/perk_rf_discovered_talent.nut
LordMidas: Did you pull the branch correctly before testing? The function is called from the hook on player.nut.

scripts/skills/perks/perk_rf_discovered_talent.nut
LordMidas: That is an intriguing suggestion indeed. I think the performance concern is negligible here. But I think the drawback of not being able to work out of the box with additional attributes added by other mods is a big one. So I would prefer an approach that is not limited by that.

scripts/skills/perks/perk_rf_discovered_talent.nut
Darxo: I have overlooked the changes in the player.nut, they were actually present on the branch I compiled and had locally.
I actually didn't know what I was even needing to do ingame to test the perk so I gave up after forcing a level-up on a char with that perk.

Darxo (approved):

PR #324, opened 2024-04-01
feat: add two handed falchion weapon

LordMidas:


Uberbagel (approved):

Darxo (approved):
I think there are too many types of swords in Reforged with hardly enough different roles for them to fill. I wish there was more effort/research/energy put into making sure that this two handed sword/cleaver hybrid really needs to be in the game and what purpose its existence serves. The same goes for Kriegsmesser

Uberbagel:
I generally feel that more unique weapons are a big positive for variety and player creativity. Just about if not all weapons we've added are unique amongst the lineup. I do not feel like we dole out these new weapons willy nilly or make them unbalanced compared to other weapons, so they're really just another option for players and enemies to have. More variety, different challenges etc...

Uberbagel (changes requested):
Everything works but it is missing the +5% chance to hit the head.

Uberbagel (approved):

Uberbagel:
We decided not to add the +5% chance to hit the head so everything has been tested and this is good to merge.

PR #325, opened 2024-04-02
Refactor weapons rebalance

LordMidas:
See each individual commit message for the changes.

### DIFF
--- a/mod_reforged/hooks/items/weapons/barbarians/rusty_warblade.nut
+++ b/mod_reforged/hooks/items/weapons/barbarians/rusty_warblade.nut
@@ -3,21 +3,25 @@
 	{
 		__original();
 		this.m.Reach = 5;
-		this.m.ShieldDamage = 22;
+		this.m.ShieldDamage = 26;
+		this.m.ArmorDamageMult = 1.3;
+		this.m.RegularDamage = 70;
+		this.m.RegularDamageMax = 90;
 	}
 
 	q.onEquip = @() function()
 	{
 		this.weapon.onEquip();
 
-		this.addSkill(::MSU.new("scripts/skills/actives/cleave", function(o) {
-			o.m.FatigueCost += 3;
+		this.addSkill(::MSU.new("scripts/skills/actives/rf_great_cleave_skill", function(o) {
 			o.m.Icon = "skills/active_182.png";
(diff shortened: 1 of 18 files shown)


mod_reforged/hooks/items/weapons/barbarians/rusty_warblade.nut
Uberbagel, line 1: The sheet says 1.1 for Armor Damage Mult. and -18 for stamina as opposed to using the default vanilla value of -16.

scripts/items/weapons/rf_kriegsmesser.nut
Uberbagel, line 1: Missing 5% chance to hit the head?

scripts/items/weapons/rf_reinforced_wooden_poleflail.nut
Uberbagel, line 1: The skills should have -2 and -5 fat cost respectively

Uberbagel (changes requested):

scripts/items/weapons/rf_kriegsmesser.nut
Uberbagel: We decided not to add this to the Kriegsmesser

Darxo (commented):
I have no idea if the changes were decided this way. I think this should only be reviewed by bagel

scripts/items/weapons/rf_estoc.nut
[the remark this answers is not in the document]
LordMidas: Now that I think about it maybe we left it off from the estoc to give it another dimension which the regular fencing sword didn't have, because the regular one can attack 3 times with the Fencer perk but the Estoc cannot. So perhaps we made them different like that.

In the rework we have made estoc be higher damage. But I still don't know if the -25% chance to hit the head should be added or not. What do you think?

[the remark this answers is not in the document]
Uberbagel: We didn't leave it off of Estoc. I was just questioning if we needed it. I do think Estoc should keep -25% chance to hit the head.

mod_reforged/hooks/items/weapons/barbarians/rusty_warblade.nut
LordMidas: We tested it and found out that it's meant to be 1.3 so that is correct.

PR #329, opened 2024-04-03
feat: add combo perk

LordMidas:


scripts/skills/perks/perk_rf_combo.nut
```diff
@@ -0,0 +1,81 @@
+this.perk_rf_combo <- ::inherit("scripts/skills/skill", {
+	m = {
+		UsedSkillID = "",
+		UsedSkillName = ""
+	},
+	function create()
+	{
+		this.m.ID = "perk.rf_combo";
+		this.m.Name = ::Const.Strings.PerkName.RF_Combo;
+		this.m.Description = ::Const.Strings.PerkDescription.RF_Combo;
```
Darxo, line 10: The perk description is not suited to be the header description of this perk in effect form when viewing the tooltip

scripts/skills/perks/perk_rf_combo.nut
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	function onWaitTurn()
+	{
+		this.m.UsedSkillID = "";
+	}
+
+	function onTurnStart()
+	{
+		this.m.UsedSkillID = "";
```
Darxo, line 72: You do not reset `UsedSkillID` onTurnEnd (as in, this skill is not deactivated then)
That will cause confusion:
- you wonder how long this effect is active after it has not vanished from a character.
- you might be confused by the AP cost of enemy attacks when hovering over them after they ended their turn.

scripts/skills/perks/perk_rf_combo.nut
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	function onCombatFinished()
+	{
+		this.skill.onCombatFinished();
+		this.m.UsedSkillID = "";
+		this.m.UsedSkillName = "";
```
Darxo, line 79: It is redundant to reset this variable.

Darxo (changes requested):

Darxo (approved):

LordMidas:
I just realized that one thing missing from this perk is the affordability preview modification. And this is one of those cases where my [improved affordability preview](https://github.com/MSUTeam/MSU/pull/346) implementation over at MSU will be required because we need to check for the `ActionPointCost` of those skills to be above 3 before this skill modifies them.

Darxo (approved):

Darxo (approved):

PR #331, opened 2024-04-03
Feat dismemberment perk

LordMidas:
Fully tested and working.


scripts/skills/perks/perk_rf_dismemberment.nut
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	function onTargetHit( _skill, _targetEntity, _bodyPart, _damageInflictedHitpoints, _damageInflictedArmor )
+	{
+		if (!_targetEntity.isAlive() || _targetEntity.getMoraleState() == ::Const.MoraleState.Ignore || _targetEntity.getMoraleState() == ::Const.MoraleState.Fleeing || !this.isSkillValid(_skill))
+			return;
+
+		if (this.m.NumInjuriesBefore < _targetEntity.getSkills().getAllSkillsOfType(::Const.SkillType.TemporaryInjury).len())
+		{
+			for (local i = this.getContainer().getActor().getCurrentProperties().getBravery(); i >= 0; i-=25)
```
Darxo, line 53: spaces before `-=` and after it would look nicer and more in line with other operations we do.

scripts/skills/perks/perk_rf_dismemberment.nut
```diff
@@ -0,0 +1,71 @@
...
+		this.m.Order = ::Const.SkillOrder.Perk;
+		this.m.IsActive = false;
+		this.m.IsStacking = false;
+		this.m.IsHidden = false;
+	}
+
+	// Get all the possible injuries, filter out any that are already present on the target or are excluded from the target
+	// Then adjust the InjuryThresholdMult so that the highest injury's threshold ends up being that of the lowest possible injury in the original injury pool
+	// This ensures that the injury inflicted is always one of the highest possible ones
+	function onBeforeTargetHit( _skill, _targetEntity, _hitInfo )
```
Darxo, line 22: I don't understand what's going on in this function even after reading the comments and the perk description.
Got any examples on how this is supposed to function? And what it's wanting to achieve?

scripts/skills/perks/perk_rf_dismemberment.nut
```diff
@@ -0,0 +1,71 @@
+this.perk_rf_dismemberment <- ::inherit("scripts/skills/skill", {
+	m = {
+		IsForceEnabled = false,
```
Darxo, line 3: A comment here would be nice to summarize what exactly IsForceEnabled does for *this* perk.

Like `// ignore axe weapon requirement` or just `// ignore weapon requirement`

scripts/skills/perks/perk_rf_dismemberment.nut
LordMidas: - Injuries A, B, C have a threshold of 25%. This means if you do 25% damage you will inflict one of these injuries.
- Injuries D, E, F have a threshold of 50%. This means if you do 50% damage you will inflict A, B, C, D, E, or F injury.
- What this perk does is that it changes the `InjuryThresholdMult` by 25 / 50 = 0.5 times and then removes all the 25% injuries from the pool. This makes the effective injury threshold of D, E, F as 50 * 0.5 = 25%. So as long as you did enough damage to inflict the lowest possible injury from the pool, you will inflict one of the highest possible injuries.

Darxo: In that case I find a clarification in the perk description in order.
I though that it was just smart injuries which cull the lowest threshold, once you entered the next one.
But what you do here is upgrade any inflicted injury into the highest possible one.

`Injuries inflicted with attacks are upgraded into one with the highest injury threshold.`
or
`Whenever you would inflict an injury with an attack, instead inflict an applicable injury with the highest injury`threshold.´

LordMidas: That's a good point. How about `When inflicting injuries, always inflict one with the highest damage threshold, even if your damage did not meet that threshold.`?

scripts/skills/perks/perk_rf_dismemberment.nut
LordMidas: I've seen that in loop declarations usually people like to keep it condensed like this. But there are many people who write expanded too. I don't mind either way but have a small preference for condensed. I'll ask Enduriel and change it if he also prefers it in your suggested style.

scripts/skills/perks/perk_rf_dismemberment.nut
Darxo: That's fine by me

LordMidas:
@Darxo When you re-review, please look at the injury threshold nested tooltip concept as well and lemme know if it looks good.

mod_reforged/hooks/config/strings.nut
```diff
@@ -1021,6 +1022,16 @@ foreach (vanillaDesc in vanillaDescriptions)
 			]
 		}]
  	}),
+ 	RF_Dismemberment = ::UPD.getDescription({
+ 		Fluff = "Welcome to the chopping block!",
+ 		Requirement = "Axe",
+ 		Effects = [{
+			Type = ::UPD.EffectType.Passive,
+			Description = [
+				"When inflicting [injuries|Concept.InjuryTemporary] with an attack, always inflict one with the highest [damage threshold|Concept.InjuryThreshold] even if the damage did not meet that [threshold|Concept.InjuryThreshold]."
```
Darxo, line 1031: You force-deleted the second effect of this perk; the one about triggering morale checks on the enemy.

scripts/skills/perks/perk_rf_dismemberment.nut
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	function onTargetHit( _skill, _targetEntity, _bodyPart, _damageInflictedHitpoints, _damageInflictedArmor )
+	{
+		if (!_targetEntity.isAlive() || _targetEntity.getMoraleState() == ::Const.MoraleState.Ignore || _targetEntity.getMoraleState() == ::Const.MoraleState.Fleeing || !this.isSkillValid(_skill))
+			return;
+
+		if (this.m.NumInjuriesBefore < _targetEntity.getSkills().getAllSkillsOfType(::Const.SkillType.TemporaryInjury).len())
+		{
+			for (local i = this.getContainer().getActor().getCurrentProperties().getBravery(); i >= 0; i -= 25)
```
Darxo, line 53: I think this will cause 1 more morale check than intended.
Right now a brother with 0 Resolve will trigger 1 check on the enemy. One with 26 Resolve will trigger 2 checks on the enemy.

A solution would be to subtract 25 one time from the starting value.
`for (local i = this.getContainer().getActor().getCurrentProperties().getBravery() - 25; i >= 0; i -= 25)`

Darxo (changes requested):

scripts/skills/perks/perk_rf_dismemberment.nut
```diff
@@ -1,7 +1,8 @@
 this.perk_rf_dismemberment <- ::inherit("scripts/skills/skill", {
 	m = {
 		IsForceEnabled = false, // true ignores weapon requirement
-		NumInjuriesBefore = 0
+		BraveryPerMoraleCheck = 25
```
Darxo, line 4: a colon is missing

Darxo (approved):
Good to go after you added that colon

PR #333, opened 2024-04-03
refactor: rework throwing mastery and hybridization

LordMidas:
- Move melee skill based hit chance bonus from throwing mastery to Hybridization
- Move damage type based effect application from Hybridization to Throwing Mastery
- Remove increased damage to shields from Throwing Spear from Hybridization

### DIFF
--- a/scripts/skills/perks/perk_rf_hybridization.nut
+++ b/scripts/skills/perks/perk_rf_hybridization.nut
@@ -57,6 +57,9 @@ this.perk_rf_hybridization <- ::inherit("scripts/skills/skill", {
 
 	function getItemActionCost( _items )
 	{
+		if (this.m.IsSpent)
+			return null;
+
 		foreach (item in _items)
 		{
 			if (item != null && item.isItemType(::Const.Items.ItemType.Weapon) && item.isWeaponType(::Const.Items.WeaponType.Throwing))
--- a/scripts/skills/perks/perk_rf_hybridization.nut
+++ b/scripts/skills/perks/perk_rf_hybridization.nut
@@ -16,15 +16,6 @@ this.perk_rf_hybridization <- ::inherit("scripts/skills/skill", {
 		this.m.IsHidden = false;
 	}
 
-	function isEnabled()
-	{
-		if (this.getContainer().getActor().isDisarmed())
-			return false;
-
-		local weapon = this.getContainer().getActor().getMainhandItem();
-		return weapon != null && weapon.isWeaponType(::Const.Items.WeaponType.Throwing);
-	}
-
 	function onUpdate( _properties )
 	{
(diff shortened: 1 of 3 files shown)


scripts/skills/perks/perk_rf_hybridization.nut
```diff
@@ -23,80 +32,23 @@ this.perk_rf_hybridization <- ::inherit("scripts/skills/skill", {
...
-				{
-					::Tactical.EventLog.log(::Const.UI.getColorizedEntityName(actor) + " has impaled " + ::Const.UI.getColorizedEntityName(_targetEntity) + " for " + effect.m.TurnsLeft + " turns");
-				}
-			}
-		}
-		else if (_skill.getDamageType().contains(::Const.Damage.DamageType.Cutting))
+		if (_targetEntity != null && _skill.m.IsWeaponSkill && this.isEnabled())
 		{
-			_targetEntity.getSkills().add(::new("scripts/skills/effects/overwhelmed_effect"));
+			_properties.RangedSkill += ::Math.floor(this.getContainer().getActor().getCurrentProperties().getMeleeSkill() * 0.2);
```
Darxo, line 39: The bonus here should be calculated in a separate `getHitchanceBonus()` function.

Darxo, line 39: The `0.2` multiplier should be made into a member variable

Darxo (changes requested):

Darxo (approved):

Uberbagel (changes requested):
Hybridization is supposed to allow swapping to or from a throwing weapon to cost 0 AP once per turn. This is not accounted for in the code or tooltips at this time.

scripts/skills/perks/perk_rf_hybridization.nut
```diff
@@ -31,6 +55,31 @@ this.perk_rf_hybridization <- ::inherit("scripts/skills/skill", {
 		}
 	}
 
+	function getItemActionCost( _items )
+	{
+		foreach (item in _items)
```
Darxo, line 60: You need to check whether `this.m.IsSpent` is `true` and then return early with null.
Otherwise you can swap endlessly in a turn.

scripts/skills/perks/perk_rf_hybridization.nut
```diff
@@ -1,21 +1,45 @@
 this.perk_rf_hybridization <- ::inherit("scripts/skills/skill", {
 	m = {
+		IsSpent = false,
```
Darxo, line 3: If you set this to `true` whenever a combat ends then you probably also want it `true` by default, so this perk is never visible outside of battle.

Darxo (changes requested):

PR #334, opened 2024-04-03
refactor: rework ghostlike and remove sneak attack perk

LordMidas:
The melee part of sneak attack is merged into ghostlike


scripts/skills/perks/perk_rf_ghostlike.nut
```diff
@@ -13,24 +18,73 @@ this.perk_rf_ghostlike <- ::inherit("scripts/skills/skill", {
...
 	}
 
-	function onUpdate( _properties )
+	function isEnabled()
 	{
-		this.m.IsHidden = true;
+		if (this.m.IsForceEnabled)
+			return true;
+
+		if (this.getContainer().getActor().getItems().getStaminaModifier([::Const.ItemSlot.Body, ::Const.ItemSlot.Head]) >= this.m.ArmorStaminaModifier)
```
Darxo, line 79: You are comparing negative numbers here so you need to invert the comparison operator to get the result you want.

At this point I propose changing the member variable `this.m.ArmorStaminaModifier` into `this.m.MaximumWeight` and making it into a positive number.
And then here you can write
```
local currentWeight = -1 * this.getContainer().getActor().getItems().getStaminaModifier([::Const.ItemSlot.Body, ::Const.ItemSlot.Head]);
if (currentWeight > this.m.MaximumWeight)
	return false;
```
 That will improve readability/parsability; at least I find working with positive weight numbers more intuitive than negative staminamodifier numbers.

scripts/skills/perks/perk_rf_ghostlike.nut
```diff
@@ -13,24 +18,73 @@ this.perk_rf_ghostlike <- ::inherit("scripts/skills/skill", {
...
-			icon = "ui/icons/special.png",
-			text = ::Reforged.Mod.Tooltips.parseString("The next movement will ignore [Zone of Control|Concept.ZoneOfControl]")
-		});
+		if (this.m.Enemies.len() != 0)
+		{
+			ret.push({
+				id = 6,
+				type = "text",
+				icon = "ui/icons/special.png",
+				text = ::MSU.Text.colorGreen("25%") + " increased damage and " + ::MSU.Text.colorGreen("+20%") + " damage ignoring armor against:"
```
Darxo, line 46: These multipliers/fractions should be member variables as they are used multiple times and are likely to be tweaked over time.

scripts/skills/perks/perk_rf_ghostlike.nut
```diff
@@ -51,8 +105,84 @@ this.perk_rf_ghostlike <- ::inherit("scripts/skills/skill", {
...
+
+	function onAnySkillUsed( _skill, _targetEntity, _properties )
+	{
+		if (_targetEntity == null || !this.isEnabled())
+			return;
+
+		if (this.m.Enemies.find(_targetEntity.getID()) != null)
+		{
+			_properties.DamageDirectAdd += 0.2;
+			_properties.DamageTotalMult *= 1.25;
```
Darxo, line 136: See Above:

> These multipliers/fractions should be member variables as they are used multiple times and are likely to be tweaked over time.

mod_reforged/hooks/config/strings.nut
```diff
@@ -1230,10 +1229,12 @@ foreach (vanillaDesc in vanillaDescriptions)
  	}),
  	RF_Ghostlike = ::UPD.getDescription({
  		Fluff = "Blink and you\'ll miss me.",
+ 		Requirement = "Weapon with up to 4 Reach and Head and Body Armor weight up to 20"
```
Darxo, line 1232: Are both parts of this effect supposed to be usable with ranged weapons?
Those weapons are considered to have 0 Reach.

E.g. I could use a loaded crossbows, move into melee range of an opponent, unload it into that enemy while profiting from the damage bonus and then move away again.

scripts/skills/perks/perk_rf_ghostlike.nut
```diff
@@ -13,24 +18,73 @@ this.perk_rf_ghostlike <- ::inherit("scripts/skills/skill", {
...
+				ret.push({
+					id = 10,
+					type = "text",
+					icon = "ui/orientation/" + enemy.getOverlayImage() + ".png",
+					text = enemy.getName()
+				});
+			}
+
+			ret.push({
+				id = 10,
```
Darxo, line 64: This needs a different ID from the one used in the for loop.

Darxo (changes requested):

mod_reforged/hooks/config/strings.nut
LordMidas: Melee only. I'll update the description to be clear.

scripts/skills/perks/perk_rf_ghostlike.nut
```diff
@@ -51,8 +107,84 @@ this.perk_rf_ghostlike <- ::inherit("scripts/skills/skill", {
 			if (numAllies >= numEnemies)
 			{
 				_properties.IsImmuneToZoneOfControl = true;
```
Darxo, line 109: This part of the effect does not check whether the global conditions of this perk are active.

scripts/skills/perks/perk_rf_ghostlike.nut
```diff
@@ -13,24 +20,73 @@ this.perk_rf_ghostlike <- ::inherit("scripts/skills/skill", {
...
+		}
+
+		return ret;
 	}
 
-	function onUpdate( _properties )
+	function isEnabled()
 	{
-		this.m.IsHidden = true;
+		if (this.m.IsForceEnabled)
```
Darxo, line 78: You don't check whether the weapon equipped is a melee weapon. There should be a check for that in front of this condition because IsForceEnabled does not turn that off.

Currently Ranged weapons (because they have reach 0) will also work with this perk

Darxo (changes requested):

scripts/skills/perks/perk_rf_ghostlike.nut
LordMidas: In the new commit I've changed the design a bit to make the perk simpler and removed `IsForceEnabled`. It now requires up to 20 armor weight for the ignore zoc part, and requires melee weapon with <=4 reach for the damage part.

Darxo (approved):

PR #337, opened 2024-04-03
refactor: rework concussive strikes

LordMidas:
- Require mace and blunt damage
- 1h maces apply dazed and if already dazed then stun
- 2h maces apply stun

### DIFF
--- a/mod_reforged/hooks/config/strings.nut
+++ b/mod_reforged/hooks/config/strings.nut
@@ -935,11 +935,12 @@ foreach (vanillaDesc in vanillaDescriptions)
  	}),
 	RF_ConcussiveStrikes = ::UPD.getDescription({
  		Fluff = "A strike to the head from this character means goodnight!",
+ 		Requirement = "Mace and Blunt Damage"
  		Effects = [{
 			Type = ::UPD.EffectType.Passive,
 			Description = [
-				"Hits to the head with any weapon apply the [Dazed|Skill+dazed_effect] effect for 1 [turn|Concept.Turn]. This duration is increased to 2 [turns|Concept.Turn] for one-handed maces.",
-				"Hits to the head with two-handed maces apply the [Stunned|Skill+stunned_effect] for 1 [turn|Concept.Turn], and if the target is immune to being [stunned|Skill+stunned_effect], apply [Dazed|Skill+dazed_effect] for 1 [turn|Concept.Turn]."
+				"Hits to the head with one-handed maces apply the [Dazed|Skill+dazed_effect] effect for 1 [turn|Concept.Turn] and if the target is already [dazed|Skill+dazed_effect], apply [Stunned|Skill+stunned_effect] for one [turn|Concept.Turn].",
+				"Hits to the head with two-handed maces apply the [Stunned|Skill+stunned_effect] for 1 [turn|Concept.Turn].",
 			]
 		}]
  	}),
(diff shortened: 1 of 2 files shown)


mod_reforged/hooks/config/strings.nut
```diff
@@ -935,11 +935,12 @@ foreach (vanillaDesc in vanillaDescriptions)
  	}),
 	RF_ConcussiveStrikes = ::UPD.getDescription({
  		Fluff = "A strike to the head from this character means goodnight!",
+ 		Requirement = "Mace and Blunt Damage"
```
Darxo, line 938: `Attack` instead of `Damage` seems more fitting as we do not care about the amount or type of damage inflicted in any way. Especially in a world where MSU introduced a system where the same attack can randomly do either one or the other damage type if multiple are present.

mod_reforged/hooks/config/strings.nut
```diff
@@ -935,11 +935,12 @@ foreach (vanillaDesc in vanillaDescriptions)
...
  		Effects = [{
 			Type = ::UPD.EffectType.Passive,
 			Description = [
-				"Hits to the head with any weapon apply the [Dazed|Skill+dazed_effect] effect for 1 [turn|Concept.Turn]. This duration is increased to 2 [turns|Concept.Turn] for one-handed maces.",
-				"Hits to the head with two-handed maces apply the [Stunned|Skill+stunned_effect] for 1 [turn|Concept.Turn], and if the target is immune to being [stunned|Skill+stunned_effect], apply [Dazed|Skill+dazed_effect] for 1 [turn|Concept.Turn]."
+				"Hits to the head with one-handed maces apply the [Dazed|Skill+dazed_effect] effect for 1 [turn|Concept.Turn] and if the target is already [dazed|Skill+dazed_effect], apply [Stunned|Skill+stunned_effect] for one [turn|Concept.Turn].",
+				"Hits to the head with two-handed maces apply the [Stunned|Skill+stunned_effect] for 1 [turn|Concept.Turn].",
```
Darxo, line 943: Is this nerf to two-handed weapons intentional? Now this is a dead perk for two-handed weapons if the enemy is stun immune. Before you would still apply dazed at the very least. 

One-handed weapons on the other hand can inflict dazed against stun-immune enemies just fine as before

Darxo (changes requested):

mod_reforged/hooks/config/strings.nut
LordMidas: The original tooltip was misleading. Two-Handed maces already apply Dazed by default via the Cudgel skill, even against stun immune enemies. However, they won't apply it against Daze immune enemies. 

So nothing has been changed for 2-handed maces in this perk.

However, we could consider buffing it for 2-handed maces so that Daze-immune enemies still get Dazed applied to them?

mod_reforged/hooks/config/strings.nut
LordMidas: We are using "Damage" in all other cases so I went with that. But yes Attack is more accurate for how the perk is coded. I guess we should change this in other cases as well to "Attack".

Darxo (approved):

Uberbagel (commented):
Tested. Log clean and perk works. However, what was our intended interaction with the new mace double-grip bonus? I think if you have concussive strikes and are using 1h mace with ohf, hits to the head should daze and therefore also stun. Currently, it's just dazing.

LordMidas:
> Tested. Log clean and perk works. However, what was our intended interaction with the new mace double-grip bonus? I think if you have concussive strikes and are using 1h mace with ohf, hits to the head should daze and therefore also stun. Currently, it's just dazing.

Should be fixed now after merging #364.

PR #339, opened 2024-04-03
refactor: rework bloodlust

LordMidas:
- Require cleaver
- Gain increased Resolve, Initiative and Fatigue Cost reduction after fatalities

### DIFF
--- a/mod_reforged/hooks/config/strings.nut
+++ b/mod_reforged/hooks/config/strings.nut
@@ -844,12 +844,13 @@ foreach (vanillaDesc in vanillaDescriptions)
  	}),
  	RF_Bloodlust = ::UPD.getDescription({
  		Fluff = "When surrounded by carnage, you feel revitalized and right at home!",
- 		Requirement = "Melee Attack",
+ 		Requirement = "Cleaver",
  		Effects = [{
 			Type = ::UPD.EffectType.Passive,
 			Description = [
-				"During your [turn|Concept.Turn], every successful attack reduces current [Fatigue|Concept.Fatigue] by " + ::MSU.Text.colorGreen("5%") + " per stack of [Bleeding|Skill+bleeding_effect] on the target and increases [Fatigue Recovery|Concept.FatigueRecovery] by " + ::MSU.Text.colorGreen("+1") + " for one [turn|Concept.Turn] per stack of [Bleeding|Skill+bleeding_effect] on the target.",
-				"[Bleeding|Skill+bleeding_effect] inflicted by the attack, or killing a target, also counts towards the bonus."
+				"After inflicting a fatality, gain " + ::MSU.Text.colorGreen("2") + " stacks of Bloodlust, up to a maximum of 2 stacks.",
+				"For each stack [Resolve|Concept.Bravery] and [Initiative|Concept.Initiative] are increased by " + ::MSU.Text.colorGreen("25%") + " and skills build up " + ::MSU.Text.colorGreen("25%") + " less [Fatigue|Concept.Fatigue], stacking [additively|Concept.StackAdditively].",
+				"You lose " + ::MSU.Text.colorRed("1") + " stack every [turn|Concept.Turn]."
 			]
 		}]
  	}),
--- a/scripts/skills/perks/perk_rf_bloodlust.nut
+++ b/scripts/skills/perks/perk_rf_bloodlust.nut
@@ -1,20 +1,20 @@
 this.perk_rf_bloodlust <- ::inherit("scripts/skills/skill", {
 	m = {
-		BleedStacksBeforeAttack = 0,
-		FatigueRecoveryStacks = 0,		
-		FatigueReductionPercentage = 5,
-		ActorFatigue = null,
-		DidHit = false
+		IsForceEnabled = false,
+		Stacks = 0,
+		MaxStacks = 2,
+		StacksPerTrigger = 2,
+		MultPerStack = 0.25
 	},
 	function create()
 	{
 		this.m.ID = "perk.rf_bloodlust";
 		this.m.Name = ::Const.Strings.PerkName.RF_Bloodlust;
-		this.m.Description = "This character gains increased vigor when next to bleeding enemies.";
+		this.m.Description = "This character gains increased vigor when inflicting fatalities.";
 		this.m.Icon = "ui/perks/rf_bloodlust.png";
 		this.m.IconMini = "rf_bloodlust_mini";
 		this.m.Type = ::Const.SkillType.Perk | ::Const.SkillType.StatusEffect;
-		this.m.Order = ::Const.SkillOrder.Perk;
+		this.m.Order = ::Const.SkillOrder.BeforeLast; // So that it runs onAfterUpdate after active skills to properly reduce their fatigue cost
 		this.m.IsActive = false;
 		this.m.IsStacking = false;
 		this.m.IsHidden = false;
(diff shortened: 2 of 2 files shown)


scripts/skills/perks/perk_rf_bloodlust.nut
```diff
@@ -22,82 +20,90 @@ this.perk_rf_bloodlust <- ::inherit("scripts/skills/skill", {
...
 
 	function getTooltip()
 	{
-		local tooltip = this.skill.getTooltip();
-		tooltip.push({
+		local ret = this.skill.getTooltip();
+
+		local mult = 1.0 + (this.m.Stacks * this.m.MultPerStack);
+		ret.push({
+			id = 10,
```
Darxo, line 32: ID's for these different tooltip lines should be unique

Darxo (changes requested):

mod_reforged/hooks/config/strings.nut
```diff
@@ -844,12 +844,12 @@ foreach (vanillaDesc in vanillaDescriptions)
...
  	RF_Bloodlust = ::UPD.getDescription({
  		Fluff = "When surrounded by carnage, you feel revitalized and right at home!",
- 		Requirement = "Melee Attack",
+ 		Requirement = "Cleaver",
  		Effects = [{
 			Type = ::UPD.EffectType.Passive,
 			Description = [
-				"During your [turn|Concept.Turn], every successful attack reduces current [Fatigue|Concept.Fatigue] by " + ::MSU.Text.colorGreen("5%") + " per stack of [Bleeding|Skill+bleeding_effect] on the target and increases [Fatigue Recovery|Concept.FatigueRecovery] by " + ::MSU.Text.colorGreen("+1") + " for one [turn|Concept.Turn] per stack of [Bleeding|Skill+bleeding_effect] on the target.",
-				"[Bleeding|Skill+bleeding_effect] inflicted by the attack, or killing a target, also counts towards the bonus."
+				"After inflicting a fatality, gain " + ::MSU.Text.colorGreen("2") + " stacks of Bloodlust. For each stack [Resolve|Concept.Bravery] and [Initiative|Concept.Initiative] are increased by " + ::MSU.Text.colorGreen("25%") + " and skills build up " + ::MSU.Text.colorGreen("25%") + " less [Fatigue|Concept.Fatigue], stacking [additively|Concept.StackAdditively].",
```
Darxo, line 851: This should say somewhere more explicitly and earlier, that the maximum amount of stacks is 2. That information is currently only in the second half of the second bullet point where it looks lost and comes a bit too late.

I would use three bullet points in total:
- How are stacks gained / Maximum Stacks
- Effect per stack
- How are stacks lost/removed?

Darxo (changes requested):

scripts/skills/perks/perk_rf_bloodlust.nut
```diff
@@ -22,82 +20,90 @@ this.perk_rf_bloodlust <- ::inherit("scripts/skills/skill", {
...
+			type = "text",
+			icon = "ui/icons/initiative.png",
+			text = ::Reforged.Mod.Tooltips.parseString(::MSU.Text.colorizeMult(mult) + " increased [Initiative|Concept.Initiative]")
+		});
+		ret.push({
+			id = 12,
+			type = "text",
 			icon = "ui/icons/fatigue.png",
-			text = "[color=" + ::Const.UI.Color.PositiveValue + "]+" + this.m.FatigueRecoveryStacks + "[/color] Fatigue Recovery on the next turn"
+			text = ::Reforged.Mod.Tooltips.parseString("Skills build up " + ::MSU.Text.colorizeMult(2.0 - mult, {InvertColor = true}) + " less [Fatigue Cost|Concept.Fatigue]")
```
Darxo, line 47: I think `2.0 - mult` is wrong here, same with `InvertColor = true`.
A mult of 0.25 (from 1 stack) reduces fatigue build up by 25%. But this line would describe that reduction as 75% less.

And from my understand it is correctly to have this kind of reduction colored in green, because it is good for the entity that receives this effect. 

I'm ok with keeping it red for weapon masteries for a bit, until we fix it in one go.
But I don't want us to continue that wrong coloring in other places. Finesse is also still wrong, but the reforged perks Skirmisher and Fluid Weapon had this fatigue cost reduction in colorGreen already.

LordMidas: The calculation is correct. `2.0 - 1.25 = 0.75` which in `colorizeMult` will print `25%` in red. But we have invert color so it will print it in green.

Darxo: Ok then it will look correct.
But this still looks overcomplicated, is it necessary this way?

LordMidas: Lol yes you are correct. I now realize that the way I did it was needlessly complicated. I'll change it.

mod_reforged/hooks/config/strings.nut
```diff
@@ -844,12 +844,13 @@ foreach (vanillaDesc in vanillaDescriptions)
...
- 		Requirement = "Melee Attack",
+ 		Requirement = "Cleaver",
  		Effects = [{
 			Type = ::UPD.EffectType.Passive,
 			Description = [
-				"During your [turn|Concept.Turn], every successful attack reduces current [Fatigue|Concept.Fatigue] by " + ::MSU.Text.colorGreen("5%") + " per stack of [Bleeding|Skill+bleeding_effect] on the target and increases [Fatigue Recovery|Concept.FatigueRecovery] by " + ::MSU.Text.colorGreen("+1") + " for one [turn|Concept.Turn] per stack of [Bleeding|Skill+bleeding_effect] on the target.",
-				"[Bleeding|Skill+bleeding_effect] inflicted by the attack, or killing a target, also counts towards the bonus."
+				"After inflicting a fatality, gain " + ::MSU.Text.colorGreen("2") + " stacks of Bloodlust, up to a maximum of 2 stacks.",
+				"For each stack [Resolve|Concept.Bravery] and [Initiative|Concept.Initiative] are increased by " + ::MSU.Text.colorGreen("25%") + " and skills build up " + ::MSU.Text.colorGreen("25%") + " less [Fatigue|Concept.Fatigue], stacking [additively|Concept.StackAdditively].",
+				"You lose " + ::MSU.Text.colorRed("1") + " stack every [turn|Concept.Turn] but a fatality will reset it to " + ::MSU.Text.colorGreen("2") + "."
```
Darxo, line 853: The second part of this bullet point is no longer needed as it is now fully covered by the first bullet point.

Darxo (changes requested):

Darxo (approved):

scripts/skills/perks/perk_rf_bloodlust.nut
```diff
@@ -22,82 +22,90 @@ this.perk_rf_bloodlust <- ::inherit("scripts/skills/skill", {
...
 			return;
 
-		local bleedCount = _targetEntity.isAlive() ? _targetEntity.getSkills().getAllSkillsByID("effects.bleeding").len() : this.m.BleedStacksBeforeAttack;
-
-		this.m.FatigueRecoveryStacks += bleedCount;
-
-		if (this.m.ActorFatigue == null) this.m.ActorFatigue = actor.getFatigue();
-
-		actor.setFatigue(::Math.max(0, this.m.ActorFatigue - this.m.ActorFatigue * (bleedCount * this.m.FatigueReductionPercentage * 0.01)));
+		local mult = this.m.MultPerStack * this.m.Stacks;
```
Darxo, line 73: I would move the +1 from the lines below up in this line. A `mult` imo is something that is already ready to be multiplied with another value and need no further adjustment.

So like this
`local mult = 1.0 + (this.m.MultPerStack * this.m.Stacks);`

Darxo (approved):
With- or without my last comment, this is good to be merged

PR #340, opened 2024-04-03
Rework hammer mastery and add pummel skill

LordMidas:
- Rework Dented Armor to 10% less melee skill and 20% less damage dealt.
- Add new active skill Pummel.
- Hammer Mastery gives Pummel with two-handed hammers.
- Hammer Mastery causes Crush Armor and Destroy Armor to apply the Dented Armor effect.
- Rework Dent Armor perk to require blunt damage only, increase armor required to 250, and clean up code.


mod_reforged/hooks/config/strings.nut
```diff
@@ -495,6 +495,24 @@ local vanillaDescriptions = [
...
+		Key = "SpecHammer",
+		Description = ::UPD.getDescription({
+			Fluff = "Master hammers and fighting against heavily armored opponents.",
+			Requirement = "Hammer",
+	 		Effects = [{
+ 				Type = ::UPD.EffectType.Passive,
+ 				Description = [
+					"Skills build up " + ::MSU.Text.colorRed("25%") + " less [Fatigue|Concept.Fatigue].",
+					"[Destroy Armor|Skill+crush_armor] and [Demolish Armor|Skill+demolish_armor_skill] inflict " + ::MSU.Text.colorGreen("33%") + " more damage against armor and apply the [Dented Armor|Skill+rf_dented_armor_effect] effect.",
+					"When using two-handed hammers gain the [Pummel|Skill+rf_pummel_skill] skill.",
```
Darxo, line 509: This line should be under the effects category `::UPD.EffectType.Active`

mod_reforged/hooks/skills/perks/perk_mastery_hammer.nut
```diff
@@ -0,0 +1,48 @@
+::Reforged.HooksMod.hook("scripts/skills/perks/perk_mastery_hammer", function(q) {
+	q.m.DentArmorSkills <- [
+		"actives.demolish_armor",
+		"actives.crush_armor"
+	];
+	q.m.MinArmorToDent <- 250;
```
Darxo, line 6: How about a different name like `ArmorMaxThresholdForDent`?
The `min` in your current name is a useful information which I want to preserve by using `Threshold`. I argue that the `max` part is more important to understand what this variable does.

scripts/skills/actives/rf_pummel_skill.nut
```diff
@@ -0,0 +1,129 @@
...
+			::Time.scheduleEvent(::TimeUnit.Virtual, 1, this.onPushThrough.bindenv(this), tag);
+		}
+
+		return success;
+	}
+
+	function onPushThrough( _tag )
+	{
+		if (_tag.TargetEntity.isAlive())
+			this.line_breaker.onUse(_tag.User, _tag.TargetTile);
```
Darxo, line 114: Linebreaker builds up 10 Fatigue on the target, this minor effect is not mentioned in the perks/skills description.

scripts/skills/actives/rf_pummel_skill.nut
```diff
@@ -0,0 +1,129 @@
...
+		}
+
+		return ret;
+	}
+
+	function isUsable()
+	{
+		return this.line_breaker.isUsable() && !this.getContainer().getActor().getCurrentProperties().IsRooted && this.getValidAttack() != null;
+	}
+
```
Darxo, line 77: You have no `onVerifyTarget` function here.
Just like in `isUsable` you should also consult `LineBreaker` whether it's usable on a specific target.

Right now you can use Pummel on rooted or knockbackimmune enemies. LineBreaker doesn't double-check for rooted so you are theoretically able to push those enemies around with Pummel, and that doesn't seem intended (if I read the vanilla code correctly).
KnockBackImmune enemies will stop LineBreaker, but you still wasted additional fatigue in that case.

Darxo (changes requested):

scripts/skills/actives/rf_pummel_skill.nut
LordMidas: This is the vanilla `onVerifyTarget` for `line_breaker.nut`. Seems like it already checks for rooted and knockback immune targets.
```squirrel
function onVerifyTarget( _originTile, _targetTile )
{
	return this.skill.onVerifyTarget(_originTile, _targetTile) && !_targetTile.getEntity().getCurrentProperties().IsRooted && !_targetTile.getEntity().getCurrentProperties().IsImmuneToKnockBackAndGrab && this.findTileToKnockBackTo(_originTile, _targetTile) != null;
}
```

scripts/skills/actives/rf_pummel_skill.nut
LordMidas: Should we remove this 10 Fatigue damage from Pummel or keep it and add it to the tooltip?

mod_reforged/hooks/skills/perks/perk_mastery_hammer.nut
LordMidas: I'm open to changing the name but `ArmorMaxThresholdForDent` is confusing in terms of wheter it is the "MaxThreshold" for "Armor" or is it the "Threshold" for "ArmorMax". Putting an underscore there could clarify it e.g. `ArmorMax_ThresholdForDent` or `Min_ArmorMax_ForDent` or `DentArmorRequiredArmorMax`. Thoughts?

Uberbagel: Out of everything proposed, MinArmorToDent seems the best combination of lack of ambiguity and cleanliness.

scripts/skills/actives/rf_pummel_skill.nut
Uberbagel: I think we should keep it for now, it adds some nice flavor to the skill.

scripts/skills/actives/rf_pummel_skill.nut
Darxo: Ah, I overlooked that Pummel inherits from linebreaker. Forget what I said then.

scripts/skills/actives/rf_pummel_skill.nut
```diff
@@ -34,7 +34,7 @@ this.rf_pummel_skill <- ::inherit("scripts/skills/actives/line_breaker", {
 			id = 11,
 			type = "text",
 			icon = "ui/icons/special.png",
-			text = "If the attack is successful, the target is pushed back a tile and you move into their position"
+			text = ::Reforged.Mod.Tooltips.parseString("If the attack is successful, use [Line Breaker|Skill+line_breaker] on the target")
```
Darxo, line 37: How a bout a `for free` at the end?
`... on the target for free")`

Darxo (approved):
With or without my last remark, this PR is good to go

Uberbagel:
I think the third bullet point here should come to the front for clarity.

image

Uberbagel (changes requested):
### Testing and Feedback

**Dent Armor**
When an entity acts before an ally and fixes their dented armor, the entity with the dented armor still only starts with 7AP that round. I don't think this should be the case. When the Dent is fixed before the entity begins their turn.

**Pummel**
I think the third bullet point here should go to the top for clarity.

image

Darxo:
> **Dent Armor** When an entity acts before an ally and fixes their dented armor, the entity with the dented armor still only starts with 7AP that round. I don't think this should be the case. When the Dent is fixed before the entity begins their turn.

This is happening because in Vanilla the Action Points of every entity is recovered at the same time at **round start**.

We could make it so that removing the dented effect will **always** recover Action Points on the target (even if that target is myself) equal to the MaxActionPoints that were removed by the dented effect.
That would mean that **rf_adjust_dented_armor_ally_skill** and **rf_adjust_dented_armor_skill** now recover 2 Action Points on the target/myself when used.
I kind of like that adjustment actually.

LordMidas:
> in Vanilla the Action Points of every entity is recovered at the same time at round start.

I can see in the `actor.onTurnStart` that the Action Points of an actor are properly recovered up to the max every turn start. So I am surprised that if the Dented Armor effect is removed before the turn started, why is the character still starting their turn at 7 AP. @Uberbagel Are you sure that it wasn't the case that the dented guy had already started his turn with dented armor, then waited, then you removed the dented armor, and then when he resumed his turn you thought that he is starting?

scripts/skills/actives/rf_pummel_skill.nut
LordMidas: I'll change it to `If the attack is successful, automatically use [Line Breaker|Skill+line_breaker] for free on the target`.

LordMidas:
In any case I just realized that we had decided in the new design to change Dented Armor to instead reduce Melee Skill by 10% and Damage by 20%. So I'll make that change and we can forget about the AP for now :P

scripts/skills/actives/rf_pummel_skill.nut
```diff
@@ -122,8 +127,11 @@ this.rf_pummel_skill <- ::inherit("scripts/skills/actives/line_breaker", {
 		if (aoo == null)
 			return null;
 
+		if (this.m.RequiredWeaponType == null)
+			return aoo;
+
 		local weapon = aoo.getItem();
-		if (!::MSU.isNull(weapon) && weapon.isItemType(::Const.Items.ItemType.TwoHanded) && weapon.isItemType(::Const.Items.ItemType.Weapon) && weapon.isWeaponType(::Const.Items.WeaponType.Hammer))
+		if (!::MSU.isNull(weapon) && weapon.isItemType(::Const.Items.ItemType.Weapon) && weapon.isWeaponType(::Const.Items.WeaponType.Hammer))
```
Darxo, line 134: You could use the member variable `m.RequiredWeaponType` here as it also contains `Hammer`

Darxo (changes requested):

Darxo (approved):

Uberbagel (changes requested):
Testing:
I had a freeze after killing an entity who had Dent Armor applied. This was done in Combat Simulator. The log is clean.

I'M DUMB THE BATTLE HAD ENDED BECAUSE THAT WAS THE LAST ENEMY UNIT AGGGH

Uberbagel (approved):
Tested. Works. Log Clean.

PR #344, opened 2024-04-04
refactor: prevent nachzehrer from using swallow whole while netted

Darxo:
I brought up this idea a long time ago and I faintly remember that it was liked back then.

This change will give the player more tools to counter the swallow whole skill from nachzehrer. This will currently make those fights slightly easier but we can always adjust their power in other ways if this turns out problematic.


Uberbagel:
I still support this change.

LordMidas:
> I still support this change.

I think net stopping them from using it completely is not good for nachzehrer fights. Being swallowed is a key component of the challenge in those fights. Nachzehrers **can** be stunned, so there's an avenue already available to stop them from swallowing you via that. Nets make it too easy. If we really have to do something with nets for nachzehrers I'd suggest that netted nachzehrers now require hit-chance to use the swallow skill rather than it being an auto-hit.

PR #345, opened 2024-04-11
refactor: rework en garde perk

LordMidas:
- Allow Double Grip bonus for southern swords with offhand item with less than 10 weight.
- Allow triggering even after movement but in this case build 15 Fatigue.
- Fix: prevent en garde from triggering when disarmed.

### DIFF
--- a/mod_reforged/hooks/config/strings.nut
+++ b/mod_reforged/hooks/config/strings.nut
@@ -1055,9 +1055,10 @@ foreach (vanillaDesc in vanillaDescriptions)
  		Effects = [{
 			Type = ::UPD.EffectType.Passive,
 			Description = [
-				"When wielding a sword, if you have not moved from your position during your [turn|Concept.Turn], use [Riposte|Skill+riposte] freely at the end of your [turn|Concept.Turn] if your weapon has [Riposte|Skill+riposte].",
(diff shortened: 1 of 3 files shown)


mod_reforged/hooks/config/strings.nut
```diff
@@ -1054,9 +1054,10 @@ foreach (vanillaDesc in vanillaDescriptions)
...
+				"The double grip bonus for southern swords is not disabled when using an offhand item with less than " + ::MSU.Text.colorRed("10") + " weight."
+				"Use [Riposte|Skill+riposte] freely at the end of your [turn|Concept.Turn] if your weapon has [Riposte|Skill+riposte]. If your weapon does not have [Riposte|Skill+riposte] and is two-handed, gain the [Rebuke|Skill+rf_rebuke_effect] effect instead for one [turn|Concept.Turn], with an additional chance of " + ::MSU.Text.colorGreen("+15%") + " for returning a missed attack. The attacks triggered from this [Rebuke|Skill+rf_rebuke_effect] do not build [Fatigue|Concept.Fatigue].",
+				"This perk only triggers if you have at least " + ::MSU.Text.colorRed(15) + " [Fatigue|Concept.Fatigue] remaining.",
+				"Does not cost any [Action Points|Concept.ActionPoints]. Additionally, does not build any [Fatigue|Concept.Fatigue] if you have not moved from your position during your [turn|Concept.Turn], otherwise builds " + ::MSU.Text.colorRed(15) + " [Fatigue|Concept.Fatigue]."
```
Darxo, line 1060: This whole description is pretty complicated; there is a lot going on. Take this just as observation/feedback. No need to redesign the perk at this point.

Darxo (approved):

PR #346, opened 2024-04-12
Feat bleed effect rework

LordMidas:
- Bleeding does not expire during battle. Stacks within the same effect instead of multiple instances. Applies 1 damage per turn per stack. Reduces Initiative and Resolve by 5% per stack up to maximum of 50%.
- Bleeder trait doubles the damage received from bleeding instead of the vanilla effect of longer bleeding.
- Adjust cleave skill tooltip for reworked bleed effect

### DIFF
--- a/mod_reforged/hooks/skills/effects/bleeding_effect.nut
+++ b/mod_reforged/hooks/skills/effects/bleeding_effect.nut
@@ -1,0 +1,134 @@
+::Reforged.HooksMod.hook("scripts/skills/effects/bleeding_effect", function(q) {
+	q.m.Stacks <- 1;
+	q.m.StacksThisTurn <- 1;
+	q.m.StacksForMoraleCheck <- 3;
+	q.m.MaxMalusMult <- 0.5;
+	q.m.MalusMultPerStack <- 0.05;
+
+	q.create = @(__original) function()
+	{
+		__original();
+		this.m.Description = "This character is bleeding profusely and will lose hitpoints each turn while acting slower and losing courage.";
+		this.m.IsStacking = false;
+	}
+
+	q.getTooltip <- function()
+	{
+		local ret = this.skill.getTooltip();
+		ret.push({
+			id = 10,
+			type = "text",
+			icon = "ui/icons/regular_damage.png",
+			text = ::Reforged.Mod.Tooltips.parseString("Receive " + ::MSU.Text.colorRed(this.getDamage()) + " damage per [turn|Concept.Turn]")
+		});
+		ret.push({
+			id = 11,
+			type = "text",
+			icon = "ui/icons/bravery.png",
+			text = ::Reforged.Mod.Tooltips.parseString(::MSU.Text.colorizeMult(this.getMalusMult()) + " less [Resolve|Concept.Bravery]")
+		});
+		ret.push({
+			id = 12,
+			type = "text",
+			icon = "ui/icons/initiative.png",
+			text = ::Reforged.Mod.Tooltips.parseString(::MSU.Text.colorizeMult(this.getMalusMult()) + " less [Initiative|Concept.Initiative]")
+		});
+		ret.push({
+			id = 13,
+			type = "text",
+			icon = "ui/icons/special.png",
+			text = ::Reforged.Mod.Tooltips.parseString("If at least " + ::MSU.Text.colorRed(this.m.StacksForMoraleCheck) + " stacks are received in a single [turn|Concept.Turn], a negative [morale check|Concept.Morale] is immediately triggered")
+		});
+		return ret;
+	}
+
+	q.getName <- function()
+	{
+		return this.m.Stacks == 1 ? this.m.Name : this.m.Name + " (x" + this.m.Stacks + ")";
+	}
+
+	q.getDescription = @() function()
+	{
+		return this.m.Description;
+	}
+
+	q.getMalusMult <- function()
+	{
+		return 1.0 - ::Math.minf(this.m.MaxMalusMult, this.m.Stacks * this.m.MalusMultPerStack);
+	}
+
+	q.getDamage <- function()
+	{
+		local ret = this.getContainer().hasSkill("trait.bleeder") ? this.m.Stacks * 2 : 1;
+		return this.getContainer().hasSkill("effects.hyena_potion") ? ::Math.floor(ret * 0.5) : ret;
+	}
+
+	// The vanilla file has an empty onUpdate function
+	q.onUpdate = @() function( _properties )
+	{
+		local mult = this.getMalusMult();
+		_properties.InitiativeMult *= mult;
+		_properties.BraveryMult *= mult;
+	}
+
+	// Based on the vanilla function but use this.getDamage() to calculate hitInfo.DamageRegular and use this.m.Overlay for spawnIcon
+	// and get rid of the part that decrements this.m.TurnsLeft and removes the effect
+	// also apply damage in instances smaller than ::Const.Morale.OnHitMinDamage to prevent morale check from this damage
+	q.applyDamage = @() function()
+	{
+		if (this.m.LastRoundApplied != ::Time.getRound())
+		{
+			this.m.LastRoundApplied = ::Time.getRound();
+			local actor = this.getContainer().getActor();
+			this.spawnIcon(this.m.Overlay, actor.getTile());
+
+			// Split the damage instances if damage is greater than 10 as 10 is the maximum damage from single bleeding in vanilla.
+			// Not splitting can potentialy cause unintended morale checks if damage is greater than ::Const.Morale.OnHitMinDamage.
+			local damage = [this.getDamage()];
+			local i = 0;
+			while (i < damage.len() && damage[i] > 10)
+			{
+				damage.push(damage[i] - 10);
+				damage[i] = 10;
+				i++;
+			}
+
+			local hitInfo = clone ::Const.Tactical.HitInfo;
+			hitInfo.DamageDirect = 1.0;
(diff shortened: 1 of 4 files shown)


mod_reforged/hooks/skills/effects/bleeding_effect.nut
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	q.getMalusMult <- function()
+	{
+		return 1.0 - ::Math.minf(this.m.MaxMalusMult, this.m.Stacks * this.m.MalusMultPerStack);
+	}
+
+	q.getDamage <- function()
+	{
+		local ret = this.getContainer().hasSkill("trait.bleeder") ? this.m.Stacks * 2 : 1;
```
Darxo, line 62: If you go with my switcheroo of the MoraleState then this can be implemented cleaner within the `bleeder_trait` by using `onBeforeDamageReceived`.
- You check whether the id of the passed skill is that of this bleeding effect and if it is, 
- you double the damage that was written in the `_hitInfo` or modify some `damageReceived` variable in `_properties`

That way this logic of doubling the damage stays in the effect which causes it (bleeder).

mod_reforged/hooks/skills/effects/bleeding_effect.nut
```diff
@@ -0,0 +1,131 @@
...
+		{
+			if (!actor.isHiddenToPlayer())
+				::Tactical.EventLog.log(::Const.UI.getColorizedEntityName(actor) + " had his bleeding wound quickly close thanks to his unnatural physiology");
+			return;
+		}
+
+		this.m.Stacks++;
+		this.spawnIcon(this.m.Overlay, actor.getTile());
+
+		if (++this.m.StacksThisTurn == this.m.StacksForMoraleCheck)
```
Darxo, line 123: You must compare using `>=` here

mod_reforged/hooks/skills/effects/bleeding_effect.nut
```diff
@@ -0,0 +1,131 @@
...
+			id = 12,
+			type = "text",
+			icon = "ui/icons/initiative.png",
+			text = ::Reforged.Mod.Tooltips.parseString(::MSU.Text.colorizeMult(this.getMalusMult()) + " reduced [Initiative|Concept.Initiative]")
+		});
+		ret.push({
+			id = 13,
+			type = "text",
+			icon = "ui/icons/special.png",
+			text = ::Reforged.Mod.Tooltips.parseString("If at least " + ::MSU.Text.colorRed(this.m.StacksForMoraleCheck) + " stacks are received in a single [turn|Concept.Turn], a negative [morale check|Concept.Morale] is immediately triggered")
```
Darxo, line 40: semicolon missing

Darxo (changes requested):

mod_reforged/hooks/skills/effects/bleeding_effect.nut
LordMidas: We use `==` because we only want to trigger one morale check if `at least 3 bleed stacks were received in a single turn`. `>=` will cause morale checks to trigger on every bleed received starting at 3 and more, that's not intended.

mod_reforged/hooks/skills/effects/bleeding_effect.nut
LordMidas: This one is correct because it's a table.

mod_reforged/hooks/skills/effects/bleeding_effect.nut
```diff
@@ -0,0 +1,131 @@
...
+				hitInfo.DamageRegular = this.getDamage()
+				actor.onDamageReceived(actor, this, hitInfo);
+			}
+		}
+	}
+
+	q.onRefresh <- function()
+	{
+		local actor = this.getContainer().getActor();
+		if (actor.getCurrentProperties().IsResistantToAnyStatuses && ::Math.rand(1, 100) <= 50)
```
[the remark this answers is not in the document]
LordMidas: Yeah I don't like it either, but I'd rather support it for the time being to keep the bleeding effect consistent - it isn't much work. But for the future we'd definitely need to rethink this and maybe rework vanilla code to make anatomists more moddable / compatible with mods. Could potentially be a work for my modular vanilla mod or we could do that within Reforged - we'll see.

mod_reforged/hooks/skills/effects/bleeding_effect.nut
```diff
@@ -0,0 +1,131 @@
...
+	// also apply damage in instances smaller than ::Const.Morale.OnHitMinDamage to prevent morale check from this damage
+	q.applyDamage = @() function()
+	{
+		if (this.m.LastRoundApplied != ::Time.getRound())
+		{
+			this.m.LastRoundApplied = ::Time.getRound();
+			local actor = this.getContainer().getActor();
+			this.spawnIcon(this.m.Overlay, actor.getTile());
+
+			// Split the damage instances if damage is greater than ::Const.Morale.OnHitMinDamage
```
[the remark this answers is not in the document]
LordMidas: Yeah I considered that morale state thing but didn't like it because of potential breakage with things. I'd rather break Nine Lives than change the morale state which may have many knock-on effects on other skills. But I agree this isn't good. But I don't like the morale state solution either. We could try to brainstorm a better solution.

mod_reforged/hooks/skills/effects/bleeding_effect.nut
LordMidas: That's a great suggestion. I really like the idea of compartmentalizing the increased bleed damage to within `bleeder_trait`. But I am not sold on the morale state fixing idea yet - so let's see if we can come up with a good solution for that which can also work with compartmentalizing this damage multiplication into bleeder trait.

mod_reforged/hooks/skills/effects/bleeding_effect.nut
Darxo: Ah, so bleeds are **always** only afflicted one at a time?

LordMidas: I'm not sure I fully understand your question. You can only add one instance of bleed at a time but multiple instances may be added in a single attack due to different perks/skills.

Darxo: Ok, I understand. Then the line makes sense to me

mod_reforged/hooks/skills/effects/bleeding_effect.nut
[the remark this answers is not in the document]
Darxo: How is it an issue if my switcheroo potentially breaks things in the future, but not now.
But it's not an issue that your implementation potentially breaks things in the future (once someone introduces an effect that increases damage taken).

Meanwhile my solution prevents the nine lives edge case aswell as allowing compartmentalizing of the bleeder trait effect and potentially other effects that increase bleed damage taken.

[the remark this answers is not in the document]
LordMidas: > But it's not an issue that your implementation potentially breaks things in the future but not now.

First off, my solution breaks things (i.e. Nine Lives) **now**, not potentially in the future. Secondly, I didn't say it's not an issue, in fact I said "I agree this isn't good".

Your proposed solution changes the morale state which may have completely "surprise" effects with other perks/skills/effects that we later add or that submodders add because no one expects a human's morale state to be `Ignore` in the middle of damage. So I don't like that solution. It can break things in a very opaque and unexpected way. 

Which is why I said let's try to brainstorm a better solution.

mod_reforged/hooks/skills/effects/bleeding_effect.nut
```diff
@@ -0,0 +1,131 @@
...
+			// so that the damage does not trigger morale checks
+			local damage = [this.getDamage()];
+			local i = 0;
+			while (i < damage.len() && damage[i] > ::Const.Morale.OnHitMinDamage)
```
Darxo, line 89: This will cause an infinite loop if a mod sets the OnHitMinDamage to 0

Darxo, line 89: You do this in order to not cause morale hits by bleed damage. But this does not achieve that yet. If the bleeding does 15 damage then that is **not** greater than the vanilla `OnHitMinDamage` because it is equal to that. So the bleeding damage will not be split. And the 15 damage will later cause a morale check.

Once you fixed that, then a `OnHitMinDamage` of 1 will now also cause an infinite loop

Darxo (changes requested):

mod_reforged/hooks/skills/effects/bleeding_effect.nut
```diff
@@ -86,11 +86,10 @@
 			// so that the damage does not trigger morale checks
```
Darxo, line 86: The comment above this line (`::Const.Morale.OnHitMinDamage`) is no longer accurate

mod_reforged/hooks/skills/effects/bleeding_effect.nut
```diff
@@ -0,0 +1,133 @@
...
+
+			local hitInfo = clone ::Const.Tactical.HitInfo;
+			hitInfo.DamageDirect = 1.0;
+			hitInfo.BodyPart = ::Const.BodyPart.Body;
+			hitInfo.BodyDamageMult = 1.0;
+			hitInfo.FatalityChanceMult = 0.0;
+			foreach (d in damage)
+			{
+				if (this.isGarbage())
+					return;
```
Darxo, line 104: You should mention nine lives here in any way as this rule basically just exists because of that perk.

Darxo (approved):

PR #353, opened 2024-04-19
feat: add info in actor tooltip about whether actor has used wait

LordMidas:


Darxo (approved):

LordMidas:
@Darxo I've re-requested review because I've changed the display mode to now add the word `again` so it shows up as `Acts again in x turns` instead of `Acts in x turns (Has Waited)`. I prefer this as it uses less space and is clear enough, especially with the nested tooltip about wait concept. It also leaves more room to add the Reach info there later.

PR #355, opened 2024-04-21
refactor: display icons for bone plating while it is still active

Darxo:
Tested ingame, seems to work

image

image


scripts/skills/effects/rf_bone_plating_effect.nut
```diff
@@ -0,0 +1,34 @@
...
+			"sounds/enemies/skeleton_hurt_02.wav",
+			"sounds/enemies/skeleton_hurt_03.wav",
+			"sounds/enemies/skeleton_hurt_04.wav",
+			"sounds/enemies/skeleton_hurt_06.wav"
+		]
+	},
+
+	function create()
+	{
+		::logWarning("unction create()");
```
LordMidas, line 14: This logWarning should be removed.

scripts/skills/effects/rf_bone_plating_effect.nut
```diff
@@ -0,0 +1,34 @@
+this.rf_bone_plating_effect <- ::inherit("scripts/skills/skill", {
```
LordMidas, line 1: The vanilla item is called `bone_platings_upgrade` so we should follow that naming convention and rename the skill file as well as the images/brush to be plural i.e. `platings`.

scripts/skills/effects/rf_bone_plating_effect.nut
```diff
@@ -0,0 +1,34 @@
...
+	m = {
+		SoundOnProcced = [
+			"sounds/enemies/skeleton_hurt_01.wav",
+			"sounds/enemies/skeleton_hurt_02.wav",
+			"sounds/enemies/skeleton_hurt_03.wav",
+			"sounds/enemies/skeleton_hurt_04.wav",
+			"sounds/enemies/skeleton_hurt_06.wav"
+		]
+	},
```
LordMidas, line 10: Vanilla doesn't put an empty line between `m,` and `function create()` so I suggest we keep that style.

scripts/skills/effects/rf_bone_plating_effect.nut
```diff
@@ -0,0 +1,34 @@
+this.rf_bone_plating_effect <- ::inherit("scripts/skills/skill", {
+	m = {
+		SoundOnProcced = [
```
LordMidas, line 3: 1. Is this a vanilla thing? I don't know for sure but it seems triggering these sounds is a new feature in Reforged. If that is correct then please make this a separate commit.
2. Instead of hard-coding these sounds in a custom array in `m` you can instead use the vanilla `::Const.Sound.ArmorBoneImpact` array directly in the call to `::Sound.play` below in this skill.

mod_reforged/hooks/items/armor/armor.nut
```diff
@@ -0,0 +1,10 @@
+::Reforged.HooksMod.hook("scripts/items/armor/armor", function(q) {
```
LordMidas, line 1: The addition of this vanilla fix should be a separate commit so that it can be reverted separately later if MSU adds this fix.

mod_reforged/hooks/items/armor_upgrades/bone_platings_upgrade.nut
```diff
@@ -0,0 +1,7 @@
+::Reforged.HooksMod.hook("scripts/items/armor_upgrades/bone_platings_upgrade", function(q) {
+	q.onCombatStarted = @() function()
+	{
+		this.m.IsUsed = true;	// (vanilla) Disable the built-in damage reduction from the item
```
LordMidas, line 4: What if a submod wants to add an additional effect to this item and checks for `this.m.IsUsed`? In order to have better submod compatibility with regards to submod compatibility perhaps a better way to do it would be leave `this.m.IsUsed` untouched and instead overwrite the vanilla `onBeforeDamageReceived` function to be empty:
```squirrel
// Overwrite the vanilla function to be empty so that damage reduction isn't applied from this item
q.onBeforeDamageReceived = @() function( _attacker, _skill, _hitInfo, _properties )
{
}
```

LordMidas (changes requested):

scripts/skills/effects/rf_bone_plating_effect.nut
Darxo: Yes it's a new feature in reforged. Though inspired by legends

mod_reforged/hooks/items/armor/armor.nut
Darxo: It was already a separate commit, unless I confuse something

LordMidas (approved):

PR #361, opened 2024-04-25
refactor(balance): adjust dynamic perk weights

Uberbagel:
Requires find and replace for strong and vigorous groups if we change the names of those. 
Also requires code to make use of isMeleeOnly, isMeleeSpecialist, isMeleeBackground and isRangedBackground templates. We can find and replace those keywords depending on how the system gets implemented.

Not calling for review from Darxo at this point until these things are sorted.


LordMidas (approved):
Should be squashed before merging.

PR #367, opened 2024-05-01
feat: add new perk Cheap Trick

Darxo:
Add new skill/effect for cheap trick

### Discussion

- Currently the effects are only granted **during** the attack. So those stats are not visible in the character sheet. 
  - Displaying and applying them always would look much better for the player
  - However this might influence other effects or skills which are not attacks but scale with melee/ranged skill

### Screenshots

image

image

image

image


scripts/skills/effects/rf_cheap_trick_effect.nut
```diff
@@ -0,0 +1,122 @@
...
+				text = ::Reforged.Mod.Tooltips.parseString(::MSU.Text.colorizeValue(this.m.RangedSkillModifier) + " [Ranged Skill|Concept.RangedSkill]"),
+			});
+		}
+
+		if (this.m.DamageMult != 1.0)
+		{
+			ret.push({
+				id = 12,
+				type = "text",
+				icon = "ui/icons/damage_dealt.png",
```
LordMidas, line 47: Vanilla almost always uses `damage_regular.png` for damage, so I'd suggest to use that. The `damage_dealt` is also the same icon we are using for `reach`.

scripts/skills/actives/rf_cheap_trick_skill.nut
```diff
@@ -0,0 +1,96 @@
...
+				text = ::MSU.Text.colorRed(::Reforged.Mod.Tooltips.parseString("Can only be used once per [turn|Concept.Turn]")),
+			});
+		}
+
+		return ret;
+	}
+
+	function isUsable()
+	{
+		return (this.skill.isUsable() && !this.m.WasUsedThisTurn);
```
LordMidas, line 77: Please remove the parantheses.

scripts/skills/actives/rf_cheap_trick_skill.nut
```diff
@@ -0,0 +1,96 @@
+this.rf_cheap_trick_skill <- ::inherit("scripts/skills/skill", {
+	m = {
+		MeleeSkillModifier = 25,
+		RangedSkillModifier = 25,
+		DamageMult = 0.50,
+
+		// Private
+		WasUsedThisTurn = false,
```
LordMidas, line 8: Minor point: in a lot of places we use `IsSpent` for this kind of flag, so for consistency you could use that name.

scripts/skills/actives/rf_cheap_trick_skill.nut
```diff
@@ -0,0 +1,96 @@
...
+	function onUse( _user, _targetTile )
+	{
+		local self = this;
+		this.getContainer().add(::MSU.new("scripts/skills/effects/rf_cheap_trick_effect", function(o) {
+			o.m.MeleeSkillModifier = self.m.MeleeSkillModifier;
+			o.m.RangedSkillModifier = self.m.RangedSkillModifier;
+			o.m.DamageMult = self.m.DamageMult;
+		}));
+
+		this.m.WasUsedThisTurn = true;
```
LordMidas, line 89: `return true;` at the end of this function.

scripts/skills/effects/rf_cheap_trick_effect.nut
```diff
@@ -0,0 +1,122 @@
...
+		MeleeSkillModifier = 0,		// Must be set during initialization
+		RangedSkillModifier = 0,
+		DamageMult = 0.0,
+	},
+	function create()
+	{
+		this.m.ID = "effects.rf_cheap_trick";
+		this.m.Name = "Cheap Trick";
+		this.m.Description = "Unleash a quick and deceptive move to catch opponents off-guard, allowing for a higher chance of landing a hit but at the cost of power.";
+		this.m.Icon = "ui/skills/rf_cheap_trick.png";	// TODO
```
LordMidas, line 13: You'll also need to add `this.m.Overlay` for this so the effect's icon spawns as well.

scripts/skills/effects/rf_cheap_trick_effect.nut
```diff
@@ -0,0 +1,122 @@
...
+			});
+		}
+
+		if (this.getContainer().getActor().isWaitActionSpent() == false)
+		{
+			ret.push({
+				id = 20,
+				type = "text",
+				icon = "ui/icons/warning.png",
+				text = ::Reforged.Mod.Tooltips.parseString("Will expire upon using \'Wait\'")
```
LordMidas, line 58: Missing comma

scripts/skills/effects/rf_cheap_trick_effect.nut
Darxo: Do we really need to show the same icon, once squared (after using the skill) and then circled (after receiving the effect) on the same character? 
That seems redundant to me

LordMidas: Yes because the skill usage shows that the character successfully used the skill and the effect shows that actor got the effect. This is consistent with all similar vanilla skills e.g. Adrenaline, Shieldwall etc. It also future proofs it in the case this effect can be achieved via some other means.

scripts/skills/actives/rf_cheap_trick_skill.nut
```diff
@@ -0,0 +1,96 @@
...
+
+		if (this.m.DamageMult != 1.0)
+		{
+			ret.push({
+				id = 12,
+				type = "text",
+				icon = "ui/icons/damage_dealt.png",
+				text = ::Reforged.Mod.Tooltips.parseString("Deal " + ::MSU.Text.colorizeMult(this.m.DamageMult) + " less damage"),
+			});
+		}
```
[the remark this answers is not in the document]
Darxo: I made a rework that still mentions Melee Skill and Ranged Skill separately.

I rather not use this vague term "chance to hit" unless we introduce a new concept for it that explicitly states that this means MeleeSkill AND RangedSkill and **will** scale with multipliers to those.

scripts/skills/effects/rf_cheap_trick_effect.nut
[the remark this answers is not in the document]
Darxo: I want to keep the condition. This warning is only relevant in this specific situation. 
This is not the vanquisher perk where there are many conditions which we all want to combine in a single line and therefor needing to show it all the time.

scripts/skills/effects/rf_cheap_trick_effect.nut
```diff
@@ -0,0 +1,122 @@
...
+
+		if (this.m.DamageMult != 1.0)
+		{
+			ret.push({
+				id = 12,
+				type = "text",
+				icon = "ui/icons/damage_dealt.png",
+				text = ::Reforged.Mod.Tooltips.parseString("Deal " + ::MSU.Text.colorizeMult(this.m.DamageMult) + " less damage"),
+			});
+		}
```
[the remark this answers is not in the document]
Darxo: Let me know if this is still a problem now as I did some changes

mod_reforged/config/perk_defs.nut
```diff
@@ -15,6 +15,14 @@
 		Icon = "ui/perks/rf_calculated_strikes.png",
 		IconDisabled = "ui/perks/rf_calculated_strikes_sw.png",
 	},
+	{
+		ID = "perk.rf_cheap_trick",
+		Script = "scripts/skills/perks/perk_rf_cheap_trick",
+		Name = ::Const.Strings.PerkName.RF_CheapTrick,
+		Tooltip = ::Const.Strings.PerkDescription.RF_CheapTrick,
+		Icon = "ui/perks/rf_cheap_trick.png",	// TODO
+		IconDisabled = "ui/perks/rf_cheap_trick_sw.png",	// TODO
```
LordMidas, line 24: // TODO can be removed now

scripts/skills/effects/rf_cheap_trick_effect.nut
```diff
@@ -0,0 +1,99 @@
...
+		// Config
+		HitChanceModifier = 0,		// Must be set during initialization
+		DamageRegularMult = 1.0,	// Must be set during initialization
+	},
+	function create()
+	{
+		this.m.ID = "effects.rf_cheap_trick";
+		this.m.Name = "Cheap Trick";
+		this.m.Description = "Unleash a quick and deceptive move to catch opponents off-guard, allowing for a higher chance of landing a hit but at the cost of power.";
+		this.m.Icon = "ui/perks/rf_cheap_trick.png";
```
LordMidas, line 12: Do you not add an IconMini for this?

LordMidas (changes requested):

scripts/skills/actives/rf_cheap_trick_skill.nut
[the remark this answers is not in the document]
LordMidas: I agree that that the `Chance to hit` is badly used in vanilla tooltips because it's also affected by multipliers. But saying that you have +x Melee Skill for the next attack isn't fully clear either. I'm afraid many players will interpret it as you gain +25 Melee Skill and +25 Ranged Skill **until you do another attack**. They will use this skill, and check their bro's stats and open a bug report that see I am not getting the stats.

Perhaps it could be worded like `during` instead of `for` the next attack. I'm not sure if that will solve it fully either.

The hitchance tooltip keeps the behavior clear for the user's side and while I agree that mechanically it is not 100% exact, I don't think that the alternative is a good solution either. We could brainstorm this to come up with a better way of representing this in tooltips. 

Idk, I don't have a perfect solution for this at this moment.

For this build I'd say change to hit chance for now (to be consistent with vanilla skills that do the same). Then in the future when we have an actual solution to this tooltip issue we could go through and update these including the vanilla ones.

scripts/skills/effects/rf_cheap_trick_effect.nut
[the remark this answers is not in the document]
LordMidas: We somehow have to convey that it expires when you end your turn and doesn't carry over into the next turn. The cleanest way to do that, what seemed to me, is what I suggested above.

scripts/skills/effects/rf_cheap_trick_effect.nut
Darxo: I do not and I don't see a reason to do it for this effect.
This effect only lasts during the turn of the active entity, it is always removed once you give control over to the next entity.

scripts/skills/actives/rf_cheap_trick_skill.nut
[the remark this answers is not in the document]
Darxo: what icon do you recommend for a tooltip line that grants "chance to hit"?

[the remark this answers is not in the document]
LordMidas: The same one that vanilla uses: `"ui/icons/hitchance.png`

scripts/skills/effects/rf_cheap_trick_effect.nut
LordMidas: Sure, that's fine.

scripts/skills/effects/rf_cheap_trick_effect.nut
```diff
@@ -0,0 +1,97 @@
...
+				icon = "ui/icons/damage_regular.png",
+				text = ::Reforged.Mod.Tooltips.parseString("Deal " + ::MSU.Text.colorizeMult(this.m.DamageRegularMult ) + " less damage with your next attack"),
+			});
+		}
+
+		ret.push({
+			id = 20,
+			type = "text",
+			icon = "ui/icons/warning.png",
+			text = ::Reforged.Mod.Tooltips.parseString("Will expire upon using \'Wait\' or ending the turn"),
```
LordMidas, line 46: Should be `[turn|Concept.Turn]`

LordMidas (approved):
Looks good. Just a hyperlink in a tooltip is missing.

PR #368, opened 2024-05-01
Feat more named names

Darxo:
This PR was a spontaneous idea of mine when I saw that vanilla only has two famed swordlance names. So I went and added a few names here and there.

It also includes a proposal for a new `::MSU.Array.removeValues` function. The singular removeValue function of MSU is not very performant if we want to remove a list of names as it would iterative several times over the same array.

### DIFF
--- a/mod_reforged/config/item_names.nut
+++ b/mod_reforged/config/item_names.nut
@@ -1,3 +1,69 @@
+// Vanilla Adjustments
+::MSU.Array.removeValues(::Const.Strings.CrossbowNames, [
+	"Betrayer"		// moved over to dagger names
+]);
+::Const.Strings.CrossbowNames.extend([
+	"Scorpion",
+	"Ironsight",
+	"Thunderbolt",
+	"Ravenshrike",
+	"Bloodbolt",
+	"Grimshot",
+	"Wyrmfang",
+	"Blackthorn",
+	"Stormbreaker",
+	"Steelwind"
+]);
+
+::MSU.Array.removeValues(::Const.Strings.DaggerNames, [
+	"Estoc"		// We have our own actual Estoc item
+]);
+::Const.Strings.DaggerNames.extend([
+	"Betrayer"		// formerly a crossbow name
+]);
+
+::Const.Strings.HandgonneNames.extend([
+	"Deafener",
+	"Eisenknall",
+	"Volcano",
+	"Ironspark",
+	"Thundermuzzle",
+	"Stormfire",
+	"Inferno",
+	"Blazeburst"
+]);
+
+::Const.Strings.LongaxeNames.extend([
+	"Culler"
+]);
+
+::Const.Strings.SwordlanceNames.extend([	// In Vanilla this has only 2 entries
+	"Bloodmoon",
+	"Ghoul\'s Bane",
+	"Soulstrike",
+	"Shadowreaper",
+	"Death\'s Hand",
+	"Grimharvest",
+	"Bloodstorm",
+	"Warbringer",
+	"Deathwind"
+]);
+
+::MSU.Array.removeValues(::Const.Strings.WarbrandNames, [
+	"Kriegsmesser",	// We have our own actual Kriegsmesser item
+]);
+
+::Const.Strings.WhipNames.extend([
+	"Slavedriver",
+	"Tendril",
+	"Vine",
+	"Bloodletter",
+	"Spikelash",
+	"Thornwhip",
+	"Bloodlash"
+]);
+
+// Reforged Lists
 ::Const.Strings.RF_KriegsmesserNames <- [
 	"Kriegsmesser",
 	"Ripper",
@@ -58,6 +124,3 @@
 	"Honor",
 	"Vanquisher"
 ];
-
-// Remove Kriegsmesser from Warbrand names because we have our own actual Kriegsmesser item
-::MSU.Array.removeByValue(::Const.Strings.WarbrandNames, "Kriegsmesser");
--- a/mod_reforged/hooks/msu.nut
+++ b/mod_reforged/hooks/msu.nut
@@ -320,4 +320,16 @@
 	}
 });
 
+::logInfo("Reforged::MSU -- adding ::MSU.Array.removeValues");
+::MSU.Array.removeValues <- function( _array, _values )
+{
+	for (local i = _array.len() - 1; i > 0; i--)
+	{
+		if (_values.find(_array[i]) != null)
+		{
+			_array.remove(i);
+		}
+	}
+}
+
 ::logWarning("------ Reforged modifications to MSU Finished------");


LordMidas:
The intended functionality of `::MSU.Array.removeValues` can be done natively via:
```
local newArray = myArray.filter(@(_, a) excludeArray.find(a) != null);
```

mod_reforged/config/item_names.nut
```diff
@@ -1,3 +1,52 @@
+// Vanilla Adjustments
+::MSU.Array.removeValues(::Const.Strings.CrossbowNames, [
+	"Betrayer",		// moved over to dagger names
+]);
+::Const.Strings.CrossbowNames.extend([
+	"Scorpion",
+	"Ironsight",
+	"Windlass",
```
LordMidas, line 8: Windlass doesn't seem appropriate for a crossbow name as it is a tool that's used to crank a crossbow.

mod_reforged/config/item_names.nut
```diff
@@ -1,3 +1,52 @@
...
+]);
+
+::MSU.Array.removeValues(::Const.Strings.DaggerNames, [
+	"Estoc",		// We have our own actual Kriegsmesser item
+]);
```
LordMidas, line 13: This seems like a mistake.

mod_reforged/config/item_names.nut
```diff
@@ -1,3 +1,52 @@
+// Vanilla Adjustments
+::MSU.Array.removeValues(::Const.Strings.CrossbowNames, [
+	"Betrayer",		// moved over to dagger names
+]);
+::Const.Strings.CrossbowNames.extend([
```
LordMidas, line 5: More cool crossbow names:
- Thunderbolt
- Ravenshrike
- Bloodbolt
- Grimshot
- Wyrmfang
- Blackthorn
- Thunderclap
- Stormbreaker
- Steelwind

mod_reforged/config/item_names.nut
```diff
@@ -1,3 +1,52 @@
...
+]);
+
+::MSU.Array.removeValues(::Const.Strings.DaggerNames, [
+	"Estoc",		// We have our own actual Kriegsmesser item
+]);
+::Const.Strings.DaggerNames.extend([
+	"Betrayer",		// formerly a crossbow name
+]);
+
+::Const.Strings.HandgonneNames.extend([
```
LordMidas, line 18: More cool Handgonne names:
- Ironspark
- Thundermuzzle
- Stormfire
- Inferno
- Blazeburst

mod_reforged/config/item_names.nut
```diff
@@ -1,3 +1,52 @@
...
+	"Volcano",
+]);
+
+::Const.Strings.LongaxeNames.extend([
+	"Culler",
+]);
+
+::Const.Strings.SwordlanceNames.extend([	// In Vanilla this has only 2 entries
+	"Bloodmoon",
+	"Ghoul\'s Bane",
```
LordMidas, line 30: I don't think we can use 2 word names because in vanilla it can do things like `Sir Roderick's xyz` and there `Sir Roderick's Ghoul's Bane` sounds weird. I'd stick to single word names.

mod_reforged/config/item_names.nut
```diff
@@ -1,3 +1,52 @@
...
+	"Soulstrike",
+	"Shadowreaper",
+	"Death\'s Hand",
```
LordMidas, line 33: Same as Ghoul's Bane above.

mod_reforged/config/item_names.nut
```diff
@@ -1,3 +1,52 @@
...
+	"Necroblade",
```
LordMidas, line 34: Necroblade sounds like an undead item, so I'd not use it.

mod_reforged/config/item_names.nut
```diff
@@ -1,3 +1,52 @@
...
+	"Deafener",
+	"Eisenknall",
+	"Volcano",
+]);
+
+::Const.Strings.LongaxeNames.extend([
+	"Culler",
+]);
+
+::Const.Strings.SwordlanceNames.extend([	// In Vanilla this has only 2 entries
```
LordMidas, line 28: Other cool swordlance names:
- Bloodstorm
- Warbringer
- Deathwind

mod_reforged/hooks/msu.nut
```diff
@@ -320,4 +320,16 @@
 	}
 });
 
+::logInfo("Reforged::MSU -- adding ::MSU.Array.removeValues");
+::MSU.Array.removeValues <- function( _array, _values )
```
LordMidas, line 324: This can be done natively in squirrel via:
```squirrel
myArray = myArray.filter(@(_, a) excludeArray.find(a) != null);
```
However one can argue that the gain of this function is more convenience and/or streamlined code as one can do:
```squirrel
::MSU.Array.removeValues(myArray, [1, 2, 3, 4]);

// instead of

local exclude = [1, 2, 3, 4];
myArray = myArray.filter(@(_, a) exclude.find(a) != null);
```
I'd ask @Enduriel's opinion on this.

mod_reforged/config/item_names.nut
```diff
@@ -1,3 +1,52 @@
...
+	"Death\'s Hand",
+	"Necroblade",
+	"Grimharvest",
+]);
+
+::MSU.Array.removeValues(::Const.Strings.WarbrandNames, [
+	"Kriegsmesser",	// We have our own actual Kriegsmesser item
+]);
+
+::Const.Strings.WhipNames.extend([
```
LordMidas, line 42: Other cool whip names:
- Spikelash
- Thornwhip
- Bloodlash
- Ironfang
- Ripper
- Warthorn

LordMidas (changes requested):
Nice names Darxo! I've added some more suggestions for cool names.

mod_reforged/config/item_names.nut
LordMidas: Ah I see there's a dagger name Estoc in vanilla dagger names. Ok then the removal is good, but please update the comment to mention Estoc instead of Kriegsmesser.

mod_reforged/config/item_names.nut
Darxo: Vanilla uses 2 word names plenty of times, sometimes even 3 word names.

And for greatswords they even have `"Thug\'s Bane",`
image

mod_reforged/config/item_names.nut
Darxo: Warbringer sounds very generic. I added the other two though

mod_reforged/config/item_names.nut
Darxo: "Thunderclap" feels more fitting for the handgonne so I do not include it

mod_reforged/config/item_names.nut
LordMidas: - Yeah Ironfang isn't the best for a whip I guess.
- Ripper could be `Fleshripper`.
- Warthorn sounds ok to me. What do you find wrong about it?

mod_reforged/config/item_names.nut
LordMidas: I see. In that case I'd not use the word "Ghoul" as it seems out of place for southern faction lore in BB. Maybe something else's bane e.g.:
- Nomad's Bane
- Mortal's Bane

Or remove the name if you don't find any other appropriate replacement for Ghoul.

mod_reforged/config/item_names.nut
Darxo: `War` is very generic
And `Thorn` only makes sense for the barbarian `thorned_whip` which does not have a famed version. I don't see how it fits with normal whips. That term would be better suited for spiked maces, pickaxes, spears, daggers.

LordMidas: I mean to me it sounds like this is a whip which will prick you like a thorn and it's used in war. So it's a war thorn in the sense that it pricks people and makes them bleed. But no big deal you can ignore the name.

Darxo: Fleshripper is already used for Kriegsmesser so I will not add it here

Uberbagel (approved):

mod_reforged/config/item_names.nut
LordMidas: I looked through the vanilla codebase and found that the word `ghoul` is used in the text of the `roaming_beasts_desert_contract` and the word `ghoulish` is used in many other texts. So, `Ghoul\'s Bane` is fine.

mod_reforged/hooks/msu.nut
Enduriel: We discussed this kind of thing internally in the MSU team and we seem to have settled on encouraging stuff like `filter` rather than adding minor functions for things like this, even though that was not the case historically.

Darxo: Is it ok to add the helper function locally just within the `item_names.nut`?
At the end of the day it still saves 1 line of code whenever it's used

Enduriel: Actually now that I think about it I was wrong and these are not equivalent and adding this function makes sense, your function modifies the array in place which may sometimes be relevant. To me this makes sense and should stay as it is.

LordMidas (approved):
I've pushed a commit removing ending commas from the new arrays and removed the comment as it was incorrect. You can squash and merge this PR.

PR #369, opened 2024-05-07
Fix pass correct killer to msu ondeath events

LordMidas:
- Fix the issue where we switch the `_killer` in our hook on actor `onDeath` function causing MSU hook on `onDeath` to receive this wrong killer.
- Clean up the switching of the `_killer` by setting it to `null` instead of to a bro, because this is enough to drop loot for the player.

### DIFF
--- a/mod_reforged/hooks/entity/tactical/actor.nut
+++ b/mod_reforged/hooks/entity/tactical/actor.nut
@@ -207,13 +207,20 @@
 		// Store the real killer here before switching _killer for loot purposes. This is used to fix MSU onDeathWithInfo and onOtherActorDeath functions to have the correct killer
 		this.m.RF_RealKiller = _killer;
 
-		if (_killer != null && _killer.getFaction() != ::Const.Faction.Player && (::Const.Faction.Player in this.m.RF_DamageReceived) && this.m.RF_DamageReceived[::Const.Faction.Player].Total / this.m.RF_DamageReceived.Total >= 0.5)
+		if (_killer != null && _killer.getFaction() != ::Const.Faction.Player && _killer.getFaction() != ::Const.Faction.PlayerAnimals)
 		{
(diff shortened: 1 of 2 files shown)


mod_reforged/hooks/entity/tactical/actor.nut
```diff
@@ -203,32 +204,22 @@
 
 	q.onDeath = @(__original) function( _killer, _skill, _tile, _fatalityType )
 	{
-		if ((::Const.Faction.Player in this.m.RF_DamageReceived) && this.m.RF_DamageReceived[::Const.Faction.Player].Total / this.m.RF_DamageReceived.Total >= 0.5)
+		// Store the real killer here before switching _killer for loot purposes. This is used to fix MSU onDeathWithInfo and onOtherActorDeath functions to have the correct killer
+		this.m.RF_RealKiller = _killer;
+
+		if (_killer != null && _killer.getFaction() != ::Const.Faction.Player && (::Const.Faction.Player in this.m.RF_DamageReceived) && this.m.RF_DamageReceived[::Const.Faction.Player].Total / this.m.RF_DamageReceived.Total >= 0.5)
```
Darxo, line 210: I don't see where damage done by the `::Const.Faction.PlayerAnimals` faction is accounted for.
As far as I can see that was already missing in the first implementation

Darxo (changes requested):

mod_reforged/hooks/entity/tactical/actor.nut
LordMidas: The damage from all factions is recorded in the `RF_DamageReceived` table in our hook on `onDamageReceived` in `skill_container`.

LordMidas: Oh I think I misunderstood your point. Do you mean that we don't include that here for loot drop? Yeah that is a valid point, I'll add a fix for that.

Darxo (approved):

PR #371, opened 2024-05-09
Safer changing of killer in actor on death

LordMidas:
After having merged #369 I realized that the part about passing the correct killer to MSU `skill_container` functions was unnecessary because MSU hooks the actor `onDeath` function _after_ Reforged. So I'm reverting that. But this also gave me the idea that we should do this changing of `_killer` as early as possible i.e. before most other mods would `hookTree` the `onDeath` of actor, so I have moved it to an Early bucket hook. This should make most mods safe from breaking due to our modification of `_killer`.

We don't have a clean setup for Early bucket hooks right now. The "proper" way would be to load all our files in the `First` bucket and then within each file we `push` the function to our custom buckets arrays and then queue them e.g. like we do currently with `FirstWorldInit` and `Late` buckets. But this requires a major refactor of every hook in the codebase so I've not done it in this PR.

### DIFF
--- a/scripts/!mods_preload/mod_reforged.nut
+++ b/scripts/!mods_preload/mod_reforged.nut
@@ -44,21 +44,17 @@ foreach (requirement in requiredMods)
 	::Reforged.HooksMod.hookTree("scripts/entity/tactical/actor", function(q) {
 		q.onDeath = @(__original) function( _killer, _skill, _tile, _fatalityType )
 		{
-			if (_killer != null && _killer.getFaction() != ::Const.Faction.Player && _killer.getFaction() != ::Const.Faction.PlayerAnimals)
-			{
-				local playerRelevantDamage = 0.0;
-				if (::Const.Faction.Player in this.m.RF_DamageReceived)
-					playerRelevantDamage += this.m.RF_DamageReceived[::Const.Faction.Player].Total;
-				if (::Const.Faction.PlayerAnimals in this.m.RF_DamageReceived)
-					playerRelevantDamage +=  this.m.RF_DamageReceived[::Const.Faction.PlayerAnimals].Total;
+			local playerRelevantDamage = 0.0;
+			if (::Const.Faction.Player in this.m.RF_DamageReceived)
+				playerRelevantDamage += this.m.RF_DamageReceived[::Const.Faction.Player].Total;
+			if (::Const.Faction.PlayerAnimals in this.m.RF_DamageReceived)
+				playerRelevantDamage +=  this.m.RF_DamageReceived[::Const.Faction.PlayerAnimals].Total;
 
-				// If _killer isn't already a player character or player animal AND the player + player animals did at least 50%
-				// of total damage to this actor, we set the _killer to null to ensure that the loot properly drops from this actor.
-				// This is because vanilla drops loot if _killer is null or belongs to Player or PlayerAnimals faction
-				// Warning: This will break any mod that hooks the original onDeath and expects _killer to represent the actual killer
-				if (playerRelevantDamage / this.m.RF_DamageReceived.Total >= 0.5)
-					_killer = null;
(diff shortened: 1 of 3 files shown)


scripts/!mods_preload/mod_reforged.nut
```diff
@@ -39,6 +39,32 @@ foreach (requirement in requiredMods)
 	queueLoadOrder.push(">" + (idx == null ? requirement : requirement.slice(0, idx)));
 }
 
+// TODO: Establish a cleaner way to organize "Early" bucket hooks on a per-file basis
+::Reforged.HooksMod.queue(queueLoadOrder, function() {
+	::Reforged.HooksMod.hookTree("scripts/entity/tactical/actor", function(q) {
+		q.onDeath = @(__original) function( _killer, _skill, _tile, _fatalityType )
+		{
+			if (_killer != null && _killer.getFaction() != ::Const.Faction.Player && _killer.getFaction() != ::Const.Faction.PlayerAnimals)
```
Darxo, line 47: Does this line mean that a last-hit still always guarantees the loot for the player?
As I understand it, you only turn non-player kills into null-kills with that new loot-mechanic so that the player overall gets loot in more situations.
But you don't change player-/null-kills into ally/enemy-kills in order to prevent the player from kill-stealing, like during early game cheese where you fight alongside a mercenary/noble army.

Is that observation correct?
If so, was that intended, are there any plans to fix that?

LordMidas: After your comment, we talked about it in Discord and decided that we don't want this cheese and only want to drop the loot if the player did at least 50% of the total damage. So I have added a commit which reverts this change.

scripts/!mods_preload/mod_reforged.nut
```diff
@@ -44,21 +44,18 @@ foreach (requirement in requiredMods)
...
+				playerRelevantDamage += this.m.RF_DamageReceived[::Const.Faction.Player].Total;
+			if (::Const.Faction.PlayerAnimals in this.m.RF_DamageReceived)
+				playerRelevantDamage +=  this.m.RF_DamageReceived[::Const.Faction.PlayerAnimals].Total;
+
+			// If _killer isn't already a player character or player animal AND the player + player animals did at least 50%
+			// of total damage to this actor, we set the _killer to null to ensure that the loot properly drops from this actor.
+			// This is because vanilla drops loot if _killer is null or belongs to Player or PlayerAnimals faction
+			// Warning: This will break any mod that hooks the original onDeath and expects _killer to represent the actual killer
+			if (playerRelevantDamage / this.m.RF_DamageReceived.Total >= 0.5)
+				_killer = null;
```
Darxo, line 58: This revert still doesn't fix the issue where the player can kill-steal and cheese early items.

For that to fix you need to change the _killer from being a player or `null` to some dummy entity from some dummy faction, when the player did **less** than 50% damage.

Darxo (changes requested):

scripts/!mods_preload/mod_reforged.nut
LordMidas: I'd say if the player didn't do 50% damage we can set the `_killer` to the dying entity himself. Would that be ok?

Darxo: Sounds good if that actually works.

Iirc in vanilla the loot is assigned to the player if an enemy dies to damage over time effects.
So in Reforged we need to intervene in this exact scenario so that the items are awared to the player **only** if he did 50%+ damage. And this part I'm missing atm

LordMidas: That part will be handled if we set every dying entity's killer to the entity himself if the player+player animals did less than 50% damage. So I will make this change.

Darxo (approved):

PR #372, opened 2024-05-10
Rework duelist validity

LordMidas:
- Remove requirement for base AP cost of skill to be <= 4.
- Add requirement for being the skill of a weapon whose primary attack has a Base AP cost of <=4 (this means it no longer work with unarmed attacks - this is acceptable imo because this keeps the code a lot cleaner).
- Still requires the skill to be melee with a base maximum range of 1 tile.

### DIFF
--- a/mod_reforged/config/items.nut
+++ b/mod_reforged/config/items.nut
@@ -1,0 +1,27 @@
+::Reforged.Items <- {
+	function isDuelistValid( _weapon )
+	{
+		if (_weapon.isEquipped())
+			return _weapon.getContainer().getActor().getSkills().getAttackOfOpportunity().isDuelistValid();
+
+		local copy = ::new(::IO.scriptFilenameByHash(_weapon.ClassNameHash));
+		local player = ::MSU.getDummyPlayer();
+		player.equip(copy);
+		local ret = player.getSkills().getAttackOfOpportunity().isDuelistValid();
+		player.unequip(copy);
+		return ret;
+	}
+
+	function getSkills( _item )
+	{
+		if (_item.isEquipped())
+			return _item.getSkills();
+
+		local copy = ::new(::IO.scriptFilenameByHash(_item.ClassNameHash));
+		local player = ::MSU.getDummyPlayer();
+		player.equip(copy);
+		local ret = copy.getSkills();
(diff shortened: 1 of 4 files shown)


Uberbagel (approved):

mod_reforged/config/items.nut
```diff
@@ -0,0 +1,27 @@
...
+		player.unequip(copy);
+		return ret;
+	}
+
+	function getSkills( _item )
+	{
+		if (_item.isEquipped())
+			return _item.getSkills();
+
+		local copy = ::new(::IO.scriptFilenameByHash(_weapon.ClassNameHash));
```
Darxo, line 20: This must be `_item_`, not `_weapon`

mod_reforged/config/items.nut
```diff
@@ -0,0 +1,27 @@
...
+		local player = ::MSU.getDummyPlayer();
+		player.equip(copy);
+		local ret = player.getSkills().getAttackOfOpportunity().isDuelistValid();
+		player.unequip(copy);
+		return ret;
+	}
+
+	function getSkills( _item )
```
Darxo, line 15: I don't see this function being used in this branch.
Have you planned anything for this?

mod_reforged/config/items.nut
```diff
@@ -0,0 +1,27 @@
+::Reforged.Items <- {
+	function isDuelistValid( _weapon )
+	{
+		if (_weapon.isEquipped())
+			return _weapon.getContainer().getActor().getSkills().getAttackOfOpportunity().isDuelistValid();
```
Darxo, line 5: Where are you checking whether the weapon has an AOO skill with 4 AP or lower like you states in the branch description?

I see you removed the old condition from inside the skill.nut hook but as far as I see now there is no more AP check at all and every melee weapon with a 1 range is duelist viable.

Darxo (changes requested):

mod_reforged/config/items.nut
LordMidas: Yes it is not being used. Just seemed like a useful function to have in case any of our team members or submodders want to get a list of the default skills of an item.

Darxo (approved):

PR #377, opened 2024-05-13
feat: add dynamic duo perk

LordMidas:


Darxo (changes requested):
I did a break at this point. Plan to continue another time

scripts/skills/actives/rf_dynamic_duo_shuffle_skill.nut
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	function getCursorForTile( _tile )
+	{
+		return ::Const.UI.Cursor.Rotation;
+	}
+
+	function isUsable()
+	{
+		return !this.m.IsSpent && this.skill.isUsable() && !this.getContainer().getActor().getCurrentProperties().IsRooted;
```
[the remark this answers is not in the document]
LordMidas: The way it has been designed is that the skill doesn't even exist when you don't have a partner i.e. it is only added via the dynamic duo perk when you select a partner. Therefore, this situation shouldn't arise. However, I'll still add these checks in especially for the `isPlacedOnMap` check. I will also make it unusable when more than 1 tile away from the partner.

scripts/skills/actives/rf_dynamic_duo_shuffle_skill.nut
```diff
@@ -47,20 +45,30 @@ this.rf_dynamic_duo_shuffle_skill <- ::inherit("scripts/skills/skill", {
...
+			});
+		}
+
+		if (::MSU.isNull(this.m.DynamicDuoPerk.getPartner()) || !this.m.DynamicDuoPerk.getPartner().isPlacedOnMap())
+		{
+			ret.push({
+				id = 22,
+				type = "text",
+				icon = "ui/tooltips/warning.png",
+				text = ::MSU.Text.colorRed("Can only be used when next to your partner")
```
Darxo, line 71: The warning here only barely fits with the conditions you test for.
I feel like the cleanest is to have an `if` into an `else if` where you test for both conditions above separately and then write a much more precise error message depending on the case.

And if you really find this warning about the partner needing to be adjacent, important. Then you should introduce a new third condition for this case to check for.

scripts/skills/effects/rf_dynamic_duo_partner_effect.nut
```diff
@@ -0,0 +1,47 @@
...
+			local perk = this.getContainer().getSkillByID("perk.rf_dynamic_duo");
+			if (perk != null)
+				perk.setPartnerSkill(null);
+		}
+	}
+
+	function onRemoved()
+	{
+		this.perk_rf_dynamic_duo_abstract.onRemoved();
+		this.setPartnerSkill(null);
```
Darxo, line 25: This is redundant. It's already done in the parent's `onRemoved()` function.

Darxo (approved):
My remaining issues are minor and not game breaking. Don't care much on whether they are resolved for this to be merged

scripts/skills/actives/rf_dynamic_duo_shuffle_skill.nut
LordMidas: I've rearranged the tooltip to be clearer about the blocking conditions. Please take another look.

PR #379, opened 2024-05-13
refactor: update perk groups with new design for version 0.5.0

LordMidas:
- Rename Large to Tough
- Rename Resilient to Vigorous
- Update all perk groups with the new design for 0.5.0


scripts/mods/mod_reforged/perk_groups/pg_rf_leadership.nut
Uberbagel, line 1: I expect we are renaming Inspire to Supporter, so the change will also need to be made here.

scripts/mods/mod_reforged/perk_groups/pg_rf_tough.nut
Uberbagel, line 1: If we're still using Flavor Text these ought to be updated.

Uberbagel (changes requested):

scripts/mods/mod_reforged/perk_groups/pg_rf_vigorous.nut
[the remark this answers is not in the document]
LordMidas: I'll probably get rid of the FlavorText stuff in Dynamic Perks system eventually so that part should be considered obsolete at the moment. We _are_ still using it in the tooltip of the background icon on the recruitment screen but we can remove that.

Uberbagel, line 1: ID still Resilient

Uberbagel (changes requested):

PR #380, opened 2024-05-14
refactor: rework whirling death perk

LordMidas:
- Gain +1 reach with every attack for one turn.
- For 1-handed flails perform a free attack against an adjacent enemy with lower current Reach than you, but with 50% less damage.
- For 2-handed flails gain +10% chance to hit the head per current Reach difference between you and the target, if you have higher Reach.

### DIFF
--- a/mod_reforged/hooks/config/strings.nut
+++ b/mod_reforged/hooks/config/strings.nut
@@ -2242,12 +2242,15 @@ foreach (vanillaDesc in vanillaDescriptions)
 		}]
  	}),
 	RF_WhirlingDeath = ::UPD.getDescription({
- 		Fluff = "Use the arc of your flailhead to target weak spots in your opponents' armor!",
+ 		Fluff = "Create a whirlwind of death with the spinning head of your flail!",
  		Requirement = "Flail",
  		Effects = [{
 			Type = ::UPD.EffectType.Passive,
 			Description = [
-				"An additional random " + ::MSU.Text.colorGreen("0-25%") + " of damage ignores armor."
+				"Every attack with flails increases your current [Reach|Concept.Reach] by " + ::MSU.Text.colorGreen("+1") + " until your next turn.",
+				"With two-handed flails, when attacking a target who has lower current [Reach|Concept.Reach] than you, gain " + ::MSU.Text.colorGreen("+10%") + " chance to hit the head per difference in current [Reach|Concept.Reach] between you and the target.",
+				"During your [turn|Concept.Reach], when attacking with one-handed flails, perform a free extra attack with " + ::MSU.Text.colorRed("50%") + " less damage against a different adjacent enemy who has lower current [Reach|Concept.Reach] than you.",
+				"Does not take into account abilities which ignore or modify [Reach|Concept.Reach] during an attack."
 			]
 		}]
  	}),
--- a/mod_reforged/hooks/config/strings.nut
+++ b/mod_reforged/hooks/config/strings.nut
@@ -2249,7 +2249,7 @@ foreach (vanillaDesc in vanillaDescriptions)
 			Description = [
 				"Every attack with flails increases your current [Reach|Concept.Reach] by " + ::MSU.Text.colorGreen("+1") + " until your next [turn|Concept.Turn].",
 				"With two-handed flails, when attacking a target who has lower current [Reach|Concept.Reach] than you, gain " + ::MSU.Text.colorGreen("+10%") + " chance to hit the head per difference in current [Reach|Concept.Reach] between you and the target.",
-				"During your [turn|Concept.Reach], when attacking with one-handed flails, perform a free extra attack with " + ::MSU.Text.colorRed("50%") + " less damage against a different adjacent enemy who has lower current [Reach|Concept.Reach] than you.",
+				"During your [turn|Concept.Turn], when attacking with one-handed flails, perform a free extra attack with " + ::MSU.Text.colorRed("50%") + " less damage against a different adjacent enemy who has lower current [Reach|Concept.Reach] than you. This extra attack does not increase your [Reach|Concept.Reach].",
 				"Does not take into account abilities which ignore or modify [Reach|Concept.Reach] during an attack."
 			]
 		}]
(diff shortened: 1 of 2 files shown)


scripts/skills/perks/perk_rf_whirling_death.nut
```diff
@@ -1,46 +1,168 @@
...
+	function isSkillValid( _skill )
+	{
+		if (!_skill.isAttack() || _skill.isRanged())
+			return false;
+
+		if (this.m.RequiredWeaponType == null)
+			return true;
+
+		local weapon = _skill.getItem();
+		return !::MSU.isNull(weapon) && weapon.isItemType(::Const.Items.ItemType.Weapon) && weapon.isWeaponType(::Const.Items.WeaponType.Flail);
```
Darxo, line 166: You should use `this.m.RequiredWeaponType` variable here instead of hard coding Flail

scripts/skills/perks/perk_rf_whirling_death.nut
```diff
@@ -1,46 +1,168 @@
...
-		if (this.getContainer().getActor().isDisarmed()) return false;
-
-		local weapon = this.getContainer().getActor().getMainhandItem();		
-		if (weapon == null || !weapon.isWeaponType(::Const.Items.WeaponType.Flail))
+	function onAnySkillExecuted( _skill, _targetTile, _targetEntity, _forFree )
+	{
+		if (this.isSkillValid(_skill) && !this.m.IsPerformingExtraAttack)
 		{
-			return false;
+			this.m.Stacks++;
```
Darxo, line 65: Do you really not get stacks from extra attacks triggered by this perk?
The extra attacks from flail spinner seem to be counted just fine.
Also the description of this perk says "every attack" with seemingly no exceptions.

mod_reforged/hooks/config/strings.nut
```diff
@@ -2242,12 +2242,15 @@ foreach (vanillaDesc in vanillaDescriptions)
...
- 		Fluff = "Use the arc of your flailhead to target weak spots in your opponents' armor!",
+ 		Fluff = "Create a whirlwind of death with the spinning head of your flail!",
  		Requirement = "Flail",
  		Effects = [{
 			Type = ::UPD.EffectType.Passive,
 			Description = [
-				"An additional random " + ::MSU.Text.colorGreen("0-25%") + " of damage ignores armor."
+				"Every attack with flails increases your current [Reach|Concept.Reach] by " + ::MSU.Text.colorGreen("+1") + " until your next [turn|Concept.Turn].",
+				"With two-handed flails, when attacking a target who has lower current [Reach|Concept.Reach] than you, gain " + ::MSU.Text.colorGreen("+10%") + " chance to hit the head per difference in current [Reach|Concept.Reach] between you and the target.",
+				"During your [turn|Concept.Reach], when attacking with one-handed flails, perform a free extra attack with " + ::MSU.Text.colorRed("50%") + " less damage against a different adjacent enemy who has lower current [Reach|Concept.Reach] than you.",
```
Darxo, line 2252: `[turn|Concept.Reach]` - Reach is the wrong concept for turn

scripts/skills/perks/perk_rf_whirling_death.nut
```diff
@@ -1,46 +1,168 @@
...
+
+	function onUpdate( _properties )
+	{
+		_properties.Reach += this.m.Stacks;
+	}
+
+	function onAnySkillUsed( _skill, _targetEntity, _properties )
+	{
+		if (this.m.IsPerformingExtraAttack)
+			_properties.DamageTotalMult *= this.m.DamageTotalMult;
```
Darxo, line 50: How does this work when **Flail Spinner** triggers during this extra attack?
My gut feeling says that **Flail Spinner**'s attack would be nerfed twice down to 25% damage because it happens deep within the extra attack from this perk, so `this.m.IsPerformingExtraAttack` is never set to `false` in between.

From the description one would think that **only** the one(1) singular attack caused by this perk would be nerfed and not any child attacks spawned by other things as a reaction to that.

Darxo (changes requested):

mod_reforged/hooks/config/strings.nut
LordMidas: Lol, nice catch!

scripts/skills/perks/perk_rf_whirling_death.nut
LordMidas: The "extra attacks" from this perk shouldn't give additional stacks of Reach, because they are a side-effect of your main attack. It's not intended - also seemed very overpowered in my preliminary testing when you can get +4 Reach with just 2 attacks. The extra attacks from Flail Spinner aren't "part of this perk" so they can give the extra Reach.

I thought this was obvious i.e. the extra attack from this perk doesn't give the extra Reach. But I can update the tooltip to be explicit about this.

scripts/skills/perks/perk_rf_whirling_death.nut
LordMidas: Basically when the target or attacker is visible to the player i.e. we are using the `::Time.scheduleEvent` methodology in both this perk and Flail Spinner, then both the perks act independently of each other because their variables are set/reset during that scheduled event.

However, when the target and attacker both are **not visible to the player**, then both the variables will be applied as then we aren't using scheduled events (this is following vanilla code style probably for performance and/or other hidden reasons). So this is definitely a situation that we should try to fix.

scripts/skills/perks/perk_rf_whirling_death.nut
```diff
@@ -1,46 +1,168 @@
...
+		{
+			if (!myTile.hasNextTile(i))
+				continue;
+
+			local nextTile = myTile.getNextTile(i);
+			if (!nextTile.IsOccupiedByActor || nextTile.isSameTileAs(_primaryTargetTile))
+				continue;
+
+			local entity = nextTile.getEntity();
+			if (!entity.isAlliedWith(user) && entity.getCurrentProperties().getReach() < user.getCurrentProperties().getReach() && _skill.onVerifyTarget(myTile, nextTile))
```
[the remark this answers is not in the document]
LordMidas: This is a good point. I didn't account for that check, but I think adjacent entities are always visible, so this should be fine without the check. And perhaps even from a design perspective, perhaps it is ok if the "whirling flail" hits even invisible adjacent entities. So, for now probably this is fine?

scripts/skills/perks/perk_rf_whirling_death.nut
LordMidas: I've pushed a fix for this in a new commit. Please check if it looks ok to you.

Darxo (approved):

PR #386, opened 2024-05-16
fix: set IsSpecializedInShields for zombie knight bodyguards to false

Darxo:
zombie knight bodyguards are the only vanilla entity where this flag is hard-codedly set to true in the base properties.
Currently this is redundant and I dislike the idea that in the future they secretly have some shield expert abilities without anything saying that in the tooltips.

So I find it consistent to once and for all reset that flag to the default value `false`

### DIFF
--- a/mod_reforged/hooks/entity/tactical/enemies/zombie_knight_bodyguard.nut
+++ b/mod_reforged/hooks/entity/tactical/enemies/zombie_knight_bodyguard.nut
@@ -2,8 +2,8 @@
 	q.onInit = @() function()
 	{
 	    this.zombie_knight.onInit();
-		// local b = this.m.BaseProperties;
-		// b.IsSpecializedInShields = true;
+		local b = this.m.BaseProperties;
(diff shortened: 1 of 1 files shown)


mod_reforged/hooks/entity/tactical/enemies/zombie_knight_bodyguard.nut
```diff
@@ -2,8 +2,8 @@
 	q.onInit = @() function()
 	{
 	    this.zombie_knight.onInit();
-		// local b = this.m.BaseProperties;
-		// b.IsSpecializedInShields = true;
+		local b = this.m.BaseProperties;
+		b.IsSpecializedInShields = false;	// this is the only enemy where this property in vanilla is hard-coded to true
```
LordMidas, line 6: We are already overwriting the vanilla function and the default value for this variable in character properties _is_ `false`. So I don't see it necessary that we should manually overwrite this. We completely overwrite the vanilla function, and show these lines as commented out to show that this is the part of the vanilla function that we "removed".

LordMidas (changes requested):

PR #391, opened 2024-05-19
feat: add support for new flags to turn off the push back from spearwall

Darxo:


mod_reforged/hooks/config/character.nut
```diff
@@ -25,6 +25,9 @@ local getClone = ::Const.CharacterProperties.getClone;
 ::Const.CharacterProperties.OffensiveReachIgnore <- 0;
 ::Const.CharacterProperties.BonusPerReachAdvantage <- 0;
 
+::Const.CharacterProperties.IsImmuneIsOnToEnterPushback <- false;	// Will an enemy spearwall push me back to my original tile when they hit the AOO?
+::Const.CharacterProperties.IsSpearwallPushingBack <- false;		// Will my spearwall-like AOO push enemies back that I hit when they try to enter?
```
LordMidas, line 29: A better name would be `IsPushingBackOnEnterAttack` or `IsPushingBackOnZOCEnterAttack` to go in line with the vanilla triggers for this attack i.e. `IsAttackingOnZoneOfControlEnter`.

mod_reforged/hooks/config/character.nut
```diff
@@ -25,6 +25,9 @@ local getClone = ::Const.CharacterProperties.getClone;
 ::Const.CharacterProperties.OffensiveReachIgnore <- 0;
 ::Const.CharacterProperties.BonusPerReachAdvantage <- 0;
 
+::Const.CharacterProperties.IsImmuneIsOnToEnterPushback <- false;	// Will an enemy spearwall push me back to my original tile when they hit the AOO?
```
LordMidas, line 28: typo extra `Is` after `IsImmune`

mod_reforged/hooks/entity/tactical/actor.nut
```diff
@@ -207,6 +207,23 @@
...
+
+		if (ret && _isOnEnter)
+		{
+			// Any of those two flags will prevent spearwall-like ZOC checks from pushing the attacker back
+			if (_entity.getCurrentProperties().IsImmuneIsOnToEnterPushback || !this.getCurrentProperties().IsSpearwallPushingBack)
+			{
+				_entity.setCurrentMovementType(oldMovementType);
+				return false;	// The attack happend but we still don't return false so that hopefully the .exe backend will not push the incoming entity back
+			}
+		}
```
LordMidas, line 224: Must return `ret` at the end of this function.

LordMidas (changes requested):

mod_reforged/hooks/config/character.nut
Darxo: It was not actually a typo. `IsOnToEnter` is the name of the variable used to determine whether the target entity entered my tile from the outside or tries to leave my tile.
But even still I don't like my name.

How about `IsImmuneToOnEnterPushback`?
So a **To** instead of the **Is**

mod_reforged/hooks/config/character.nut
Darxo: I like writing out the ZOC in this case.

How about `IsPushingBackOnZoneOfControlEnter`? So also without the `Attack` part.

LordMidas: That would be misleading because this property only pushes back after a successful attack.

PR #398, opened 2024-05-23
fix: combo perk to check skill validity before its use

LordMidas:
- Fixes issues with skills that remove themselves after use
- Fixes issues with the used skill potentially having a different AP cost after having been used.


Darxo (approved):
Good explanation

PR #399, opened 2024-05-23
balance: reduce base melee skill and defense of conscript

LordMidas:
- Drop base Melee Defense because they have Phalanx in Reforged which gives them additional Reach and permanent Shieldwall next to an allied Shieldwall. 
- Drop base Melee Skill because they have Backstabber in Reforged.


LordMidas:
Closing as Conscripts have been rebalanced in #405.

PR #400, opened 2024-05-23
fix: add missing validity checks for long reach surround

LordMidas:


Darxo (approved):
When you add a comment explanation then the pr is good to be merged

PR #401, opened 2024-05-23
feat: add keybind support for wait round action

Darxo:
The default Hotkey is CTRL + Spacebar
Reminder that normal Wait hotkey is Spacebar

image

End Turn for all Characters (vanilla) has the hotkey **R**. I prefer a single button hotkey but there are not so many options left. And honestly you probably use **Wait Round** a bit less than **End Round**.

### DIFF
--- a/mod_reforged/msu_systems/mod_settings.nut
+++ b/mod_reforged/msu_systems/mod_settings.nut
@@ -13,3 +13,16 @@ tacticalTooltipPage.addRangeSetting("CollapseActivesWhenX", 5, 0, 20, 1, "Collap
 tacticalTooltipPage.addBooleanSetting("TacticalTooltip_CollapseAsText", false, "Collapse as Text", "If enabled, then skills collapse using their names as text, otherwise they collapse using their icons.");
 tacticalTooltipPage.addBooleanSetting("ShowStatusPerkAndEffect", true, "Show Status Perk And Effect", "Some Perks are also Status Effects. Usually their Effect is hidden until some condition is fulfilled. When this setting is enabled, these perks show up in the Perks category even when they show up under Effects (e.g. when their effect is active). When disabled, when they appear under Effects, they will be hidden from the Perks category. This can help save space on the tooltip.");
 tacticalTooltipPage.addBooleanSetting("HeaderForEmptyCategories", false, "Show Header for empty categories");
+
+::Reforged.Mod.Keybinds.addSQKeybind("Tactical_WaitRound", "h", ::MSU.Key.State.Tactical, function()
+{
+	if (this.m.MenuStack.hasBacksteps() || this.isInputLocked() || this.isInCharacterScreen())
+	{
+		return false;
+	}
+	else
+	{
+		::Tactical.TurnSequenceBar.onWaitTurnAllButtonPressed();
+		return true;
+	}
+}, "Wait Turn with all Characters");
--- a/mod_reforged/msu_systems/tooltips.nut
+++ b/mod_reforged/msu_systems/tooltips.nut
@@ -25,7 +25,20 @@
 	},
 	Tactical = {
 		Button = {
-			WaitTurnAllButton = ::MSU.Class.BasicTooltip("Wait Round", "Have all your characters use \'Wait\' on their turn.")
+			WaitTurnAllButton = ::MSU.Class.CustomTooltip(function( _ ) {
+				return [
+					{
+						id = 1,
+						type = "title",
+						text = format("Wait Round (%s)", ::MSU.System.Keybinds.KeybindsByMod["mod_reforged"]["Tactical_WaitRound"].getKeyCombinations())
+					},
+					{
+						id = 2,
+						type = "description",
+						text = "Have all your characters use \'Wait\' on their turn."
+					}
+				];
+			})
 		}
 	}
 });
--- a/mod_reforged/msu_systems/tooltips.nut
+++ b/mod_reforged/msu_systems/tooltips.nut
@@ -35,7 +35,7 @@
 					{
 						id = 2,
 						type = "description",
-						text = "Have all your characters use \'Wait\' on their turn."
+						text = ::Reforged.Mod.Tooltips.parseString("Have all your characters use [Wait|Concept.Wait] on their [turn|Concept.Turn].")
 					}
 				];
 			})


mod_reforged/msu_systems/mod_settings.nut
```diff
@@ -13,3 +13,10 @@ tacticalTooltipPage.addRangeSetting("CollapseActivesWhenX", 5, 0, 20, 1, "Collap
 tacticalTooltipPage.addBooleanSetting("TacticalTooltip_CollapseAsText", false, "Collapse as Text", "If enabled, then skills collapse using their names as text, otherwise they collapse using their icons.");
 tacticalTooltipPage.addBooleanSetting("ShowStatusPerkAndEffect", true, "Show Status Perk And Effect", "Some Perks are also Status Effects. Usually their Effect is hidden until some condition is fulfilled. When this setting is enabled, these perks show up in the Perks category even when they show up under Effects (e.g. when their effect is active). When disabled, when they appear under Effects, they will be hidden from the Perks category. This can help save space on the tooltip.");
 tacticalTooltipPage.addBooleanSetting("HeaderForEmptyCategories", false, "Show Header for empty categories");
+
+::Reforged.Mod.Keybinds.addSQKeybind("WaitRound", "ctrl+space", ::MSU.Key.State.Tactical, function()
```
LordMidas, line 17: For convention let's prefix the scope of the keybind to its ID. Something like `tactical_WaitRound`. We do this in MSU too and it's good to keep things organized.

mod_reforged/msu_systems/mod_settings.nut
```diff
@@ -13,3 +13,10 @@ tacticalTooltipPage.addRangeSetting("CollapseActivesWhenX", 5, 0, 20, 1, "Collap
...
+{
+	if (this.m.MenuStack.hasBacksteps() || this.isInputLocked() || this.isInCharacterScreen()) return;
```
LordMidas, line 19: Newline the return. I'd do it without braces in this case to keep things concise.

LordMidas (approved):
Looks good to me after the couple of changes requested. I'd show this to Taro/Enduriel just in case because they have experience with the keybinds system.

mod_reforged/msu_systems/mod_settings.nut
Darxo: This is how MSU does the formatting of keybind conditions. 
I copied it thinking that in this case you are ok with it considering you must have approved it over there too

mod_reforged/msu_systems/mod_settings.nut
Enduriel: I actually prefer 2 as well

mod_reforged/msu_systems/mod_settings.nut
Darxo: Here is the link to that section in MSU https://github.com/MSUTeam/MSU/blob/e0f6edf0465ee0f46c49d199009acd2bcc462ece/msu/vanilla_mod/vanilla_keybinds.nut#L287

I'm really interested in whether you think that section in MSU is hard to read/parse and should be changed.

LordMidas: That's older MSU code before Enduriel decided that we need to newline after `if` conditions. Older MSU code has many single-line `if` statements. I don't really mind the way it looks in this case in a single line but I was thinking that Enduriel would maybe now want it to be newlined and I like newlined too (without braces in this case).

mod_reforged/msu_systems/mod_settings.nut
```diff
@@ -13,3 +13,10 @@ tacticalTooltipPage.addRangeSetting("CollapseActivesWhenX", 5, 0, 20, 1, "Collap
 tacticalTooltipPage.addBooleanSetting("TacticalTooltip_CollapseAsText", false, "Collapse as Text", "If enabled, then skills collapse using their names as text, otherwise they collapse using their icons.");
 tacticalTooltipPage.addBooleanSetting("ShowStatusPerkAndEffect", true, "Show Status Perk And Effect", "Some Perks are also Status Effects. Usually their Effect is hidden until some condition is fulfilled. When this setting is enabled, these perks show up in the Perks category even when they show up under Effects (e.g. when their effect is active). When disabled, when they appear under Effects, they will be hidden from the Perks category. This can help save space on the tooltip.");
 tacticalTooltipPage.addBooleanSetting("HeaderForEmptyCategories", false, "Show Header for empty categories");
+
+::Reforged.Mod.Keybinds.addSQKeybind("Tactical_WaitRound", "ctrl+space", ::MSU.Key.State.Tactical, function()
```
LordMidas, line 17: As discussed on Discord, `Ctrl+Space` is not a good keybind to have due to conflicts with Windows. I suggested the use of `Alt+R` because `R` is already the end-round button and having an `alt` modifier for such keys to do something which is similar in theme seems like a good design. Other choices such as `H` or `G` are keys which could potentially be used for a more frequently used thing, because Wait-All is a very niche use case.

LordMidas (changes requested):

mod_reforged/msu_systems/mod_settings.nut
Darxo: I tend towards **H**

mod_reforged/msu_systems/mod_settings.nut
Enduriel: This should still `return false`. I just realized this isn't something we've really documented but the `return null` is handled to preserve backwards compatibility, not because it's the 'right' way to do this.

mod_reforged/msu_systems/mod_settings.nut
LordMidas: Let's finalize this and go with `Alt+R` for now.

Darxo: Alt is the hotkey for toggling the entity overlays though.
And from my tests that overlay is toggled every time I let go of my Alt key. Even if I suppress vanilla keybinds via MSU setting.

Darxo:
In the mod [Autopilot New](https://www.nexusmods.com/battlebrothers/mods/675) this action has the hotkey **"H"**.
So one reason to go with H is because it will be the one that most people expect this action to be.

LordMidas:
> In the mod [Autopilot New](https://www.nexusmods.com/battlebrothers/mods/675) this action has the hotkey **"H"**. So one reason to go with H is because it will be the one that most people expect this action to be.

Ok let's go with `H` then.

mod_reforged/msu_systems/tooltips.nut
```diff
@@ -25,7 +25,7 @@
 	},
 	Tactical = {
 		Button = {
-			WaitTurnAllButton = ::MSU.Class.BasicTooltip("Wait Round", "Have all your characters use \'Wait\' on their turn.")
+			WaitTurnAllButton = ::MSU.Class.BasicTooltip("Wait Round (Alt+R)", "Have all your characters use \'Wait\' on their turn.")
```
LordMidas, line 28: Is it possible that the keybind mentioned in the brackets is dynamically pulled from what the user has set the keybind to?

Darxo: I only know that this is in line with the other vanilla tooltips (which obviously didnt expect to ever be changed)

LordMidas: Try using `::MSU.Class.CustomTooltip`. Like shown below. Test if that works for fetching the keybind dynamically and if yes then let's go with that.

::MSU.Class.CustomTooltip(function( _ ) {
	return [
		{
			id = 1,
			type = "title",
			text = "Wait Round " + GETKEYBINDHERE
		},
		{
			id = 2,
			type = "description",
			text = DESCRIPTIONHERE
		}
	];
});

Darxo: I tried that out and it works.
Only ugly thing about is is that MSU does not offer an interface for accessing the keybind combination, or the keybind object in the first place.
So I had to do that in a manual way

LordMidas: @TaroEld @Enduriel We need to add a way in MSU to get a keybind object by id.

LordMidas: @Darxo Could you please push the commit with your current manual approach?

Darxo: Done

LordMidas (approved):

PR #402, opened 2024-05-23
fix: puncture being affected by reforged reworked double grip

LordMidas:
I really don't like that we have to overwrite the vanilla function and do the switcheroo but I don't know if there is a cleaner way. Please suggest a better way if possible.

### DIFF
--- a/mod_reforged/hooks/skills/special/double_grip.nut
+++ b/mod_reforged/hooks/skills/special/double_grip.nut
@@ -2,6 +2,7 @@
 	// Is set during onUpdate so that hybrid weapons are properly dealt with in functions other than onUpdate
 	// because we go through weapon types alphabetically and choose the first applicable type
 	q.m.CurrWeaponType <- null;
+	q.m.MeleeDamageMult_Dagger <- 1.2;
 
 	q.create = @(__original) function()
 	{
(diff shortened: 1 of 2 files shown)


mod_reforged/hooks/skills/special/double_grip.nut
```diff
@@ -297,6 +293,11 @@
 				}
 			}
 		}
+
+		if (_skill.getID() != "actives.puncture" && this.m.CurrWeaponType == ::Const.Items.WeaponType.Dagger)
+		{
+			_properties.MeleeDamageMult *= 1.2;
```
Darxo, line 299: I prefer if you do the vanilla approach but all within this double_grip.nut file.
- So first you grant the damage bonus during `onUpdate` so that it is displayed in the characters stats page how a player would 
- Then during `onAnySkillUsed` you divide the `MeleeDamageMult` through the same value so that it is hopefully back to the normal damage value for puncture only

For that you should then introduce a shared member variable for the dagger damage mult

mod_reforged/hooks/skills/special/double_grip.nut
```diff
@@ -294,9 +299,9 @@
 			}
 		}
 
-		if (_skill.getID() != "actives.puncture" && this.m.CurrWeaponType == ::Const.Items.WeaponType.Dagger)
+		if (_skill.getID() == "actives.puncture" && this.m.CurrWeaponType == ::Const.Items.WeaponType.Dagger)
```
Darxo, line 302: Just like in vanilla we now have to re-add some check like `canDoubleGrip` here.

Darxo (changes requested):

mod_reforged/hooks/skills/special/double_grip.nut
LordMidas: That check is already there due to the `this.m.CurrWeaponType` thing because that is only set to Dagger during `onUpdate` if `canDoubleGrip`.

Darxo (approved):

PR #403, opened 2024-05-24
refactor: move reload skills slightly forward in SkillOrder

Darxo:
This will fix issues where new skills like Cheap Trick are put before Reload

### DIFF
--- a/mod_reforged/hooks/skills/actives/reload_bolt.nut
+++ b/mod_reforged/hooks/skills/actives/reload_bolt.nut
@@ -3,6 +3,7 @@
 	{
 		__original();
 		this.m.IsRemovedAfterBattle = false;
+		this.m.Order = ::Const.SkillOrder.OffensiveTargeted;	// Players expect this skill to come earlier than most other active skills because of muscle memory, so we move it forward to be immediately after the shoot skill
 	}
 
 	// Overwrite vanilla function to prevent removal of this skill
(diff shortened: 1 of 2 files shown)


mod_reforged/hooks/skills/actives/reload_bolt.nut
```diff
@@ -3,6 +3,7 @@
 	{
 		__original();
 		this.m.IsRemovedAfterBattle = false;
+		this.m.Order -= 100;	// Player expect this perk to come earlier than most other active skills because of muscle memory, so we move it forward a bit
```
LordMidas, line 6: I'd just set this to `this.m.Order = ::Const.SkillOrder.OffensiveTargeted` as that is the order of shoot_bolt. This will ensure that this skill will **always** appear immediately after the shoot bolt skill which is the most intuitive.

LordMidas, line 6: - Comment should say `skill`, not `perk`.
- After the proposed skill order change, the comment can be updated to say that we set it like this so that it appears immediately after the shoot bolt skill.

LordMidas (changes requested):

LordMidas (approved):
Would be nice if you can change `Player` to `Players` in the comment before merging.

LordMidas:
Please change the commit message to remove mention of Cheap Trick because this change should be independent of that. The commit message should simply say that we want this reload skill to always be immediately after the shoot skill.

PR #404, opened 2024-05-25
fix: give stalwart a unique icon to differentiate it from battering ram

Darxo:


mod_reforged/hooks/skills/perks/perk_stalwart.nut
```diff
@@ -0,0 +1,7 @@
+::Reforged.HooksMod.hook("scripts/skills/perks/perk_stalwart", function(q) {
+	q.create = @(__original) function()
+	{
+		__original();
+		this.m.Icon = "ui/perks/perk_18";	// Unused vanilla perk icon of an anvil
```
LordMidas, line 5: That's actually a "weight" icon and we are using it for Encumbrance in Reforged. So we need a different icon.

LordMidas (changes requested):

mod_reforged/hooks/skills/perks/perk_stalwart.nut
Darxo: Nevermind then. I'll close this PR then, didn't wanna do a completely new icon for them effects just for enemy tooltips

PR #409, opened 2024-05-28
refactor: improve split_shield hook making it more moddable

Darxo:
Extract shield damage calculation into new function. Hide tooltips for effects that are disabled.
Use this.getName() instead of a hard coded name for the combat log.

### DIFF
--- a/mod_reforged/hooks/skills/actives/split_shield.nut
+++ b/mod_reforged/hooks/skills/actives/split_shield.nut
@@ -54,7 +54,7 @@
 		if (shield != null)
 		{
 			this.spawnAttackEffect(_targetTile, ::Const.Tactical.AttackEffectSplitShield);
-			local damage = this.RF_getShieldDamage();
+			local damage = this.RF_getShieldDamage(targetEntity);
 
 			local conditionBefore = shield.getCondition();
 			shield.applyShieldDamage(damage);
@@ -106,7 +106,7 @@
 	}
 
 // New Functions
-	q.RF_getShieldDamage <- function()
+	q.RF_getShieldDamage <- function( _targetEntity = null )
 	{
 		if (::MSU.isNull(this.getContainer()) || ::MSU.isNull(this.getContainer().getActor()))
 			return 1; // return a non-zero number so the tooltip gets added
--- a/mod_reforged/hooks/skills/actives/split_shield.nut
+++ b/mod_reforged/hooks/skills/actives/split_shield.nut
@@ -70,7 +70,7 @@
 			{
 				if (this.m.SoundOnHit.len() != 0)
 				{
-					::Sound.play(this.m.SoundOnHit[::Math.rand(0, this.m.SoundOnHit.len() - 1)], ::Const.Sound.Volume.Skill, _targetTile.getEntity().getPos());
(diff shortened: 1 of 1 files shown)


mod_reforged/hooks/skills/actives/split_shield.nut
LordMidas, line 1: I pushed a commit that removes the `_user` parameter from the damage function, as it seemed unnecessary. I've also added handling for when it is the dummy player.

mod_reforged/hooks/skills/actives/split_shield.nut
```diff
@@ -3,13 +3,14 @@
...
-		if (this.getShieldDamage(this.getContainer().getActor()) != 0)
+		local shieldDamage = this._RF_getShieldDamage();
+		if (shieldDamage != 0)
 		{
 			ret.push({
 				id = 7,
 				type = "text",
 				icon = "ui/icons/shield_damage.png",
-				text = "Inflicts " + ::MSU.Text.colorGreen(this.getShieldDamage(this.getContainer().getActor())) + " damage to shields"
+				text = format("Inflicts%s damage to shields", ::MSU.isNull(this.getContainer()) || this.getContainer().getActor().getID() == ::MSU.getDummyPlayer().getID() ? "" : " " + ::MSU.Text.colorRed(shieldDamage))
```
Darxo, line 13: You are missing a check for `::MSU.isNull(this.getContainer().getActor())` here before you access `getID`.
At least in the `RF_getShieldDamage` you added such a check.

LordMidas: Vanilla sometimes seems to have that check and sometimes it doesn't. I personally think that check is unnecessary as an item_container should be expected to have an actor. I am not aware of a case where an is present in a container that doesn't have an actor, at least in the expected flow of the game. We can probably get away with removing that check from the other functions too, I guess I was just being "over safe" with it there.

mod_reforged/hooks/skills/actives/split_shield.nut
```diff
@@ -108,6 +115,16 @@
 		return true;
 	}
 
+// New Functions
+	q.RF_getShieldDamage <- function()
+	{
+		if (::MSU.isNull(this.getContainer()) || ::MSU.isNull(this.getContainer().getActor()))
+			return 1; // return a non-zero number so the tooltip gets added
```
LordMidas, line 122: I am not happy about my solution here to return 1. It will add the damage tooltip to the skill file always - but here we allow with null weapon returning 0 which will remove that entry.

So perhaps it is worth it to assume that this skill will always be attached to a weapon and therefore we can simply always access the weapon's damage and return that?

mod_reforged/hooks/skills/actives/split_shield.nut
Darxo: I also like removing that check and seeing how it goes under the assumption that is really never happens.

mod_reforged/hooks/skills/actives/split_shield.nut
Darxo: For the beast scenario we want something like `this.m.ShieldDamageOverride` which makes the `RF_getShieldDamage` function ignore anything else and return that value.

Apart from that I think we can expect a weapon to be present.

Darxo:
I pushed a few commits.
One of them rewrites your idea of displaying a default message when we are not sure if this skill applies shield damage.

PR #411, opened 2024-05-30
refactor: remove 10% hitchance penalty for hand to hand combat

Darxo:
As suggested 3 months ago and liked by both of you, this is the hand to hand buff.

### DIFF
--- a/mod_reforged/hooks/skills/actives/hand_to_hand.nut
+++ b/mod_reforged/hooks/skills/actives/hand_to_hand.nut
@@ -1,0 +1,29 @@
+::Reforged.HooksMod.hook("scripts/skills/actives/hand_to_hand", function(q) {
+	q.getTooltip = @(__original) function()
+	{
+		local ret = __original();
+
+		foreach (i, entry in ret)
+		{
+			if (entry.id == 6)
+			{
+				ret.remove(i); // Remove any mention about the hitchance debuff
+				break;
+			}
+		}
+
+		return ret;
+	}
+
+	q.onAnySkillUsed = @(__original) function( _skill, _targetEntity, _properties )
+	{
+		local old_MeleeSkill = _properties.MeleeSkill;
+
+		__original(_skill, _targetEntity, _properties);
+
+		if (_skill == this)
+		{
+			_properties.MeleeSkill = old_MeleeSkill; // This reverts the vanilla -10 Modifier
+		}
+	}
+});


LordMidas (approved):
I've pushed a commit cleaning up a couple of things.

LordMidas:
The commit message should be changed from `refactor:` to `balance:`

mod_reforged/hooks/skills/actives/hand_to_hand.nut
```diff
@@ -3,11 +3,11 @@
 	{
 		local ret = __original();
 
-		for (local index = (ret.len() - 1); index >= 0; index--)
+		foreach (i, entry in ret)
```
Darxo, line 6: I chose the decrementing for loop just to be future proof in case we want to change/remove another entry line.

Darxo:
> The commit message should be changed from `refactor:` to `balance:`

I realize that the change itself it very simple, just that I add a new hook and change tooltip lines. This is not a simple change of a variable value.
Are you still sure such changes should be labelled `balance` aswell?

mod_reforged/hooks/skills/actives/hand_to_hand.nut
LordMidas: Alright I switched it to a decrementing for loop.

LordMidas:
> > The commit message should be changed from `refactor:` to `balance:`
> 
> I realize that the change on a surface level is very simple. Except that under the hood I add a new hook and change tooltip lines. This is not as simple as changing a variable value. Are you still sure such changes should be labelled `balance` aswell?

In terms of the reasoning of this change it is a balance change considering Reach now applies in Reforged. However, in terms of a change to the skill itself it can be considered a refactor (if we think about it independent of Reach). So it is a bit of a grey area. If you want to go with refactor here for now, that's fine, but balance seems preferable to me (if you put in commit message description that this was done considering the Reach mechanic in Reforged).

mod_reforged/hooks/skills/actives/hand_to_hand.nut
LordMidas: The decrementing one has a catch though and that is if a mod/someone else adds a tooltip entry with id 6 then it will remove that one instead. So we also need to check for the icon as you had done earlier - but that isn't safe either as someone could add an entry with id 6 and hitchance icon different from the one we are intending to remove. So, maybe the incrementing loop is better?

Darxo: We can't put any other operations in that for loop like additional removal or changes of other tooltips. So that whole for loop at that point might aswell be a global utility `removeTooltipLine` function which takes a tooltip and then up to three additional arguments for id, icon and text.

Even still, an incrementing for loop will also fail when a mod adds new global tooltip lines to the `getDefaultTooltip` function in the skill.nut`, with the id 6.

LordMidas: > We can't put any other operations in that for loop like additional removal or changes of other tooltips. So that whole for loop at that point might aswell be a global utility removeTooltipLine function which takes a tooltip and then up to three additional arguments for id, icon and text.

I don't know if we need a helper function for this but maybe it'd be useful. Seems like it is a rather niche use case and if you want to replace multiple entries then it is much more efficient to use a single loop rather than multiple calls to a function like that.

> Even still, an incrementing for loop will also fail when a mod adds new global tooltip lines to the getDefaultTooltip function in the skill.nut, with the id 6.

I think it is reasonable to say that this is a much less likely scenario than someone adding additional entries to the tooltip of individual skills at the bottom with the id 6.

In any case, it doesn't really matter to "pre-emptively optimize" for future potential scenarios. The most stable solution at this point is going to be one with an incrementing loop and if/when we want to remove more entries in the future we can switch to a decrementing one.

mod_reforged/hooks/skills/actives/hand_to_hand.nut
```diff
@@ -0,0 +1,29 @@
+::Reforged.HooksMod.hook("scripts/skills/actives/hand_to_hand", function(q) {
+	q.getTooltip = @(__original) function()
+	{
+		local ret = __original();
+
+		for (local i = ret.len() - 1; i >= 0; i++)
```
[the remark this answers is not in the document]
LordMidas: I've reverted the commit that changed it to a decrementing loop, so we are back to an incrementing loop now as discussed in the other conversation.

Darxo:
Looks good

LordMidas (approved):

PR #413, opened 2024-06-04
feat: add applicable mastery concept to nested tooltips

Enduriel:


mod_reforged/config/z_nested_tooltips.nut
```diff
@@ -88,6 +88,7 @@ local getThresholdForInjury = function( _script )
...
 		ZoneOfControl = ::MSU.Class.BasicTooltip("Zone of Control", ::Reforged.Mod.Tooltips.parseString("Most melee characters exert Zone of Control on their surrounding tiles. Trying to move out of enemy Zone of Control will trigger one free attack from each controlling enemy until the first attack hit or every attack missed. A hit will cancel the movement.")),
 		BagSlots = ::MSU.Class.BasicTooltip("Bag Slots", ::Reforged.Mod.Tooltips.parseString("Bag slots can be used to store additional weapons or utility items for use during combat. Every character has " + ::new("scripts/items/item_container").getUnlockedBagSlots() + " bag slots by default and can have a maximum of " + ::Const.ItemSlotSpaces[::Const.ItemSlot.Bag] + ".")),
+		ApplicableMastery = ::MSU.Class.BasicTooltip("Applicable Mastery", ::Reforged.Mod.Tooltips.parseString("A weapon requires its wielder to possess just one of its applicable masteries for all skills to fully benefit from mastery effects.")),
```
LordMidas, line 91: BB Players don't really understand this so I'd go in a bit more detail:
```squirrel
"In vanilla Battle Brothers weapon mastery [perks|Concept.Perk] affect skills, not weapons. So if a Sword/Cleaver hybrid weapon has the [Slash|Skill+slash] skill and the [Decapitate|Skill+decapitate] skill, then [Slash|Skill+slash] benefits from [Sword Mastery|Perk+perk_mastery_sword] and [Decapitate|Skill+decapitate] benefits from [Cleaver Mastery|Perk+perk_mastery_cleaver]. This makes it difficult to work with hybrid weapons.

In Reforged, a user requires only one of the applicable masteries of a weapon for all skills of that weapon to fully benefit from mastery effects. So, in the example above, the user would need either [Sword Mastery|Perk+perk_mastery_sword] or [Cleaver Mastery|Perk+perk_mastery_cleaver] and the weapon would behave as if the user had [Sword Mastery|Perk+perk_mastery_sword] for [Slash|Skill+slash] and [Cleaver Mastery|Perk+perk_mastery_cleaver] for [Decapitate|Skill+decapitate].

LordMidas (changes requested):

mod_reforged/config/z_nested_tooltips.nut
Enduriel: I do not see how that is necessary, they don't need to understand how it works in vanilla, just how it works in reforged, and for that, it is sufficiently clear imo.

LordMidas: I know from experience because of the number of times I've had to explain this to people lol, that they don't understand what it means that all skills benefit from mastery effects i.e. it isn't clear to them that by picking Cleaver Mastery on a Sword/Cleaver hybrid they are causing Riposte to now have the extra chance to hit that is normally received from Sword Mastery.

Enduriel: I see, if you think it's necessary we can try and include that but I still do not think an example is appropriate as a nested tooltip, I think something more abstracted and shorter would be better.

Uberbagel: I agree with Endy on this one and think the current description in this PR is good.

LordMidas: I disagree and I think that the nested tooltip should provide a detailed explanation. I've had to explain this concept numerous times to players because they don't really understand it. It would be ok to be very concise if it were a Tip of the Day but if it is a nested tooltip we can take the liberty to be more verbose.

LordMidas: I've gone with my concise example version above and rebased the branch.

LordMidas (approved):

PR #414, opened 2024-06-04
Lone wolf greatsword

Uberbagel:
I think Lone Wolf should start with greatsword instead of longsword, as it fits the theme of the origin and character more, and also is more in line with how vanilla lone wolf is intended to be


Darxo:
This PR contains unrelated commits. You should move them to a different branch and then drop them from this PR

Uberbagel:
> This PR contains two unrelated commits. You should move them to a different branch and then drop them from this PR

Thanks, accident

LordMidas (approved):

PR #424, opened 2024-06-08
refactor: rework encourage skill

LordMidas:
- Reduce range to 2 tiles.
- Increase resolve bonus to 50%.
- Make usable once per turn.
- Prevent use on Fleeing characters.
- Allow using on characters whose morale state is lower than you per tile of distance.

### DIFF
--- a/scripts/skills/actives/rf_encourage_skill.nut
+++ b/scripts/skills/actives/rf_encourage_skill.nut
@@ -78,7 +78,7 @@ this.rf_encourage_skill <- this.inherit("scripts/skills/skill", {
 			return false;
 
 		local actor = this.getContainer().getActor();
-		return actor.getFaction() == target.getFaction() && target.getMoraleState() < actor.getMoraleState() - actor.getTile().getDistanceTo(target.getTile());
+		return actor.getFaction() == target.getFaction() && actor.getMoraleState() - target.getMoraleState() >= actor.getTile().getDistanceTo(target.getTile());
 	}
 
 	function isUsable()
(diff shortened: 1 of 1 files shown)


scripts/skills/actives/rf_encourage_skill.nut
```diff
@@ -44,64 +46,68 @@ this.rf_encourage_skill <- this.inherit("scripts/skills/skill", {
...
+// New Functions
+	function getEncourageBonus()
 	{
-		if (::Tactical.Entities.getFlags().has(_targetEntity.getID() + this.getID() + "MarkedWithRoundNumber"))
-		{
-			return ::Tactical.Entities.getFlags().get(_targetEntity.getID() + this.getID() + "MarkedWithRoundNumber");
-		}
-
-		return 0;
+		local encourageBonus = this.getContainer().getActor().getCurrentProperties().getBravery() * this.m.EncourageBonusFraction;
```
Darxo, line 110: Vanilla floors this value in the `rally_the_troops` file. We should also do that or round it at least so that the checkMorale function doesn't receive an unexpected value.

Though I don't understand why we not already see decimal places on the value in the current implementation

Darxo (approved):
Looks good. The thing I commented on was already present before

scripts/skills/actives/rf_encourage_skill.nut
LordMidas: We are using `::Math.max` function here already and that floors it.

PR #425, opened 2024-06-08
feat: add more southern names and extend last names using basic names

LordMidas:
This adds more variety for southern names:
- Add 59 new names.
- Extend the vanilla "southern last names" list by prefixing `Ibn` and `Al-` to all the first names.

### DIFF
--- a/mod_reforged/hooks/config/character_names.nut
+++ b/mod_reforged/hooks/config/character_names.nut
@@ -308,10 +308,23 @@ foreach(characterName in ::Const.Strings.CharacterNames)
 // The array contains some duplicates even in vanilla
 ::Const.Strings.SouthernNames = ::MSU.Array.removeDuplicates(::Const.Strings.SouthernNames);
 
-foreach (name in ::Const.Strings.SouthernNames)
+foreach (i, name in ::Const.Strings.SouthernNames)
 {
 	::Const.Strings.SouthernNamesLast.push("Ibn " + name);
 	::Const.Strings.SouthernNamesLast.push("Al-" + name);
+	// For Officer Titles we push all Al- variants but fewer Ibn variants
+	// and a small number of long variants of the form "Al-Name Ibn Name"
+	// this is similar to the distribution of vanilla SouthernOfficerTitles
+	::Const.Strings.SouthernOfficerTitles.push("Al-" + name);
+	if (i % 3 == 0)
+	{
+		::Const.Strings.SouthernOfficerTitles.push("Ibn " + name);
+	}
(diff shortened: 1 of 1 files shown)


Uberbagel:
> Extend the vanilla "southern last names" list by prefixing Ibn and Al- to all the first names.

Can you explain what this means in practical terms?

LordMidas:
> > Extend the vanilla "southern last names" list by prefixing Ibn and Al- to all the first names.
> 
> Can you explain what this means in practical terms?

`Ibn` means "son of` and `Al-` means "the" in Arabic. Vanilla has two lists `SouthernNames` and `SouthernNamesLast`. It generates characters by using a name from `SouthernNames` and then sometimes it affixes a last name from `SouthernNamesLast`. These last names in vanilla are composed of formations such as `Ibn Sahr` or `Al-Faris`. I think there is no reason why can't expand that setup by just prefixing `Ibn` and `Al-` to all the first names. This results in a lot more possible valid last names.

Uberbagel:
So now we lose Ibn Sahr and Al-Faris and only get Ibn Al-???

LordMidas:
> So now we lose Ibn Sahr and Al-Faris and only get Ibn Al-???

No. We extend the last names list by prefixing Ibn and Al- to all the first names. So if a first name was "Adil" for example then both "Ibn Adil" and "Al-Adil" are now additional valid last names to roll from.

mod_reforged/hooks/config/character_names.nut
```diff
@@ -242,3 +242,71 @@ foreach(characterName in ::Const.Strings.CharacterNames)
...
+	"Zuhayr",
+	"Zahid",
+	"Zubayr"
+]);
+
+foreach (name in ::Const.Strings.SouthernNames)
+{
+	::Const.Strings.SouthernNamesLast.push("Ibn " + name);
+	::Const.Strings.SouthernNamesLast.push("Al-" + name);
+}
```
Darxo, line 312: The name `Wasi` appears in `SouthernNames` and also in `SouthernNamesLast`. There might be more name overlaps.
Furthermore as described above, two names appear twice in the list bove.

You should remove duplicates after generating this name list, like we do above with the knight names

Darxo (changes requested):

Uberbagel:
Sounds good then

Uberbagel (approved):

Darxo (approved):

Darxo (approved):

Uberbagel (approved):

PR #426, opened 2024-06-08
refactor: show troop cost instead of strength in world entity tooltip

LordMidas:
The cost is a better indication of the "resources" that were spent on this troop composition and serves a better purpose to compare the spawns for which this tooltip entry was added.

### DIFF
--- a/mod_reforged/hooks/entity/world/location.nut
+++ b/mod_reforged/hooks/entity/world/location.nut
@@ -26,7 +26,7 @@
 		{
 			foreach (troop in ::Const.World.Spawn.Troops)
 			{
-				if (troop.ID = t.ID && troop.Script == t.Script)
+				if (troop.ID == t.ID && troop.Script == t.Script)
 				{
 					worth += troop.Cost;
 					break;
--- a/mod_reforged/hooks/entity/world/location.nut
+++ b/mod_reforged/hooks/entity/world/location.nut
@@ -20,13 +20,20 @@
 	q.getTooltip = @(__original) function()
 	{
 		local ret = __original();
+
+		local worth = 0;
+		foreach (t in this.m.Troops)
+		{
+			worth += t.Cost;
+		}
+
 		if (!this.isHiddenToPlayer() && this.m.Troops.len() != 0 && this.getFaction() != 0)
 		{
 			ret.push({
 				id = 100,
 				type = "text",
 				icon = "ui/icons/icon_contract_swords.png",
-				text = format("Strength: %s / %s", ::MSU.Text.colorGreen(::World.State.getPlayer().getStrength()), ::MSU.Text.colorRed(this.getStrength()))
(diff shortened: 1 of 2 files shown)


Uberbagel (approved):

LordMidas:
With some testing I've seen that sometimes it doesn't show the tooltip of the party and I get `the index Cost does not exist` error in the log. Apparently some troops exist in a party without a `.Cost`? I will try to find out what the cause for this could be.

mod_reforged/hooks/entity/world/location.nut
```diff
@@ -20,13 +20,27 @@
...
+		}
+
 		if (!this.isHiddenToPlayer() && this.m.Troops.len() != 0 && this.getFaction() != 0)
 		{
 			ret.push({
 				id = 100,
 				type = "text",
 				icon = "ui/icons/icon_contract_swords.png",
-				text = format("Strength: %s / %s", ::MSU.Text.colorGreen(::World.State.getPlayer().getStrength()), ::MSU.Text.colorRed(this.getStrength()))
+				text = format("Strength: %s / %s", ::MSU.Text.colorGreen(::World.State.getPlayer().getStrength()), ::MSU.Text.colorRed(worth))
```
Darxo, line 43: `Strength` is now misleading. Only the first value is strength (player party strength)

I suggest changing it into
`Cost: %s (Player Strength: %s)`

After all, we are viewing a different party so the assumption would be that the first/main value is about that party and not us

mod_reforged/hooks/entity/world/party.nut
```diff
@@ -2,13 +2,27 @@
...
+		}
+
 		if (!this.isHiddenToPlayer() && this.m.Troops.len() != 0)
 		{
 			ret.push({
 				id = 100,
 				type = "text",
 				icon = "ui/icons/icon_contract_swords.png",
-				text = format("Strength: %s / %s", ::MSU.Text.colorGreen(::World.State.getPlayer().getStrength()), ::MSU.Text.colorRed(this.getStrength()))
+				text = format("Strength: %s / %s", ::MSU.Text.colorGreen(::World.State.getPlayer().getStrength()), ::MSU.Text.colorRed(worth))
```
Darxo, line 25: Same comment as above

Darxo (changes requested):

mod_reforged/hooks/entity/world/location.nut
LordMidas: Yeah I tried at first to write it as `Player Strength: %s / Target Resources: %s` or something like that but it becomes too long and goes on two lines and I want to avoid that. So then I went with just this even though it isn't really Strength.

But I agree I don't like this either. So, maybe something like `Cost: %s (Player: %s)` is enough?

Darxo: That ok

LordMidas:
Closing as we are approaching public release and this display will be removed from the tooltip anyway.

PR #427, opened 2024-06-08
refactor: rework bolster perk to grant resolve on morale checks

LordMidas:


Darxo (commented):
I dislike how opaque the effect is. There is no tooltip that shows the bonus someone is actually emitting and the bonus someone else is receiving.
And the morale checks themselves also don't show any roles.

From a design pov I would rather pivot to a simpler design for such an effect which can actually be verified by the player.

PR #428, opened 2024-06-08
balance: bloodlust to recover fatigue instead of reducing fatigue cost

LordMidas:
Instead of a flat Fatigue Recovery bonus e.g. +3 per stack, I went with a % based bonus so the entire perk is much more clean this way. 50% increased Fatigue Recovery means for a bro with 15 base recovery you get 7 additional Fatigue recovery next turn. And this is fine imo.


Uberbagel (approved):

Darxo (changes requested):
The perk/skill should clarify, that you lose stacks at the start of your turn and not at the end of your turn. 
Almost every buff/debuff in BB is removed at the end of a characters turn.

Darxo (approved):

PR #430, opened 2024-06-09
refactor: add bolas to tip of the day and mention shield wall exception

Darxo:


mod_reforged/hooks/config/tip_of_the_day.nut
```diff
@@ -15,6 +15,9 @@ for (local index = (::Const.TipOfTheDay.len() - 1); index >= 0; index--)
         case "Losing is fun.":
         case "Some people will use you and throw you away.":
             ::Const.TipOfTheDay.remove(index);
+			break;
+		case "Flails ignore the defense bonus of shields.":
+			::Const.TipOfTheDay[index] = "Flails and Bolas ignore only the defense bonus of shields. The defense granted by \'Shieldwall\' still applies.";
```
LordMidas, line 20: Would be good to make it shorter so it's faster to read on a loading screen. Something like `Flails and Bolas ignore the defense granted by shields, but not by \'Shieldwall\'`.

mod_reforged/hooks/config/tip_of_the_day.nut
```diff
@@ -15,6 +15,9 @@ for (local index = (::Const.TipOfTheDay.len() - 1); index >= 0; index--)
         case "Losing is fun.":
         case "Some people will use you and throw you away.":
             ::Const.TipOfTheDay.remove(index);
+			break;
+		case "Flails ignore the defense bonus of shields.":
```
LordMidas, line 19: Please add a `break;` after it for completion.

LordMidas (changes requested):

mod_reforged/hooks/config/tip_of_the_day.nut
```diff
@@ -15,6 +15,10 @@ for (local index = (::Const.TipOfTheDay.len() - 1); index >= 0; index--)
...
+			::Const.TipOfTheDay[index] = "Flails and Bolas ignore the defense granted by shields, but not by \'Shieldwall\'";
```
LordMidas, line 20: I just realized that we somehow need to convey here that the `Bola` thing is a Reforged-only thing i.e. add `Reforged: ` at the start. But we don't want to imply that the `Shieldwall` thing is a Reforged-only thing. How about `Reforged: Similar to Flails, Bolas also ignore the defense granted by shields, but not by \'Shieldwall\'.`?

LordMidas (changes requested):

PR #432, opened 2024-06-15
refactor: add tip of the day for dismissing brothers

Darxo:
This tip is supposed to emphasize that this happens automatically. 
Otherwise a new player might start stripping those characters every time they are dismissed.

### DIFF
--- a/mod_reforged/hooks/config/tip_of_the_day.nut
+++ b/mod_reforged/hooks/config/tip_of_the_day.nut
@@ -19,6 +19,7 @@ for (local index = (::Const.TipOfTheDay.len() - 1); index >= 0; index--)
 }
 
 ::Const.TipOfTheDay.extend([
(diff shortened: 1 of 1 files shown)


mod_reforged/hooks/config/tip_of_the_day.nut
```diff
@@ -19,6 +19,7 @@ for (local index = (::Const.TipOfTheDay.len() - 1); index >= 0; index--)
 }
 
 ::Const.TipOfTheDay.extend([
+	"Dismissing a brother will transfer all of their equipped items into your stash if you have space."
```
LordMidas, line 22: This is unnecessary imo as this information is present inside the Dismiss dialogue box. So the first time a new player sees that box after clicking Dismiss they will learn it.

image

LordMidas (changes requested):

Issue #281, opened 2024-03-02
Schrats display 2 Uproot skills in their combat tooltip


Darxo:
![image](https://github.com/Battle-Modders/mod-reforged/assets/2252464/38856906-7a6f-4d1d-a758-b4cd233036fd)

![image](https://github.com/Battle-Modders/mod-reforged/assets/2252464/a862471c-1a07-4d7d-a66a-e2643b376b64)

Maybe the solution here is to introduce a new boolean for skills which decides whether they are shown/hidden in the combat tooltip.

Darxo:
The same is true for Unholds:
![image](https://github.com/Battle-Modders/mod-reforged/assets/2252464/560024c8-7c8b-42ef-8df2-c517e02d5cc0)

LordMidas:
Should be resolved as of #452

PR #434, opened 2024-06-17
feat hide some active skills in tooltip

Darxo:
- add new flag for preventing any skill from showing up in combat tooltips
- turn that flag on for zoc versions of sweep, tail slam and uproot
- fix: enforce the skill order for displaying active skills in the tooltip

image
image
image

Closes: #281


Darxo:
The next skill in line to be hidden from tooltips is `rf_en_garde_toggle_skill`. I feel like this active is just bloat and I don't think enemies will toggle this anyways.

LordMidas:
This is definitely a possible solution. However, I would like to discuss this further in the tech team first e.g. another solution could be to exclude all skills for which the entity does not have an AI behavior.

LordMidas:
This can now be closed as we did it in an alternative approach (#452).

PR #438, opened 2024-06-22
refactor: add new log and isHiddenToPlayer logic to command skill

Darxo:
- Anyone using command skill now creates a default log (like for example using Shieldwall)
- Enemies using command skill outside of your view no longer produces logs

### DIFF
--- a/scripts/skills/actives/rf_command_skill.nut
+++ b/scripts/skills/actives/rf_command_skill.nut
@@ -76,11 +76,24 @@ this.rf_command_skill <- this.inherit("scripts/skills/skill", {
 		local target = _targetTile.getEntity();
 		::Tactical.TurnSequenceBar.moveEntityToFront(target.getID());
 
+		if (!_user.isHiddenToPlayer())
+		{
+			local logText = ::Const.UI.getColorizedEntityName(_user) + " uses Command";
+			if (!target.isHiddenToPlayer())
+			{
+				logText += " on " + ::Const.UI.getColorizedEntityName(target);
+			}
+			::Tactical.EventLog.log(logText);
+		}
+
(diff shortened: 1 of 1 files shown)


scripts/skills/actives/rf_command_skill.nut
```diff
@@ -75,12 +75,19 @@ this.rf_command_skill <- this.inherit("scripts/skills/skill", {
 	{
 		local target = _targetTile.getEntity();
 		::Tactical.TurnSequenceBar.moveEntityToFront(target.getID());
+		if (!_user.isHiddenToPlayer())
```
LordMidas, line 78: Shouldn't this require both the user and the target to be visible? That seems consistent with other log entries e.g. for vanilla attacks.

LordMidas (changes requested):

scripts/skills/actives/rf_command_skill.nut
Darxo: goblin night vision spell only requires the user to be visible
image

possess does not care who is visible
image

I prefer when the player has to fight a bit more for such information on a battle. And that means discovering certain enemies before they get such information about them. So I like the name of the caster to not be spoiled in the log.

From another POV you can think of it that you (the battle brothers) don't hear the command because the commander is too far away. And the log just represents what you see/hear/smell/feel.

LordMidas: Yes that's fine. But in your implementation you show both the user's name and the target's name even when only the user is visible. I suggest to have two conditions:
1. If only user is visible then it should say `user uses command`
2. If both user and target are visible then `user uses command on target`.

LordMidas (approved):

PR #439, opened 2024-06-23
refactor: add sound effects for adjusting dented armor

Darxo:
These are the same sfx used for when you repair gear in towns.

It may be inrealistic how "adjusting armor", and no less quickly and on the battlefield causes smithing sfx. But I find it a completely find sfx as it's also not used by other skills in combat.

### DIFF
--- a/scripts/skills/actives/rf_adjust_dented_armor_ally_skill.nut
+++ b/scripts/skills/actives/rf_adjust_dented_armor_ally_skill.nut
@@ -8,7 +8,14 @@ this.rf_adjust_dented_armor_ally_skill <- ::inherit("scripts/skills/skill", {
 		this.m.Icon = "skills/rf_adjust_dented_armor_ally_skill.png";
 		this.m.IconDisabled = "skills/rf_adjust_dented_armor_ally_skill_bw.png";
 		this.m.Overlay = "rf_adjust_dented_armor_ally_skill";
-		this.m.SoundOnHit = [];
+		this.m.SoundOnUse = [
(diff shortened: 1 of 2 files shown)


scripts/skills/actives/rf_adjust_dented_armor_ally_skill.nut
LordMidas, line 1: I think the blacksmith sounds are too out of place for the combat and what the skill represents. How about `sounds/ambience/settlement/fortification_armor_weapons_00.wav` and its variations? I think that sounds more appropriate.

LordMidas (changes requested):

scripts/skills/actives/rf_adjust_dented_armor_ally_skill.nut
```diff
@@ -8,7 +8,14 @@ this.rf_adjust_dented_armor_ally_skill <- ::inherit("scripts/skills/skill", {
 		this.m.Icon = "skills/rf_adjust_dented_armor_ally_skill.png";
 		this.m.IconDisabled = "skills/rf_adjust_dented_armor_ally_skill_bw.png";
 		this.m.Overlay = "rf_adjust_dented_armor_ally_skill";
-		this.m.SoundOnHit = [];
+		this.m.SoundOnUse = [
+			"sounds/ambience/settlement/fortification_armor_weapons_00",
```
LordMidas, line 12: I believe you need to add `.wav` at the end of the path for all sounds.

LordMidas (approved):
After adding `.wav` to all sound paths, looks good to merge.

PR #440, opened 2024-06-25
fix: suppporter crashing when actor is not placed on map

LordMidas:
E.g. during the use of skills like Rotation.


scripts/skills/perks/perk_rf_supporter.nut
```diff
@@ -37,27 +38,30 @@ this.perk_rf_supporter <- this.inherit("scripts/skills/skill", {
...
-		{
 			return;
-		}
 
 		local actor = this.getContainer().getActor();
-		if (actor.getFaction() == _targetEntity.getFaction() && actor.getID() != _targetEntity.getID())
+		if (actor.getFaction() != _targetEntity.getFaction() || actor.getID() == _targetEntity.getID() || !actor.isPlacedOnMap() || actor.getTile().getDistanceTo(_targetTile) > this.m.MinDistanceAPRecovery)
+			return;
+
+		this.m.RecoveredActionPoints = ::Math.min(actor.getActionPointsMax() - actor.getActionPoints(), this.m.ActionPointsRecovered);
```
Darxo, line 53: By this time the character has not yet spent any action points, so this will often times recover no action points.

scripts/skills/perks/perk_rf_supporter.nut
```diff
@@ -5,7 +5,8 @@ this.perk_rf_supporter <- this.inherit("scripts/skills/skill", {
 		ActionPointsRecovered = 2,
 
 		// Private
-		IsSpent = false
+		IsSpent = false,
+		RecoveredActionPoints = 0
```
Darxo, line 9: I suggest turning this into `WillRecoverActionPoints = false`.
The logic further down will be similar that you check the adjacency condition before you execute the skill. But you do not calculate the recovered action points by that time because it won't be accurate/as expected

Darxo (changes requested):

Darxo (approved):

PR #442, opened 2024-07-05
refactor: show threshold fatigue value in fresh and furious tooltip

LordMidas:


scripts/skills/perks/perk_rf_fresh_and_furious.nut
```diff
@@ -55,12 +55,25 @@ this.perk_rf_fresh_and_furious <- ::inherit("scripts/skills/skill", {
...
+				text = ::Reforged.Mod.Tooltips.parseString("Becomes disabled when starting a turn with " + ::MSU.Text.colorizeFraction(this.m.FatigueThreshold) + " or more [Fatigue|Concept.Fatigue] built")
+			});
+		}
+		else
+		{
+			tooltip.push({
+				id = 12,
+				type = "text",
+				icon = "ui/icons/warning.png",
+				text = ::Reforged.Mod.Tooltips.parseString(format("Becomes disabled when starting a turn with %s (%s) or more [Fatigue|Concept.Fatigue] built", ::MSU.Text.colorizeFraction(this.m.FatigueThreshold), ::MSU.Text.colorRed(::Math.round(this.m.FatigueThreshold * this.getContainer().getActor().getFatigue()))))
```
Darxo, line 74: The first value is colored green while value in brackets is colored red. You should use the same color on both.

image

In the perk tooltip we use red color for the threshold. Here we used green color previously.
I'd suggest `colorNeutral` if it was an option. Otherwise just choose one of the two colors for all

Darxo (changes requested):

scripts/skills/perks/perk_rf_fresh_and_furious.nut
LordMidas: It's supposed to be red. If it was being shown green previously that's unintended. Will fix.

Darxo (approved):

PR #446, opened 2024-07-12
Known issues fixes

LordMidas:


scripts/skills/perks/perk_rf_offhand_training.nut
```diff
@@ -76,23 +76,12 @@ this.perk_rf_offhand_training <- ::inherit("scripts/skills/skill", {
...
-				 if (i.getID().find("buckler") != null)
-				 {
-					 validItemsCount++;
-				 }
-				 else
-				 {
-					 return null;
-				 }
-			}
+			validItemsCount++;
```
Darxo, line 84: You might aswell `return 0` here to simplify the logic of this function

mod_reforged/hooks/config/strings.nut
```diff
@@ -1769,7 +1769,7 @@ foreach (vanillaDesc in vanillaDescriptions)
 			Type = ::UPD.EffectType.Passive,
 			Description = [
 				"Once per [turn|Concept.Turn], the first use of the tool in your offhand costs no [Action Points|Concept.ActionPoints].",
-				"Once per [turn|Concept.Turn], switching a buckler or tool in your offhand costs no [Action Points|Concept.ActionPoints]. Does not stack with other free item swapping skills e.g. Quick Hands.",
+				"Once per [turn|Concept.Turn], switching offhand items with weight less than " + ::MSU.Text.colorRed("10") + " costs no [Action Points|Concept.ActionPoints]. Does not stack with other free item swapping skills e.g. Quick Hands.",
```
Darxo, line 1772: This still doesn't specify that two-handed weapons can be switched but not regular shields.

"Switching offhand items ..." sounds to me like this effect only allows you to switch two offhand items.

Darxo (changes requested):

scripts/skills/perks/perk_rf_offhand_training.nut
LordMidas: No because we have to iterate over all the items to ensure that the previous condition i.e. `i.getStaminaModifier() <= -10` isn't being applied to any item.

mod_reforged/hooks/config/strings.nut
LordMidas: How about `switching to or from an offhand item with less than 10 weight`?

LordMidas: I guess that still leaves room for an item being less than 10 weight and the other being heavier. So it doesn't seem accurate. Maybe something like `switching to and from offhand items with less than 10 weight` is clear?

Darxo: My proposal:
`"Once per [turn|Concept.Turn], switching an offhand item costs no [Action Points|Concept.ActionPoints]. 
Does not work when switching offhand items with a weight of " + ::MSU.Text.colorRed("10") + " or more.
Does not stack with other free item swapping skills e.g. Quick Hands.",`

LordMidas: Looks good, I'll use that.

Darxo: You requested a review but I can't see this change in the commits yet

LordMidas: > You requested a review but I can't see this change in the commits yet

It appears to me. Maybe you were looking at the wrong page?

Darxo (approved):

PR #447, opened 2024-07-12
refactor: require attack during turn for unstoppable to gain stack

LordMidas:


scripts/skills/perks/perk_rf_unstoppable.nut
```diff
@@ -76,7 +76,6 @@ this.perk_rf_unstoppable <- ::inherit("scripts/skills/skill", {
 	{
 		if (!this.m.IsGainingStackThisTurn)
 		{
-			this.m.IsGainingStackThisTurn = true;
```
Darxo, line 79: this needs to start at `false` on each turn now. 
So at some point during `onTurnEnd` or `onTurnStart` you need to reset it back to `false`

Darxo (changes requested):

Darxo (approved):

PR #450, opened 2024-07-13
feat: add descriptions for all lore keeper active skills

Darxo:


Uberbagel (approved):
I can't speak to the amount of information we want to give players about these skills, but all of the texts and descriptions I otherwise approve of.

PR #452, opened 2024-07-14
refactor: hide active skills on npcs for which they dont have behavior

LordMidas:


mod_reforged/actor_tooltip_functions.nut
```diff
@@ -271,6 +271,24 @@
 	local ret = [];
 
 	local skills = _actor.getSkills().getAllSkillsOfType(::Const.SkillType.Active);
+	if (!_actor.isPlayerControlled())
```
Darxo, line 274: I suggest adding a comment for this block summarizing what is being done. Could be a commend in front of this if with an empty line above it.

Darxo, line 274: I wonder if `isPlayerControlled` is future proof. 
I can see situations with player animals that have skills which they won't use via behaviors but which can be triggered via a player skill (e.g. an animal command) or which can be used once the player takes over control via a submod or future reforged integration.
Infact if the player takes over control over any actor which is not from the player faction, he will not be able to see actives in those tooltips even though he can use them.

Darxo (changes requested):

mod_reforged/actor_tooltip_functions.nut
LordMidas: > I can see situations with player animals that have skills which they won't use via behaviors but which can be triggered via a player skill (e.g. an animal command)
This is definitely a concern. We can exclude PlayerAnimals faction from this treatment.

> Infact if the player takes over control over any actor which is not from the player faction, he will not be able to see actives in those tooltips even though he can use them.
This is not the case. As far as I know, `isPlayerControlled()` returns true for any character that the player controls irrespective of the character's faction.

Darxo: I heard and saw situations where player can control their unleashed animals via mods. Are you sure those animals then change faction to become player faction for this time being? Because that is a requirement for `isPlayerControlled` to return true.

Darxo: I feel that just checking for `if (_actor.m.IsControlledByPlayer)` is more future proof even though it looks worse than calling a function. 
Maybe @Enduriel has some input, he has worked with a multiplayer system and I remember him running into an issue around isPlayerControlled.
Or @TaroEld who has done work with the combat simulator which also deals with non player faction being controlled by player.

LordMidas: > Because that is a requirement for isPlayerControlled to return true.

Ah yes, you're right. The `this.m.IsControlledByPlayer` sounds like the correct option. Let's see what Enduriel and Taro say.

Enduriel: Oh wow I super don't like this, it would break any potential multiplayer compatibility and probably cause other problems as skills are being *removed*. I think this is a fundamentally flawed solution and the PR title is tbh misleading.

LordMidas:
Lol Endy, skills are being removed from the actor tooltip. Not from the
actor himself.

mod_reforged/actor_tooltip_functions.nut
LordMidas: Endy, the skills are being removed from our actor tooltip. Not from the actor himself.

Darxo (approved):

PR #455, opened 2024-07-20
refactor: rework offhand training

LordMidas:
- Remove free swap.
- Instead of only tool, extend the one free use per turn to all offhand items less than 10 weight.

### DIFF
--- a/mod_reforged/hooks/config/strings.nut
+++ b/mod_reforged/hooks/config/strings.nut
@@ -1768,8 +1768,7 @@ foreach (vanillaDesc in vanillaDescriptions)
 		Effects = [{
 			Type = ::UPD.EffectType.Passive,
 			Description = [
-				"Once per [turn,|Concept.Turn] the first use of the tool in your offhand costs no [Action Points.|Concept.ActionPoints]",
-				"Once per [turn,|Concept.Turn] switching an offhand item costs no [Action Points.|Concept.ActionPoints] Does not work when switching items with a weight of " + ::MSU.Text.colorNegative("10") + " or more. Does not stack with other free item swapping skills e.g. Quick Hands.",
+				"Once per [turn,|Concept.Turn] the first use of your offhand item weighing less than " + ::MSU.Text.colorRed(10) + " costs no [Action Points.|Concept.ActionPoints]",
 				"When equipped with a net, gain the [Trip Artist|NullEntitySkill+rf_trip_artist_effect] effect."
 			]
 		}]
--- a/scripts/skills/perks/perk_rf_offhand_training.nut
+++ b/scripts/skills/perks/perk_rf_offhand_training.nut
@@ -1,8 +1,8 @@
 this.perk_rf_offhand_training <- ::inherit("scripts/skills/skill", {
 	m = {
 		StaminaModifierThreshold = -10,
-		IsFreeUse = false,
-		IsConsumingFreeUse = false // Used in onBeforeAnySkillExecuted to flip IsFreeUse in onAnySkillExecuted
+		IsSpent = true,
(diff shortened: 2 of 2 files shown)


Uberbagel:
Does the one free use at less than 10 weight count for shieldwall?

LordMidas:
> Does the one free use at less than 10 weight count for shieldwall?

Yes

Uberbagel:
> > Does the one free use at less than 10 weight count for shieldwall?
> 
> Yes

I'm somewhat concerned at the strength of this but it's worth trying out.

Uberbagel (approved):

scripts/skills/perks/perk_rf_offhand_training.nut
```diff
@@ -21,34 +20,19 @@ this.perk_rf_offhand_training <- ::inherit("scripts/skills/skill", {
...
-			});
-		}
-
-		return tooltip;
+		local ret = this.skill.getTooltip();
+		ret.push({
+			id = 10,
+			type = "text",
+			icon = "ui/icons/special.png",
+			text = ::Reforged.Mod.Tooltips.parseString("The first use of any offhand item weighing less than " + ::MSU.Text.colorNegative(-this.m.StaminaModifierThreshold) + " this [turn|Concept.Turn] costs no [Action Points|Concept.ActionPoints]")
```
Darxo, line 33: Now that this extends to all skills provided by that offhand item, we should mention that it in the tooltip explicitely.
`The first skill use of any offhand item [...]`

mod_reforged/hooks/config/strings.nut
```diff
@@ -1768,8 +1768,7 @@ foreach (vanillaDesc in vanillaDescriptions)
  		Effects = [{
 			Type = ::UPD.EffectType.Passive,
 			Description = [
-				"Once per [turn|Concept.Turn], the first use of the tool in your offhand costs no [Action Points|Concept.ActionPoints].",
-				"Once per [turn|Concept.Turn], switching an offhand item costs no [Action Points|Concept.ActionPoints]. Does not work when switching items with a weight of " + ::MSU.Text.colorRed("10") + " or more. Does not stack with other free item swapping skills e.g. Quick Hands.",
+				"Once per [turn|Concept.Turn], the first use of your offhand item weighing less than " + ::MSU.Text.colorRed(10) + " costs no [Action Points|Concept.ActionPoints].",
```
Darxo, line 1771: Now that this extends to all skills provided by that offhand item, we should mention that it in the tooltip explicitely.
`[...] the first skill use of your offhand item [...]`

Darxo (changes requested):

scripts/skills/perks/perk_rf_offhand_training.nut
LordMidas: I think that's not necessary and makes it too mechanical in my opinion. I prefer the current wording in the PR i.e. `the first use of any offhand item` - the fact that it refers to that item's skills is implicit.

scripts/skills/perks/perk_rf_offhand_training.nut
```diff
@@ -1,7 +1,6 @@
 this.perk_rf_offhand_training <- ::inherit("scripts/skills/skill", {
 	m = {
 		StaminaModifierThreshold = -10,
-		IsSpent = true,
```
Darxo, line 4: I prefer the term `IsSpent` for the only variable enforcing the main behavior of this skill.
It makes the code easier to parse and is already used for similar effects in other skills plenty of times
e.g.
`return this.m.IsSpent`
or 
```
if (this.m.IsSpent)
  return;
```

Darxo (changes requested):

Darxo (approved):

PR #456, opened 2024-07-20
feat: scale xp received by bros with ratio of target xp to bro xp value

LordMidas:
This increases the reward for taking down stronger opponents with your company and reduces the reward for taking down weaker opponents.


Uberbagel:
I pretty strongly dislike this feature for reasons I've stated in the 'Incentivizing Risk' thread

PR #461, opened 2024-07-29
refactor: string, description and tooltip improvements

Uberbagel:
I've made a large number of small changes during my work updating the dev sheet that I think improve readability and style. Comment where you disagree with my changes please.

### DIFF
--- a/mod_reforged/hooks/config/strings.nut
+++ b/mod_reforged/hooks/config/strings.nut
@@ -252,7 +252,7 @@ local vanillaDescriptions = [
 			Effects = [{
 				Type = ::UPD.EffectType.Passive,
 				Description = [
-					"Reduce the [Hitpoint|Concept.Hitpoints] damage threshold for triggering [morale checks|Concept.Morale] with your attacks to 1.",
+					"The damage threshold for triggering [morale checks|Concept.Morale] with your attacks is reduced to 1.",
 					"These morale checks now have an additional penalty to [Resolve|Concept.Bravery] of " + ::MSU.Text.colorPositive("20%") + " of your [Resolve.|Concept.Resolve]"
 				]
 			}]
@@ -301,7 +301,7 @@ local vanillaDescriptions = [
 				Type = ::UPD.EffectType.Active,
 				Description = [
 					"Unlocks the [Adrenaline|Skill+adrenaline_skill] skill which puts you first in the [turn|Concept.Turn] order for the next [round,|Concept.Round] to have another [turn|Concept.Turn] before your enemies do.",
-					"While [Adrenaline|Skill+adrenaline_effect] is active you cannot receive and are not affected by [temporary injuries.|Concept.InjuryTemporary]"
+					"Under the effects of [Adrenaline|Skill+adrenaline_effect] you cannot receive and are not affected by [temporary injuries.|Concept.InjuryTemporary]"
 				]
 			}]
 		})
@@ -447,7 +447,7 @@ local vanillaDescriptions = [
 			Effects = [{
 				Type = ::UPD.EffectType.Passive,
 				Description = [
-					"When no ally of your faction is adjacent to you, gain a " + ::MSU.Text.colorPositive("+15%") + " bonus to [Melee Skill,|Concept.MeleeSkill] [Ranged Skill,|Concept.RangeSkill] [Melee Defense,|Concept.MeleeDefense] [Ranged Defense,|Concept.RangeDefense] and [Resolve.|Concept.Bravery]"
+					"When no ally of your faction is adjacent to you, gain a " + ::MSU.Text.colorPositive("15%") + " bonus to [Melee Skill,|Concept.MeleeSkill] [Ranged Skill,|Concept.RangeSkill] [Melee Defense,|Concept.MeleeDefense] [Ranged Defense,|Concept.RangeDefense] and [Resolve.|Concept.Bravery]"
 				]
 			}]
 		})
@@ -541,7 +541,7 @@ local vanillaDescriptions = [
 				{
 					Type = ::UPD.EffectType.Active,
 					Description = [
-						"Unlocks the [Take Aim|Skill+rf_take_aim_skill] skill which allows you to target opponents behind obstacles with a [crossbow,|Item+crossbow] or hit more targets with a [handgonne.|Item+handgonne]"
+						"Unlocks the [Take Aim|Skill+rf_take_aim_skill] skill which allows you to target opponents behind obstacles with a [crossbow|Item+crossbow] or hit more targets with a [handgonne.|Item+handgonne]"
 					]
 				},
 
@@ -847,8 +847,8 @@ local vanillaDescriptions = [
 			Effects = [{
 				Type = ::UPD.EffectType.Passive,
 				Description = [
-					"Once per battle, upon receiving a killing blow, survive instead with a few hitpoints left and have all damage over time effects (e.g. bleeding, poisoned) cured.",
-					"When this effect triggers, gain increased stats until your next turn"
+					"Once per battle, upon receiving a killing blow, survive instead with a few [Hitpoints|Concept.Hitpoints] left and have all damage over time effects (e.g. [bleeding|Skill+bleeding_effect], [poisoned|Skill+spider_poison_effect]) cured.",
+					"Upon triggering, gain increased [attributes|Concept.CharacterAttribute] until your next [turn.|Concept.Turn]"
 				]
 			}]
 		})
@@ -907,7 +907,7 @@ foreach (vanillaDesc in vanillaDescriptions)
 			Type = ::UPD.EffectType.Passive,
 			Description = [
 				"[Resolve|Concept.Bravery] is increased by " + ::MSU.Text.colorPositive("10%") + " at all times.",
-				"Additionally, for positive morale checks, [Resolve|Concept.Bravery] is increased by a further " + ::MSU.Text.colorPositive("10%") + ".",
+				"Additionally, at positive morale checks, [Resolve|Concept.Bravery] is increased by a further " + ::MSU.Text.colorPositive("10%") + ".",
 				"When at Confident Morale, all the bonuses of this perk are doubled, and [Melee Skill,|Concept.MeleeSkill] [Ranged Skill,|Concept.RangeSkill] [Melee Defense,|Concept.MeleeDefense] and [Ranged Defense|Concept.RangeDefense] are increased by " + ::MSU.Text.colorPositive("5%") + "."
 			]
 		}]
@@ -999,7 +999,7 @@ foreach (vanillaDesc in vanillaDescriptions)
 		Effects = [{
 			Type = ::UPD.EffectType.Passive,
 			Description = [
-				"Attacks against [surrounded|Concept.Surrounding] targets deal " + ::MSU.Text.colorPositive("10%") + " more damage and have " + ::MSU.Text.colorPositive("+10%") + " armor penetration per ally surrounding the target."
+				"Attacks against [surrounded|Concept.Surrounding] targets deal " + ::MSU.Text.colorPositive("10%") + " more damage and have " + ::MSU.Text.colorPositive("+10%") + " armor penetration per character surrounding the target."
 			]
 		}]
 	}),
@@ -1091,7 +1091,7 @@ foreach (vanillaDesc in vanillaDescriptions)
 		Effects = [{
 			Type = ::UPD.EffectType.Passive,
 			Description = [
-				"Attacks against characters with a lower [morale|Concept.Morale] than you deal " + ::MSU.Text.colorPositive("10%") + " increased damage per level of [morale|Concept.Morale] difference."
+				"Attacks against characters with a lower [morale|Concept.Morale] than you deal " + ::MSU.Text.colorPositive("10%") + " more damage per level of [morale|Concept.Morale] difference."
 			]
 		}]
 	}),
@@ -1140,8 +1140,8 @@ foreach (vanillaDesc in vanillaDescriptions)
 		Effects = [{
 			Type = ::UPD.EffectType.Passive,
 			Description = [
-				"Hits to the head with one-handed maces apply the [Dazed|Skill+dazed_effect] effect for 1 [turn|Concept.Turn] and if the target is already [Dazed,|Skill+dazed_effect] apply [stunned|Skill+stunned_effect] for one [turn.|Concept.Turn]",
-				"Hits to the head with two-handed maces apply [stunned|Skill+stunned_effect] for 1 [turn.|Concept.Turn]"
+				"Hits to the head with one-handed maces apply the [dazed|Skill+dazed_effect] effect for 1 [turn|Concept.Turn] and if the target is already [dazed,|Skill+dazed_effect] apply [stunned|Skill+stunned_effect] for one [turn.|Concept.Turn]",
+				"Hits to the head with two-handed maces [stun|Skill+stunned_effect] the target for 1 [turn.|Concept.Turn]"
 			]
 		}]
 	}),
@@ -1183,7 +1183,7 @@ foreach (vanillaDesc in vanillaDescriptions)
 				"Gain a stack whenever you end your [turn|Concept.Turn] without having used [Wait,|Concept.Wait] up to a maximum of 3 stacks.",
 				"Gain " + ::MSU.Text.colorPositive("+15") + " [Resolve|Concept.Bravery] and [Initiative|Concept.Initiative] while you have at least 1 stack.",
 				"Skills build up " + ::MSU.Text.colorPositive("15%") + " less [Fatigue|Concept.Fatigue] while you have at least 2 stacks.",
-				"Deal " + ::MSU.Text.colorPositive("15%") + " additional damage while you have 3 stacks.",
+				"Deal " + ::MSU.Text.colorPositive("15%") + " more damage while you have 3 stacks.",
 				"You lose all stacks if you use [Wait.|Concept.Wait]"
 			]
 		}]
@@ -1259,7 +1259,7 @@ foreach (vanillaDesc in vanillaDescriptions)
 			Type = ::UPD.EffectType.Passive,
 			Description = [
 				"A successful hit increases the damage of your next attack by " + ::MSU.Text.colorNegative("20%") + ".",
-				"The effect is lost upon moving, swapping an item, using any skill except a single-target attack, missing an attack, or waiting or ending your [turn.|Concept.Turn]"
+				"The effect is lost upon moving, swapping an item, using any skill except a single-target attack, missing an attack, or [waiting|Concept.Wait] or ending your [turn.|Concept.Turn]"
 			]
 		}]
 	}),
@@ -1345,7 +1345,7 @@ foreach (vanillaDesc in vanillaDescriptions)
 				Type = ::UPD.EffectType.Passive,
 				Description = [
 					"Always start battles at Confident [morale.|Concept.Morale]",
-					"Gain " ::MSU.Text.colorPositive("50%") + " increased [Resolve|Concept.Resolve] against negative morale checks."
+					"[Resolve|Concept.Bravery] is increased by " + ::MSU.Text.colorPositive("50%") + " against negative [morale checks.|Concept.Morale]"
 				]
 			},
 			{
@@ -1466,7 +1466,7 @@ foreach (vanillaDesc in vanillaDescriptions)
 			Type = ::UPD.EffectType.Passive,
 			Description = [
 				"When the number of adjacent allies is greater than or equal to the number of adjacent enemies, you may ignore [Zone of Control|Concept.ZoneOfControl] for your next movement action."
-				"When wielding a melee weapon with up to 4 [Reach|Concept.Reach] when you end your movement adjacent to an enemy, the next attack against that enemy deals " + ::MSU.Text.colorPositive("25%") + " additional damage and " + ::MSU.Text.colorPositive("+20%") + " damage ignoring armor. This bonus expires upon using any action other than an attack."
+				"When wielding a melee weapon with up to 4 [Reach|Concept.Reach] when you end your movement adjacent to an enemy, the next attack against that enemy deals " + ::MSU.Text.colorPositive("25%") + " more damage and " + ::MSU.Text.colorPositive("+20%") + " damage ignoring armor. This bonus expires upon using any action other than an attack."
 			]
 		}]
 	}),
@@ -1500,7 +1500,7 @@ foreach (vanillaDesc in vanillaDescriptions)
 			{
 				Type = ::UPD.EffectType.Active,
 				Description = [
-					"Unlocks the [Pocket Sand|Skill+rf_pocket_sand_skill] skill which allows you to store sand in your empty [bag slots|Concept.BagSlots] with which to distract nearby enemies."
+					"Unlocks the [Pocket Sand|Skill+rf_pocket_sand_skill] skill which allows you to store sand in your empty [bag slots|Concept.BagSlots] to distract nearby enemies with."
 				]
 			}
 		]
@@ -1548,7 +1548,7 @@ foreach (vanillaDesc in vanillaDescriptions)
 			{
 				Type = ::UPD.EffectType.Active,
 				Description = [
-					"Gain the [Encourage|Skill+rf_encourage_skill] skill that allows you to increase the [morale|Concept.Morale] of an ally."
+					"Gain the [Encourage|Skill+rf_encourage_skill] skill which allows you to increase the [morale|Concept.Morale] of an ally."
 				]
 			}
 		]
@@ -1612,7 +1612,7 @@ foreach (vanillaDesc in vanillaDescriptions)
 		Effects = [{
 			Type = ::UPD.EffectType.Passive,
 			Description = [
-				"While wielding a [net,|Item+throwing_net] every successful melee attack against an adjacent target has a chance to [net|Skill+net_effect] the target without expending your currently held [net.|Item+throwing_net] The chance is equal to the hit chance of the attack.",
+				"While holding a [net,|Item+throwing_net] every successful melee attack against an adjacent target has a chance to [net|Skill+net_effect] the target without expending your currently held [net.|Item+throwing_net] The chance is equal to the hit chance of the attack.",
 				"You cannot use or swap this [net|Item+throwing_net] until the target breaks free or dies. If you move more than 1 tile away from the target, the target remains [netted|Skill+net_effect] but you lose the [net.|Item+throwing_net]",
 				"The [net effect|Skill+net_effect] can be broken out of with 100% effectiveness.",
 				"Gain " + ::MSU.Text.colorPositive("+2") + " [Reach|Concept.Reach] while holding a [net|Item+throwing_net] and not currently [netting|Skill+net_effect] a target."
@@ -1626,7 +1626,7 @@ foreach (vanillaDesc in vanillaDescriptions)
 			Type = ::UPD.EffectType.Passive,
 			Description = [
 				"The [Action Point|Concept.ActionPoints] cost of spear attacks is reduced by " + ::MSU.Text.colorPositive(1) + ".",
-				"During your turn, the first attack from a spear deals " + ::MSU.Text.colorPositive("25%") + " increased damage."
+				"During your [turn,|Concept.Turn] the first attack from a spear deals " + ::MSU.Text.colorPositive("25%") + " increased damage."
 			]
 		}]
 	}),
@@ -1754,7 +1754,7 @@ foreach (vanillaDesc in vanillaDescriptions)
 		Effects = [{
 			Type = ::UPD.EffectType.Passive,
 			Description = [
-				"The [Action Point|Concept.ActionPoints] cost of reloading a Handgonne is reduced by " + ::MSU.Text.colorPositive(2) + ".",
+				"The [Action Point|Concept.ActionPoints] cost of reloading a [handgonne|Item+handgonne] is reduced by " + ::MSU.Text.colorPositive(2) + ".",
 				"When using a crossbow, for each point of current [Ranged Skill|Concept.RangeSkill] above 90, damage is increased by " + ::MSU.Text.colorPositive("1%") + " up to a maximum of " + ::MSU.Text.colorPositive("30%") + "."
 			]
 		}]
@@ -2028,12 +2028,12 @@ foreach (vanillaDesc in vanillaDescriptions)
 	}),
 	RF_SwordmasterBladeDancer = ::UPD.getDescription({
 		Fluff = "Let's Dance!",
-		Requirement = "Non-Fencing Sword"
+		Requirement = "Sword"
 		Effects = [{
 			Type = ::UPD.EffectType.Passive,
 			Description = [
-				"Gain [Initiative|Concept.Initiative] equal to the armor ignore percentage of your equipped sword.",
-				"The [Action Point|Concept.ActionPoints] cost of non-AOE skills is reduced by " + ::MSU.Text.colorPositive(1) + " and these skills build up " + ::MSU.Text.colorPositive("25%") + " less [Fatigue.|Concept.Fatigue]",
+				"Gain additional [Initiative|Concept.Initiative] equal to the armor ignore percentage of your equipped sword.",
+				"When using a non-fencing sword, the [Action Point|Concept.ActionPoints] cost of non-AOE skills is reduced by " + ::MSU.Text.colorPositive(1) + " and these skills build up " + ::MSU.Text.colorPositive("25%") + " less [Fatigue.|Concept.Fatigue]",
 				"Allows [Kata Step|Skill+rf_kata_step_skill] to be usable even while holding something, e.g. a shield, in your offhand.",
 				"[Kata Step|Skill+rf_kata_step_skill] costs " + ::MSU.Text.colorPositive(2) + " fewer [Action Points|Concept.ActionPoints] and builds " + ::MSU.Text.colorPositive(2) + " less [Fatigue,|Concept.Fatigue] both down to a minimum of 0."
 			]
@@ -2067,7 +2067,7 @@ foreach (vanillaDesc in vanillaDescriptions)
 	}),
 	RF_SwordmasterJuggernaut = ::UPD.getDescription({
 		Fluff = "There\'s a fine line between bravery and stupidity.",
-		Requirement = "Two-Handed Non-Fencing Sword"
+		Requirement = "Two-Handed Sword"
 		Effects = [
 			{
 				Type = ::UPD.EffectType.Active,
@@ -2153,7 +2153,7 @@ foreach (vanillaDesc in vanillaDescriptions)
 			{
 				Type = ::UPD.EffectType.Passive,
 				Description = [
-					"Every hit or miss against any target increases your [Initiative|Concept.Initiative] by " + ::MSU.Text.colorPositive("+15") + ". This bonus lasts over into your next [turn|Concept.Turn] but only until the first skill used or upon waiting that [turn.|Concept.Turn]",
+					"Every hit or miss against any target increases your [Initiative|Concept.Initiative] by " + ::MSU.Text.colorPositive("+15") + ". This bonus is carried over into your next [turn|Concept.Turn] but only until the first skill used or upon [waiting|Concept.Wait] that [turn.|Concept.Turn]",
 					"The first two hits against opponents who act after you in the current [round|Concept.Round] recover " + ::MSU.Text.colorPositive("2") + " [Action Points|Concept.ActionPoints] each. A miss against any target or a hit against a target who acted before you increments the counter, but does not recover any [Action Points.|Concept.ActionPoints]"
 				]
 			}
--- a/mod_reforged/hooks/config/strings.nut
+++ b/mod_reforged/hooks/config/strings.nut
@@ -1140,7 +1140,7 @@ foreach (vanillaDesc in vanillaDescriptions)
 		Effects = [{
 			Type = ::UPD.EffectType.Passive,
 			Description = [
-				"Hits to the head with one-handed maces [daze|Skill+dazed_effect] the target for 1 [turn|Concept.Turn] and if the target is already [dazed,|Skill+dazed_effect] [stun|Skill+stunned_effect] for one [turn instead.|Concept.Turn]",
+				"Hits to the head with one-handed maces [daze|Skill+dazed_effect] the target for 1 [turn|Concept.Turn] and if the target is already [dazed,|Skill+dazed_effect] [stun|Skill+stunned_effect] for one [turn|Concept.Turn] instead.",
 				"Hits to the head with two-handed maces [stun|Skill+stunned_effect] the target for 1 [turn.|Concept.Turn]"
 			]
 		}]
--- a/mod_reforged/hooks/config/strings.nut
+++ b/mod_reforged/hooks/config/strings.nut
@@ -239,7 +239,7 @@ local vanillaDescriptions = [
 				Type = ::UPD.EffectType.Active,
 				Description = [
 					"Unlocks the [Rotation|Skill+rotation] skill which allows you to switch places with an adjacent allied character while ignoring [Zone of Control.|Concept.ZoneOfControl]",
-					"Does not work if either character is [stunned,|Skill+stunned_effect] [rooted|Skill+rooted_effect] or otherwise disabled."
+					"Cannot be used if either character is [stunned,|Skill+stunned_effect] [rooted|Skill+rooted_effect] or otherwise disabled."
 				]
 			}]
 		})
@@ -252,7 +252,7 @@ local vanillaDescriptions = [
 			Effects = [{
 				Type = ::UPD.EffectType.Passive,
 				Description = [
-					"Reduce the [Hitpoint|Concept.Hitpoints] damage threshold for triggering [Morale Checks|Concept.Morale] with your attacks to 1.",
+					"Reduce the [Hitpoint|Concept.Hitpoints] damage threshold for triggering [morale checks|Concept.Morale] with your attacks to 1.",
 					"These morale checks now have an additional penalty to [Resolve|Concept.Bravery] of " + ::MSU.Text.colorPositive("20%") + " of your [Resolve.|Concept.Resolve]"
 				]
 			}]
@@ -286,7 +286,7 @@ local vanillaDescriptions = [
 			Effects = [{
 				Type = ::UPD.EffectType.Active,
 				Description = [
-					"Unlocks the [Rally|Skill+rally_the_troops] skill which can raise [morale|Concept.Morale] of all nearby allies to a steady level.",
+					"Unlocks the [Rally|Skill+rally_the_troops] skill which can raise [morale|Concept.Morale] of all nearby allies of your faction to a steady level.",
 					"The higher the [Resolve|Concept.Bravery] of the character using this skill, the higher the chance to succeed."
 				]
 			}]
@@ -301,7 +301,7 @@ local vanillaDescriptions = [
 				Type = ::UPD.EffectType.Active,
 				Description = [
 					"Unlocks the [Adrenaline|Skill+adrenaline_skill] skill which puts you first in the [turn|Concept.Turn] order for the next [round,|Concept.Round] to have another [turn|Concept.Turn] before your enemies do.",
-					"Under the effects of [adrenaline|Skill+adrenaline_effect] you cannot receive and are not affected by [temporary injuries.|Concept.InjuryTemporary]"
+					"While [Adrenaline|Skill+adrenaline_effect] is active you cannot receive and are not affected by [temporary injuries.|Concept.InjuryTemporary]"
 				]
 			}]
 		})
@@ -427,7 +427,7 @@ local vanillaDescriptions = [
 				{
 					Type = ::UPD.EffectType.Active,
 					Description = [
-						"Unlocks the [Indomitable|Skill+indomitable] skill that reduces incoming damage and grants immunity to being [culled,|Perk+perk_rf_cull] [stunned,|Skill+stunned_effect] knocked back or grabbed."
+						"Unlocks the [Indomitable|Skill+indomitable] skill that reduces incoming damage and grants immunity to being [stunned,|Skill+stunned_effect] knocked back or grabbed."
 					]
 				},
 				{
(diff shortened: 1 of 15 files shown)


scripts/skills/actives/rf_blitzkrieg_skill.nut
```diff
@@ -33,7 +33,14 @@ this.rf_blitzkrieg_skill <- ::inherit("scripts/skills/skill", {
...
 			icon = "ui/icons/special.png",
-			text = ::Reforged.Mod.Tooltips.parseString("Every ally within " + ::MSU.Text.colorPositive("4") + " tiles who has at least " + ::MSU.Text.colorNegative("10") + " [Fatigue|Concept.Fatigue] remaining and is not [stunned|Skill+stunned_effect] or [fleeing|Concept.Morale] will get the [Adrenaline|Skill+adrenaline_effect] effect and build " + ::MSU.Text.colorNegative("10") + " [Fatigue|Concept.Fatigue]")
+			text = ::Reforged.Mod.Tooltips.parseString("You and allies of your faction within " + ::MSU.Text.colorPositive("4") + " tiles who have at least " + ::MSU.Text.colorNegative("10") + " [Fatigue|Concept.Fatigue] remaining gain the [Adrenaline|Skill+adrenaline_effect] effect and build " + ::MSU.Text.colorNegative("10") + " [Fatigue|Concept.Fatigue]")
+		});
+
+		ret.push({
+			id = 19,
+			type = "text",
+			icon = "ui/icons/warning.png",
+			text = "Will not affect allies who are [fleeing|Concept.Morale], [stunned|Skill+stunned_effect], or [sleeping|Skill+sleeping_effect]"
```
LordMidas, line 43: `Does not affect`

LordMidas, line 43: Mentioning `sleeping` here is definitely an improvement. Ideally we should introduce a new `Stunned` concept which lists all the various effects which count as stunning a person i.e. the effects which set `properties.IsStunned` to `true`. I say this because there are other effects which do that e.g. the `horrified_effect`. But we don't have to do that immediately.

scripts/skills/actives/rf_gain_ground_skill.nut
```diff
@@ -8,7 +8,7 @@ this.rf_gain_ground_skill <- ::inherit("scripts/skills/skill", {
 	{
 		this.m.ID = "actives.rf_gain_ground";
 		this.m.Name = "Gain Ground";
-		this.m.Description = ::Reforged.Mod.Tooltips.parseString("Keep going! After killing an enemy you may move into their tile ignoring [Zone of Control.|Concept.ZoneOfControl] Can only be used immediately after a successful kill.");
+		this.m.Description = ::Reforged.Mod.Tooltips.parseString("Keep going! Immediatly after killing an adjacent target you may move into their tile ignoring [Zone of Control.|Concept.ZoneOfControl] and with reduced [Action Point|Concept.ActionPoints] and [Fatigue|Concept.Fatigue] cost.");
```
LordMidas, line 11: Don't add the reduced AP and Fatigue cost part here because that is subject to moddability via member fields in this skill. That part is automatically added to the cost string in the tooltip of the active skill with the exact numbers as applicable.

scripts/skills/actives/rf_hold_steady_skill.nut
```diff
@@ -32,6 +32,13 @@ this.rf_hold_steady_skill <- ::inherit("scripts/skills/skill", {
 			text = ::Reforged.Mod.Tooltips.parseString("You and your allies within " + ::MSU.Text.colorPositive(4) " tiles gain the [Holding Steady|Skill+rf_hold_steady_effect] effect for one [turn|Concept.Turn]")
 		});
 
+		ret.push({
+			id = 19,
+			type = "text",
+			icon = "ui/icons/warning.png",
+			text = "Will not affect allies who are [fleeing|Concept.Morale], [stunned|Skill+stunned_effect], or [sleeping|Skill+sleeping_effect]"
```
LordMidas, line 39: - `Does not affect`
- We try to start warnings at id 20 in Reforged. So this should be 20 and the next one should be 21.

scripts/skills/actives/rf_blitzkrieg_skill.nut
```diff
@@ -33,7 +33,14 @@ this.rf_blitzkrieg_skill <- ::inherit("scripts/skills/skill", {
 			id = 10,
 			type = "text",
 			icon = "ui/icons/special.png",
-			text = ::Reforged.Mod.Tooltips.parseString("Every ally within " + ::MSU.Text.colorPositive("4") + " tiles who has at least " + ::MSU.Text.colorNegative("10") + " [Fatigue|Concept.Fatigue] remaining and is not [stunned|Skill+stunned_effect] or [fleeing|Concept.Morale] will get the [Adrenaline|Skill+adrenaline_effect] effect and build " + ::MSU.Text.colorNegative("10") + " [Fatigue|Concept.Fatigue]")
+			text = ::Reforged.Mod.Tooltips.parseString("You and allies of your faction within " + ::MSU.Text.colorPositive("4") + " tiles who have at least " + ::MSU.Text.colorNegative("10") + " [Fatigue|Concept.Fatigue] remaining gain the [Adrenaline|Skill+adrenaline_effect] effect and build " + ::MSU.Text.colorNegative("10") + " [Fatigue|Concept.Fatigue]")
+		});
+
+		ret.push({
+			id = 19,
```
LordMidas, line 40: Same comment as other place about warning id numbers.

scripts/skills/actives/rf_net_pull_skill.nut
```diff
@@ -41,7 +41,7 @@ this.rf_net_pull_skill <- ::inherit("scripts/skills/skill", {
 			id = 10,
 			type = "text",
 			icon = "ui/icons/vision.png",
-			text = "Has a range of " + ::MSU.Text.colorPositive("2") + " tiles"
+			text = "Has a range of " + ::MSU.Text.colorPositive("2") + " tiles and max height difference of 1"
```
LordMidas, line 44: I'm skeptical about adding the height difference. It isn't vanilla convention and isn't mentioned for any other 2 tile skill. In most cases this is automatically implied and learned by BB players that 2 tile melee skills have a height difference of 1. I'd say don't add this.

scripts/skills/actives/rf_net_pull_skill.nut
```diff
@@ -55,6 +55,18 @@ this.rf_net_pull_skill <- ::inherit("scripts/skills/skill", {
...
+			id = 15,
+			type = "text",
+			icon = "ui/icons/special.png",
+			text = "Will expend the currently held net"
+		});
+		ret.push({
+			id = 20,
+			type = "text",
+			icon = "ui/icons/warning.png",
+			text = "There must be an open tile for the target to be pulled into"
```
LordMidas, line 68: This open tile entry is redundant imo (i.e. it is obvious from the use case of the skill) and should be removed.

scripts/skills/actives/rf_onslaught_skill.nut
```diff
@@ -32,6 +32,13 @@ this.rf_onslaught_skill <- ::inherit("scripts/skills/skill", {
 			text = ::Reforged.Mod.Tooltips.parseString("You and your allies within " + ::MSU.Text.colorPositive(4) " tiles gain the [Onslaught|Skill+rf_onslaught_effect] effect for one [turn|Concept.Turn]")
 		});
 
+		ret.push({
+			id = 19,
+			type = "text",
+			icon = "ui/icons/warning.png",
+			text = "Will not affect allies who are [fleeing|Concept.Morale], [stunned|Skill+stunned_effect], or [sleeping|Skill+sleeping_effect]"
```
LordMidas, line 39: `Does not affect`

scripts/skills/actives/rf_onslaught_skill.nut
```diff
@@ -32,6 +32,13 @@ this.rf_onslaught_skill <- ::inherit("scripts/skills/skill", {
 			text = ::Reforged.Mod.Tooltips.parseString("You and your allies within " + ::MSU.Text.colorPositive(4) " tiles gain the [Onslaught|Skill+rf_onslaught_effect] effect for one [turn|Concept.Turn]")
 		});
 
+		ret.push({
+			id = 19,
```
LordMidas, line 36: Start warnings at 20

scripts/skills/actives/rf_swordmaster_charge_skill.nut
```diff
@@ -50,7 +50,7 @@ this.rf_swordmaster_charge_skill <- ::inherit("scripts/skills/actives/rf_swordma
 			id = 11,
 			type = "text",
 			icon = "ui/icons/special.png",
-			text = "Will move to the targeted tile and attack a random adjacent enemy with " + ::MSU.Text.colorPositive((this.m.DamageMultBonus * 100) + "%") + " more damage"
+			text = "Will move to the targeted tile and attack a random adjacent enemy with " + ::MSU.Text.colorPositive((this.m.DamageMultBonus * 100) + "%") + " increased damage"
```
LordMidas, line 53: We've shifted away from `increased/decreased` and are going with `more/less` for multiplicative modifiers. Is there a particular reason why you prefer `increased` here? If it is because of better language flow then perhaps change `... with x% increased damage` to `... dealing x% more damage`.

scripts/skills/actives/rf_swordmaster_tackle_skill.nut
```diff
@@ -41,7 +41,7 @@ this.rf_swordmaster_tackle_skill <- ::inherit("scripts/skills/actives/rf_swordma
 			id = 11,
 			type = "text",
 			icon = "ui/icons/special.png",
-			text = ::Reforged.Mod.Tooltips.parseString("If the attack is successful, the target is [stunned|Skill+stunned_effect] and you exchange positions with the target")
+			text = ::Reforged.Mod.Tooltips.parseString("If the attack is successful, [stun|Skill+stunned_effect] and exchange positions with the target")
```
LordMidas, line 44: I think the original is better here.

scripts/skills/effects/rf_onslaught_effect.nut
```diff
@@ -37,7 +37,7 @@ this.rf_onslaught_effect <- ::inherit("scripts/skills/skill", {
 				id = 12,
 				type = "text",
 				icon = "ui/icons/special.png",
-				text = ::Reforged.Mod.Tooltips.parseString("Gain one use of the [Linebreaker|Skill+rf_line_breaker_skill] skill with reduced [Action Point|Concept.ActionPoints] and [Fatigue|Concept.Fatigue] Cost")
+				text = ::Reforged.Mod.Tooltips.parseString("Gain one use of the [Linebreaker|Skill+rf_line_breaker_skill] skill that costs 1 less [Action Point|Concept.ActionPoints] and 10 less [Fatigue|Concept.Fatigue]")
```
LordMidas, line 40: - `1 fewer`
- `and builds 10 less`

Both 1 and 10 should be colored via `::MSU.Text.colorPositive` (see other examples in code for the syntax).

mod_reforged/hooks/config/strings.nut
```diff
@@ -252,7 +252,7 @@ local vanillaDescriptions = [
 			Effects = [{
 				Type = ::UPD.EffectType.Passive,
 				Description = [
-					"Reduce the [Hitpoint|Concept.Hitpoints] damage threshold for triggering [Morale Checks|Concept.Morale] with your attacks to 1.",
+					"Reduce the [Hitpoint|Concept.Hitpoints] damage threshold for triggering [morale checks|Concept.Morale] with your attacks to 1.",
```
LordMidas, line 255: I'd change this line to `The damage threshold for triggering morale checks with your attacks is reduced to 1`

mod_reforged/hooks/config/strings.nut
```diff
@@ -301,7 +301,7 @@ local vanillaDescriptions = [
 				Type = ::UPD.EffectType.Active,
 				Description = [
 					"Unlocks the [Adrenaline|Skill+adrenaline_skill] skill which puts you first in the [turn|Concept.Turn] order for the next [round,|Concept.Round] to have another [turn|Concept.Turn] before your enemies do.",
-					"Under the effects of [adrenaline|Skill+adrenaline_effect] you cannot receive and are not affected by [temporary injuries.|Concept.InjuryTemporary]"
+					"While [Adrenaline|Skill+adrenaline_effect] is active you cannot receive and are not affected by [temporary injuries.|Concept.InjuryTemporary]"
```
LordMidas, line 304: I prefer the original here.

mod_reforged/hooks/config/strings.nut
```diff
@@ -447,7 +447,7 @@ local vanillaDescriptions = [
 			Effects = [{
 				Type = ::UPD.EffectType.Passive,
 				Description = [
-					"With no adjacent ally of your faction, gain a " + ::MSU.Text.colorPositive("+15%") + " bonus to [Melee Skill,|Concept.MeleeSkill] [Ranged Skill,|Concept.RangeSkill] [Melee Defense,|Concept.MeleeDefense] [Ranged Defense,|Concept.RangeDefense] and [Resolve.|Concept.Bravery]"
+					"When no ally of your faction is adjacent to you, gain a " + ::MSU.Text.colorPositive("+15%") + " bonus to [Melee Skill,|Concept.MeleeSkill] [Ranged Skill,|Concept.RangeSkill] [Melee Defense,|Concept.MeleeDefense] [Ranged Defense,|Concept.RangeDefense] and [Resolve.|Concept.Bravery]"
```
LordMidas, line 450: Take this opportunity to change the `+15%` to `15%` here to show that its multiplicative.

mod_reforged/hooks/config/strings.nut
```diff
@@ -534,11 +534,17 @@ local vanillaDescriptions = [
...
 						"[Handgonnes|Item+handgonne] now require " + ::MSU.Text.colorNegative("6") + " [Action Points|Concept.ActionPoints] to [reload|Skill+reload_handgonne_skill] and can be fired every turn instead of every other turn."
 					]
-				}
+				},
+				{
+					Type = ::UPD.EffectType.Active,
+					Description = [
+						"Unlocks the [Take Aim|Skill+rf_take_aim_skill] skill which allows you to target opponents behind obstacles with a [crossbow,|Item+crossbow] or hit more targets with a [handgonne.|Item+handgonne]"
+					]
+				},
```
LordMidas, line 546: Remove comma

mod_reforged/hooks/config/strings.nut
```diff
@@ -648,7 +654,7 @@ local vanillaDescriptions = [
 					"Skills build up " + ::MSU.Text.colorNegative("25%") + " less [Fatigue.|Concept.Fatigue]",
 					"The [Action Point|Concept.ActionPoints] cost of [Spearwall|Skill+spearwall] is reduced by " + ::MSU.Text.colorPositive("1") + ".",
 					"[Spearwall|Skill+spearwall] is no longer disabled once an opponent manages to overcome it. Instead, [Spearwall|Skill+spearwall] can still be used and continues to give free attacks on any further opponent attempting to enter the [Zone of Control|Concept.ZoneOfControl]",
-					"When starting your [turn|Concept.Turn] with a spear equipped, the first spear attack during your [turn|Concept.Turn] costs no [Action Points|Concept.ActionPoints] and builds no [Fatigue,|Concept.Fatigue] but does " + ::MSU.Text.colorNegative("25%") + " less damage. Expires upon switching your weapon.",
+					"When starting your [turn|Concept.Turn] with a spear equipped, the first spear attack during your [turn|Concept.Turn] costs no [Action Points|Concept.ActionPoints] and builds no [Fatigue,|Concept.Fatigue] but deals " + ::MSU.Text.colorNegative("25%") + " reduced damage. Expires upon switching your weapon.",
```
LordMidas, line 657: Same comment about more/less for multiplicative changes as mentioned earlier.

mod_reforged/hooks/config/strings.nut
```diff
@@ -743,9 +749,9 @@ local vanillaDescriptions = [
 					"Damage to [Hitpoints|Concept.Hitpoints] is reduced by " + ::MSU.Text.colorPositive("50%") + " and that to armor by " + ::MSU.Text.colorPositive("25%") + ".",
 					"The bonus drops exponentially when wearing head and body armor with a total penalty to [Maximum Fatigue|Concept.MaximumFatigue] above 15. The lighter your armor and helmet, the more you benefit.",
 					"Does not affect damage from mental attacks or status effects, but can help to avoid receiving them.",
-					"When attacking a target against whom you have a [Reach Disadvantage,|Concept.ReachAdvantage] reduce this disadvantage by " + ::MSU.Text.colorPositive(1) + " if your [Initiative|Concept.Initiative] is higher than that of your target."
+					"When attacking a target against whom you have a [Reach Disadvantage,|Concept.ReachAdvantage] reduce this disadvantage by " + ::MSU.Text.colorPositive(1) + " if your [Initiative|Concept.Initiative] is greater than that of your target."
 					"[Brawny|Perk+perk_brawny] does not affect this perk.",
-					"Cannot be picked if you have [Poise.|Perk+perk_rf_poise]"
+					"Cannot be taken if you have [Poise.|Perk+perk_rf_poise]"
```
LordMidas, line 754: Unsure about this. Please explain why taken is better than picked.

mod_reforged/hooks/config/strings.nut
```diff
@@ -841,7 +848,7 @@ local vanillaDescriptions = [
 				Type = ::UPD.EffectType.Passive,
 				Description = [
 					"Once per battle, upon receiving a killing blow, survive instead with a few hitpoints left and have all damage over time effects (e.g. bleeding, poisoned) cured.",
-					"When this effect triggeres, gain increased stats until your next turn"
+					"When this effect triggers, gain increased stats until your next turn"
```
LordMidas, line 851: This line should be: `"Upon triggering, gain increased [attributes|Concept.CharacterAttribute] until your next [turn.|Concept.Turn]"`

mod_reforged/hooks/config/strings.nut
```diff
@@ -841,7 +848,7 @@ local vanillaDescriptions = [
 				Type = ::UPD.EffectType.Passive,
 				Description = [
 					"Once per battle, upon receiving a killing blow, survive instead with a few hitpoints left and have all damage over time effects (e.g. bleeding, poisoned) cured.",
```
LordMidas, line 850: change hitpoints to `[Hitpoints|Concept.Hitpoints]` and hyperlink both bleeding and poisoned to their respective effects e.g. `[bleeding+Skill+bleeding_effect], [poisoned|Skill+spider_poison_effect]`

mod_reforged/hooks/config/strings.nut
```diff
@@ -900,7 +907,7 @@ foreach (vanillaDesc in vanillaDescriptions)
 			Type = ::UPD.EffectType.Passive,
 			Description = [
 				"[Resolve|Concept.Bravery] is increased by " + ::MSU.Text.colorPositive("10%") + " at all times.",
-				"Additionally, at positive morale checks, [Resolve|Concept.Bravery] is increased by a further " + ::MSU.Text.colorPositive(10) + ".",
+				"Additionally, for positive morale checks, [Resolve|Concept.Bravery] is increased by a further " + ::MSU.Text.colorPositive("10%") + ".",
```
LordMidas, line 910: This is minor but `at` morale checks is vanilla convention e.g. in superstitious trait. So perhaps we should stick with that just for consistency across tooltips (OCD :P) unless we want to change vanilla to `for` too .

mod_reforged/hooks/config/strings.nut
```diff
@@ -992,7 +999,7 @@ foreach (vanillaDesc in vanillaDescriptions)
 		Effects = [{
 			Type = ::UPD.EffectType.Passive,
 			Description = [
-				"Attacks against [surrounded|Concept.Surrounding] targets deal " + ::MSU.Text.colorPositive("10%") + " more damage and have " + ::MSU.Text.colorPositive("+10%") + " armor penetration per character surrounding the target."
+				"Attacks against [surrounded|Concept.Surrounding] targets deal " + ::MSU.Text.colorPositive("10%") + " more damage and have " + ::MSU.Text.colorPositive("+10%") + " armor penetration per ally surrounding the target."
```
LordMidas, line 1002: `character` is correct because surrounding in vanilla works even if a character of a faction hostile to you is surrounding the target as long as he is also hostile to your target.

mod_reforged/hooks/config/strings.nut
```diff
@@ -1001,7 +1008,7 @@ foreach (vanillaDesc in vanillaDescriptions)
 		Effects = [{
 			Type = ::UPD.EffectType.Active,
 			Description = [
-				"Unlocks the [Blitzkrieg|Skill+rf_blitzkrieg_skill] skill which allows you and the rest of your company to go first in the next round of combat."
+				"Unlocks the [Blitzkrieg|Skill+rf_blitzkrieg_skill] skill which allows you and the rest of your company to gain the [Adrenaline|Skill+adrenaline_effect] effect."
```
LordMidas, line 1011: I prefer the original wording here.

mod_reforged/hooks/config/strings.nut
```diff
@@ -1084,7 +1091,7 @@ foreach (vanillaDesc in vanillaDescriptions)
 		Effects = [{
 			Type = ::UPD.EffectType.Passive,
 			Description = [
-				"Attacks against characters with a lower [Morale|Concept.Morale] than you deal " + ::MSU.Text.colorPositive("10%") + " increased damage per level of [Morale|Concept.Morale] difference."
+				"Attacks against characters with a lower [morale|Concept.Morale] than you deal " + ::MSU.Text.colorPositive("10%") + " increased damage per level of [morale|Concept.Morale] difference."
```
LordMidas, line 1094: Take this opportunity to change `increased` to `more` here.

mod_reforged/hooks/config/strings.nut
```diff
@@ -1133,8 +1140,8 @@ foreach (vanillaDesc in vanillaDescriptions)
 		Effects = [{
 			Type = ::UPD.EffectType.Passive,
 			Description = [
-				"Hits to the head with one-handed maces apply the [Dazed|Skill+dazed_effect] effect for 1 [turn|Concept.Turn] and if the target is already [dazed,|Skill+dazed_effect] apply [Stunned|Skill+stunned_effect] for one [turn.|Concept.Turn]",
-				"Hits to the head with two-handed maces apply the [Stunned|Skill+stunned_effect] for 1 [turn.|Concept.Turn]"
+				"Hits to the head with one-handed maces apply the [Dazed|Skill+dazed_effect] effect for 1 [turn|Concept.Turn] and if the target is already [Dazed,|Skill+dazed_effect] apply [stunned|Skill+stunned_effect] for one [turn.|Concept.Turn]",
```
LordMidas, line 1143: I prefer the `dazed` here with lowercase as it flows in the sentence better imo.

mod_reforged/hooks/config/strings.nut
```diff
@@ -1133,8 +1140,8 @@ foreach (vanillaDesc in vanillaDescriptions)
...
+				"Hits to the head with two-handed maces apply [stunned|Skill+stunned_effect] for 1 [turn.|Concept.Turn]"
```
LordMidas, line 1144: Can change it to `"Hits to the head with two-handed maces [stun|Skill+stunned_effect] the target for 1 [turn.|Concept.Turn]"`

mod_reforged/hooks/config/strings.nut
```diff
@@ -1176,7 +1183,7 @@ foreach (vanillaDesc in vanillaDescriptions)
 				"Gain a stack whenever you end your [turn|Concept.Turn] without having used [Wait,|Concept.Wait] up to a maximum of 3 stacks.",
 				"Gain " + ::MSU.Text.colorPositive("+15") + " [Resolve|Concept.Bravery] and [Initiative|Concept.Initiative] while you have at least 1 stack.",
 				"Skills build up " + ::MSU.Text.colorPositive("15%") + " less [Fatigue|Concept.Fatigue] while you have at least 2 stacks.",
-				"Deal " + ::MSU.Text.colorPositive("15%") + " more damage while you have 3 stacks.",
+				"Deal " + ::MSU.Text.colorPositive("15%") + " additional damage while you have 3 stacks.",
```
LordMidas, line 1186: `more` is correct here. We should reserve the term `additional` for things which additively change a modifier.

mod_reforged/hooks/config/strings.nut
```diff
@@ -1251,7 +1258,7 @@ foreach (vanillaDesc in vanillaDescriptions)
 		Effects = [{
 			Type = ::UPD.EffectType.Passive,
 			Description = [
-				"A successful hit increases the damage of your next attack(s) by " + ::MSU.Text.colorNegative("20%") + ".",
+				"A successful hit increases the damage of your next attack by " + ::MSU.Text.colorNegative("20%") + ".",
 				"The effect is lost upon moving, swapping an item, using any skill except a single-target attack, missing an attack, or waiting or ending your [turn.|Concept.Turn]"
```
LordMidas, line 1262: Use this opportunity to add `[waiting|Concept.Wait]` here

mod_reforged/hooks/config/strings.nut
```diff
@@ -1338,7 +1345,7 @@ foreach (vanillaDesc in vanillaDescriptions)
 				Type = ::UPD.EffectType.Passive,
 				Description = [
 					"Always start battles at Confident [morale.|Concept.Morale]",
-					::MSU.Text.colorPositive("50%") + " increased [Resolve|Concept.Resolve] against negative morale checks."
+					"Gain " ::MSU.Text.colorPositive("50%") + " increased [Resolve|Concept.Resolve] against negative morale checks."
```
LordMidas, line 1348: `[Resolve|Concept.Bravery] is increased by " + ::MSU.Text.colorPositive("50%") + " against negative [morale checks.|Concept.Morale]"`

mod_reforged/hooks/config/strings.nut
```diff
@@ -1454,12 +1461,12 @@ foreach (vanillaDesc in vanillaDescriptions)
 	}),
 	RF_Ghostlike = ::UPD.getDescription({
 		Fluff = "Blink and you\'ll miss me.",
-		Requirement = "Total armor weight up to 20"
+		Requirement = "Total armor weight of 20 or less"
```
LordMidas, line 1464: Longer requirement strings look uglier in the tooltip so `up to 20` is better imo.

mod_reforged/hooks/config/strings.nut
```diff
@@ -1454,12 +1461,12 @@ foreach (vanillaDesc in vanillaDescriptions)
...
 		Effects = [{
 			Type = ::UPD.EffectType.Passive,
 			Description = [
 				"When the number of adjacent allies is greater than or equal to the number of adjacent enemies, you may ignore [Zone of Control|Concept.ZoneOfControl] for your next movement action."
-				"When wielding a melee weapon with up to 4 [Reach|Concept.Reach] when you end your movement adjacent to an enemy, the next attack against that enemy deals " + ::MSU.Text.colorPositive("25%") + " more damage and " + ::MSU.Text.colorPositive("+20%") + " damage ignoring armor. This bonus expires upon any action other than an attack."
+				"When wielding a melee weapon with up to 4 [Reach|Concept.Reach] when you end your movement adjacent to an enemy, the next attack against that enemy deals " + ::MSU.Text.colorPositive("25%") + " additional damage and " + ::MSU.Text.colorPositive("+20%") + " damage ignoring armor. This bonus expires upon using any action other than an attack."
```
LordMidas, line 1469: `more damage` is correct. Not `additional`.

LordMidas, line 1469: I don't think `upon using any action` is better than `upon any action`. I prefer the original as it flows better in language, `using` sounds too gamey and mechanical when referring to something like an `action` that a person `does`.

mod_reforged/hooks/config/strings.nut
```diff
@@ -1493,7 +1500,7 @@ foreach (vanillaDesc in vanillaDescriptions)
 			{
 				Type = ::UPD.EffectType.Active,
 				Description = [
-					"Unlocks the [Pocket Sand|Skill+rf_pocket_sand_skill] skill which allows you to store sand in your empty [bag slots|Concept.BagSlots] to distract nearby enemies with."
+					"Unlocks the [Pocket Sand|Skill+rf_pocket_sand_skill] skill which allows you to store sand in your empty [bag slots|Concept.BagSlots] with which to distract nearby enemies."
```
LordMidas, line 1503: I prefer the original here.

mod_reforged/hooks/config/strings.nut
```diff
@@ -1541,7 +1548,7 @@ foreach (vanillaDesc in vanillaDescriptions)
 			{
 				Type = ::UPD.EffectType.Active,
 				Description = [
-					"Gain the [Encourage|Skill+rf_encourage_skill] skill that allows you to increase the [Morale|Concept.Morale] of an ally."
+					"Gain the [Encourage|Skill+rf_encourage_skill] skill that allows you to increase the [morale|Concept.Morale] of an ally."
```
LordMidas, line 1551: Minor: We're going with `which allows` with pretty much every other skill (and that's how vanilla does it in perk tooltips too) so let's change this from `that allows` to `which allows`.

mod_reforged/hooks/config/strings.nut
```diff
@@ -1600,12 +1607,12 @@ foreach (vanillaDesc in vanillaDescriptions)
 		}]
 	}),
 	RF_Kingfisher = ::UPD.getDescription({
-		Fluff = "\'There is no limit to how many times you can go fishing.\'",
-		Requirement = "Net",
+		Fluff = "\'Teach a man to fish and he'll be worth his salt to the end of his days.\'",
```
LordMidas, line 1610: Very nice

mod_reforged/hooks/config/strings.nut
```diff
@@ -1600,12 +1607,12 @@ foreach (vanillaDesc in vanillaDescriptions)
...
+		Requirement = "Throwing Net",
 		Effects = [{
 			Type = ::UPD.EffectType.Passive,
 			Description = [
-				"While holding a [net,|Item+throwing_net] every successful melee attack against an adjacent target has a chance to [net|Skill+net_effect] the target without expending your currently held [net.|Item+throwing_net] The chance is equal to the hit chance of the attack.",
+				"While wielding a [net,|Item+throwing_net] every successful melee attack against an adjacent target has a chance to [net|Skill+net_effect] the target without expending your currently held [net.|Item+throwing_net] The chance is equal to the hit chance of the attack.",
```
LordMidas, line 1615: I prefer `holding` rather than `wielding` for nets.

mod_reforged/hooks/config/strings.nut
```diff
@@ -1619,7 +1626,7 @@ foreach (vanillaDesc in vanillaDescriptions)
 			Type = ::UPD.EffectType.Passive,
 			Description = [
 				"The [Action Point|Concept.ActionPoints] cost of spear attacks is reduced by " + ::MSU.Text.colorPositive(1) + ".",
-				"During your turn, the first attack from a spear deals " + ::MSU.Text.colorPositive("25%") + " more damage."
+				"During your turn, the first attack from a spear deals " + ::MSU.Text.colorPositive("25%") + " increased damage."
```
LordMidas, line 1629: Same comment about more/less as in other places.

LordMidas, line 1629: Take this opportunity to convert turn to hyperlink: `[turn,|Concept.Turn]`

mod_reforged/hooks/config/strings.nut
```diff
@@ -1665,9 +1672,10 @@ foreach (vanillaDesc in vanillaDescriptions)
 			Description = [
 				"Damage to [Hitpoints|Concept.Hitpoints] is reduced by " + ::MSU.Text.colorNegative("30%") + " and to Armor by " + ::MSU.Text.colorNegative("20%") + ".",
 				"The bonus drops exponentially when wearing head and body armor with a total penalty to [Maximum Fatigue|Concept.MaximumFatigue] above 35.",
-				"When attacking a target against whom you have a [Reach Disadvantage,|Concept.ReachAdvantage] reduce this disadvantage by " + ::MSU.Text.colorPositive(1) + " if your [Initiative|Concept.Initiative] is higher than that of your target."
+				"Does not affect damage from mental attacks or status effects, but can help to avoid receiving them.",
+				"When attacking a target against whom you have a [Reach Disadvantage,|Concept.ReachAdvantage] reduce this disadvantage by " + ::MSU.Text.colorPositive(1) + " if your [Initiative|Concept.Initiative] is greater than that of your target."
 				"[Brawny|Perk+perk_brawny] does not affect this perk.",
-				"Cannot be picked if you have [Nimble.|Perk+perk_nimble]"
+				"Cannot be taken if you have [Nimble.|Perk+perk_nimble]"
```
LordMidas, line 1678: Same comment about taken vs picked as the other places. Just marking it, no need for comment on this conversation.

mod_reforged/hooks/config/strings.nut
```diff
@@ -1746,7 +1754,7 @@ foreach (vanillaDesc in vanillaDescriptions)
 		Effects = [{
 			Type = ::UPD.EffectType.Passive,
 			Description = [
-				"The [Action Point|Concept.ActionPoints] Cost of reloading a Handgonne is reduced by " + ::MSU.Text.colorPositive(2) + ".",
+				"The [Action Point|Concept.ActionPoints] cost of reloading a Handgonne is reduced by " + ::MSU.Text.colorPositive(2) + ".",
```
LordMidas, line 1757: Let's add hyperlink for `[handgonne|Item+handgonne]` and decapitalize it.

mod_reforged/hooks/config/strings.nut
```diff
@@ -2019,12 +2028,12 @@ foreach (vanillaDesc in vanillaDescriptions)
...
 		Fluff = "Let's Dance!",
-		Requirement = "Sword"
+		Requirement = "Non-Fencing Sword"
 		Effects = [{
 			Type = ::UPD.EffectType.Passive,
 			Description = [
-				"Gain additional [Initiative|Concept.Initiative] equal to the armor ignore percentage of your equipped sword.",
-				"When using a non-fencing sword, the [Action Point|Concept.ActionPoints] costs of non-AOE skills is reduced by " + ::MSU.Text.colorPositive(1) + " and these skills build up " + ::MSU.Text.colorPositive("25%") + " less [Fatigue.|Concept.Fatigue]",
+				"Gain [Initiative|Concept.Initiative] equal to the armor ignore percentage of your equipped sword.",
+				"The [Action Point|Concept.ActionPoints] cost of non-AOE skills is reduced by " + ::MSU.Text.colorPositive(1) + " and these skills build up " + ::MSU.Text.colorPositive("25%") + " less [Fatigue.|Concept.Fatigue]",
```
LordMidas, line 2036: We currently don't restrict this entire perk to Fencing Swords. Only the AP cost part was restricted away from them. We can restrict the entire perk if you prefer this design but then this has to happen in a separate PR/commit from this strings improvement PR.

mod_reforged/hooks/config/strings.nut
```diff
@@ -2019,12 +2028,12 @@ foreach (vanillaDesc in vanillaDescriptions)
...
 	RF_SwordmasterBladeDancer = ::UPD.getDescription({
 		Fluff = "Let's Dance!",
-		Requirement = "Sword"
+		Requirement = "Non-Fencing Sword"
 		Effects = [{
 			Type = ::UPD.EffectType.Passive,
 			Description = [
-				"Gain additional [Initiative|Concept.Initiative] equal to the armor ignore percentage of your equipped sword.",
-				"When using a non-fencing sword, the [Action Point|Concept.ActionPoints] costs of non-AOE skills is reduced by " + ::MSU.Text.colorPositive(1) + " and these skills build up " + ::MSU.Text.colorPositive("25%") + " less [Fatigue.|Concept.Fatigue]",
+				"Gain [Initiative|Concept.Initiative] equal to the armor ignore percentage of your equipped sword.",
```
LordMidas, line 2035: I prefer keeping the `additional` here because otherwise people will get confused if the initiative is a percentage based bonus or a flat increase.

mod_reforged/hooks/config/strings.nut
```diff
@@ -2077,8 +2086,8 @@ foreach (vanillaDesc in vanillaDescriptions)
 				Type = ::UPD.EffectType.Passive,
 				Description = [
 					"Gain the [Decapitate|Skill+decapitate] skill with swords.",
-					"Gain all the perks of the Cleaver perk group when using a sword. Swords now additionally qualify as Cleavers for you for the purposes of these perks.",
-					"Attacks from swords inflict [Bleeding|Skill+bleeding_effect] on the target."
+					"Gain all the perks of the Cleaver perk group when using a sword. Swords now additionally qualify as Cleavers for the purposes of these perks.",
+					"Attacks from swords inflict [Bleeding|Skill+bleeding_effect] as if they were Cleavers."
```
LordMidas, line 2090: No need for this addition imo.

mod_reforged/hooks/config/strings.nut
```diff
@@ -2058,7 +2067,7 @@ foreach (vanillaDesc in vanillaDescriptions)
 	}),
 	RF_SwordmasterJuggernaut = ::UPD.getDescription({
 		Fluff = "There\'s a fine line between bravery and stupidity.",
-		Requirement = "Two-Handed Sword"
+		Requirement = "Two-Handed Non-Fencing Sword"
```
LordMidas, line 2070: This change is reasonable but the perk's code doesn't coincide with this. So this should be extracted into a separate PR where we change the perk's behavior.

mod_reforged/hooks/config/strings.nut
```diff
@@ -2144,7 +2153,7 @@ foreach (vanillaDesc in vanillaDescriptions)
 			{
 				Type = ::UPD.EffectType.Passive,
 				Description = [
-					"Every hit or miss against any target increases your [Initiative|Concept.Initiative] by " + ::MSU.Text.colorPositive("+15") + ". This bonus lasts over into your next [turn|Concept.Turn] but only until the first skill used or waiting that [turn.|Concept.Turn]",
+					"Every hit or miss against any target increases your [Initiative|Concept.Initiative] by " + ::MSU.Text.colorPositive("+15") + ". This bonus lasts over into your next [turn|Concept.Turn] but only until the first skill used or upon waiting that [turn.|Concept.Turn]",
```
LordMidas, line 2156: - `lasts over` could be replaced with `is carried over`
- waiting should be hyperlinked: `[waiting|Concept.Wait]`

mod_reforged/hooks/config/strings.nut
```diff
@@ -2225,13 +2234,13 @@ foreach (vanillaDesc in vanillaDescriptions)
...
 				Description = [
-					"Immediately upon killing an opponent, if you move into the tile of the killed target, the [Action Point|Concept.ActionPoints] cost of the next skill used is " + ::MSU.Text.colorPositive("halved") + "."
+					"Immediately after killing an opponent, if you move into the tile of the killed target, the [Action Point|Concept.ActionPoints] cost of the next skill used is " + ::MSU.Text.colorPositive("halved") + "."
 				]
 			},
 			{
 				Type = ::UPD.EffectType.Active,
 				Description = [
-					"Unlocks the [Gain Ground|Skill+rf_gain_ground_skill] skill which, immediately after killing an adjacent target, allows you to move into their tile ignoring [Zone of Control|Concept.ZoneOfControl] with reduced [Action Point|Concept.ActionPoints] cost and [Fatigue|Concept.Fatigue] cost of movement."
+					"Unlocks the [Gain Ground|Skill+rf_gain_ground_skill] skill which, immediately after killing an adjacent opponent, allows you to move into their tile ignoring [Zone of Control|Concept.ZoneOfControl] and with reduced [Action Point|Concept.ActionPoints] and [Fatigue|Concept.Fatigue] cost."
```
LordMidas, line 2243: Let's decide if we want target or opponent. I'm fine with both but perhaps `opponent` sounds more mercenary like. In the perk's tooltip you changed `enemy` to `target` in this PR. Whatever you end up deciding to go with make sure that both the passive/active part of this perk tooltip and the actual perk's effect tooltip have the same.

LordMidas (changes requested):
This is great work and overall very useful changes. I have suggested some improvements.

scripts/skills/actives/rf_blitzkrieg_skill.nut
Uberbagel: I thought about this as I was doing these. I agree, sounds like an improvement. Perhaps a new concept called incapacitated instead of putting it all under stun.

scripts/skills/actives/rf_swordmaster_charge_skill.nut
Uberbagel: Yeah better language flow and feel it's more refined language. What do you think?

scripts/skills/actives/rf_swordmaster_tackle_skill.nut
Uberbagel: Felt like saying target twice is redundant. Thoughts?

LordMidas: I see yeah that's a good observation. Then your change is fine.

mod_reforged/hooks/config/strings.nut
Uberbagel: Flow and feels more refined.

mod_reforged/hooks/config/strings.nut
Uberbagel: at is fine

mod_reforged/hooks/config/strings.nut
Uberbagel: But it's significant that it's adrenaline because adrenaline now ignores injuries.

mod_reforged/hooks/config/strings.nut
Uberbagel: Should the top line then be changed to:

"Hits to the head with one-handed maces [daze|Skill+dazed_effect] the target for 1 [turn|Concept.Turn] and if the target is already [dazed,|Skill+dazed_effect] [stun|Skill+stunned_effect] the target for one [turn|Concept.Turn] instead."

mod_reforged/hooks/config/strings.nut
Uberbagel: up to 20 isn't really proper though imo. It's saying you're required to have armor weight, which isn't true. We could do `Total armor weight below 21`? I saw it done this way in one other place.

mod_reforged/hooks/config/strings.nut
Uberbagel: How about `upon taking any action other than an attack`?

mod_reforged/hooks/config/strings.nut
Uberbagel: non-restricted is fine if someone wants to bredit a mega swordsman or something

mod_reforged/hooks/config/strings.nut
Uberbagel: My thought was this added some clarity as to how many stacks of bleed would be added per hit. What do you think?

mod_reforged/hooks/config/strings.nut
Uberbagel: Perk descriptions within strings tend to be a bit more flowery and less precise, so that's why I used opponent there and target within the Gain Ground skill itself. I'll change target to opponent within Gain Ground skill if you prefer though.

mod_reforged/hooks/config/strings.nut
LordMidas: Yes sounds good. You can make it more concise by removing the mention of `the target` after stun.

mod_reforged/hooks/config/strings.nut
LordMidas: Ah yes that's a good point. Yeah then this looks good.

mod_reforged/hooks/config/strings.nut
LordMidas: I think the meaning of `up to 20` is identical to that of `20 or less`. Because it means you can't exceed 20. It doesn't mean you are required to have any.

mod_reforged/hooks/config/strings.nut
LordMidas: Yes that sounds good.

mod_reforged/hooks/config/strings.nut
LordMidas: I think that part is implied quite a bit by the entire perk description's context so it seem superfluous.

mod_reforged/hooks/config/strings.nut
LordMidas: I see then it's fine like this the way you have.

scripts/skills/actives/rf_swordmaster_charge_skill.nut
LordMidas: I think in this case `dealing x% more damage` would be the best.

scripts/skills/actives/rf_blitzkrieg_skill.nut
LordMidas: So for now keep it as you've done and once we have the #466 PR merged in we can then redo the relevant tooltips to use those concepts.

mod_reforged/hooks/config/strings.nut
LordMidas: I think `deals x% less damage` here sounds fine too.

mod_reforged/hooks/config/strings.nut
LordMidas: I don't feel that way. I may be wrong but my impression is that the community uses "pick" for perks a lot more than "take". In the vanilla tooltips themselves they use "pick", "learn" or "select" in the context of a perk. E.g.:
- Tip of the Day: `"Learn the \'Rotation\' or \'Footwork\' perks for additional mobility in battle."`
- Experience tooltip: `"Characters gain experience as they or their allies slay enemies in battles... ...he\'ll level up and be able to increase his attributes and pick a perk that grants a unique bonus..."`
- Levelup Tooltip: `"This character has leveled up. Increase his attributes and select a perk!"`

I don't like the use of `take` for perks.

mod_reforged/hooks/config/strings.nut
```diff
@@ -1140,7 +1140,7 @@ foreach (vanillaDesc in vanillaDescriptions)
 		Effects = [{
 			Type = ::UPD.EffectType.Passive,
 			Description = [
-				"Hits to the head with one-handed maces apply the [dazed|Skill+dazed_effect] effect for 1 [turn|Concept.Turn] and if the target is already [dazed,|Skill+dazed_effect] apply [stunned|Skill+stunned_effect] for one [turn.|Concept.Turn]",
+				"Hits to the head with one-handed maces [daze|Skill+dazed_effect] the target for 1 [turn|Concept.Turn] and if the target is already [dazed,|Skill+dazed_effect] [stun|Skill+stunned_effect] for one [turn instead.|Concept.Turn]",
```
LordMidas, line 1143: `instead` needs to be outside the hyperlink.

LordMidas (approved):
I'll fix the remaining thing and merge the PR.

PR #466, opened 2024-08-01
feat: add nested tooltip concepts Disabled and Rooted

LordMidas:


Uberbagel (approved):

mod_reforged/config/z_nested_tooltips.nut
```diff
@@ -65,6 +65,8 @@ local getThresholdForInjury = function( _script )
 	}
 
 	::MSU.Table.merge(::Reforged.NestedTooltips.Tooltips.Concept, {
+		Disabled = ::MSU.Class.BasicTooltip("Disabled", ::Reforged.Mod.Tooltips.parseString("A disabled character is unable to act and will skip their [turn|Concept.Turn]. Receiving a disabling [effect|Concept.StatusEffect] will set the current [Action Points|Concept.ActionPoints] of the character to 0.\n\nExamples of [effects|Concept.StatusEffect] which can cause a character to become disabled include [Stunned|Skill+stunned_effect] and [Sleeping.|Skill+sleeping_effect]")),
```
Darxo, line 68: I just checked and `sleeping_effect` only sets the Action Points to 0 when the turn of that brother starts. This is important because you can actually remove this effect allowing them to have a regular turn.

However `stunned_effect` aswell as `horrified_effect` both set the Action Points to 0 during `onUpdate`

LordMidas: I see. In that case we should not mention it and leave that behavior to be individual for each of those effects, and be a case by case basis. People will learn based on the effect.

Darxo (approved):

PR #470, opened 2024-08-04
refactor: modify gilders embrace based on Reforged shield design

LordMidas:
- Set ReachIgnore to 3
- Set StaminaModifier to be 2 less than that of sipar

### DIFF
--- a/mod_reforged/hooks/items/shields/legendary/gilders_embrace_shield.nut
+++ b/mod_reforged/hooks/items/shields/legendary/gilders_embrace_shield.nut
@@ -1,0 +1,11 @@
+::Reforged.HooksMod.hook("scripts/items/shields/legendary/gilders_embrace_shield", function(q) {
+	q.create = @(__original) function()
+	{
+		__original();
+		local sipar = ::new("scripts/items/shields/oriental/metal_round_shield");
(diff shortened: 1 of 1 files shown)


Uberbagel:
I wonder if this does enough for Gilder's Embrace? In vanilla, the comparison between Gilder's Embrace and the Sipar Shield is as follows:

Sipar:
Mdef - 18
Rdef - 18
Weight - 18
Durability - 60

Gilder's Embrace:
Mdef - 25
Rdef - 25
Weight - 16
Durability - 786
Reach Ignore - 3
Special - Everyone striking this shield receives the Dazed effect

Now in Reforged with this change, the comparison is:

Reforged Sipar:
Mdef - 25
Rdef - 10
Weight - 22
Durability - 300

Reforged Gilder's Embrace:
Mdef - 25
Rdef - 25
Weight - 20
Durability - 300
Reach Ignore - 3
Special - Everyone striking this shield receives the Dazed effect

I think this is a big reduction in comparative effectiveness compared to Vanilla. I think in Reforged, GE should either:

Increase Mdef/Rdef to 28
or
Reduce weight to 18
or
Increase Reach Ignore to 4

I think GE should also retain 786 durability for the flavor.

Darxo (approved):

PR #474, opened 2024-08-07
feat: add onDropLoot experimental module

LordMidas:
This is in response to this [feature request](https://github.com/MSUTeam/MSU/issues/220) over at MSU. We implement an experimental version of it in Reforged and if everything looks good we can consider it as a candidate for inclusion in MSU or in Modular Vanilla mod.

### DIFF
--- a/mod_reforged/hooks/experimental_modules/onDropLoot.nut
+++ b/mod_reforged/hooks/experimental_modules/onDropLoot.nut
@@ -1,0 +1,60 @@
+// Original feature request: https://github.com/MSUTeam/MSU/issues/220
+
+// Goal:
+// - Add a new function to all actors which can be used to modify the existing loot they drop on death, or to add new loot.
+
+::Reforged.HooksMod.hook("scripts/entity/tactical/actor", function(q) {
+	q.onDropLoot <- function( _originalLoot, _killer, _skill, _tile, _fatalityType )
+	{
+		if (_tile != null)
+		{
+			foreach (item in _originalLoot)
+			{
+				item.drop(_tile);
+			}
+		}
+	}
+});
+
+::Reforged.QueueBucket.VeryLate.push(function() {
+	::Reforged.HooksMod.hookTree("scripts/entity/tactical/actor", function(q) {
+		q.onDeath = @(__original) function( _killer, _skill, _tile, _fatalityType )
+		{
+			local loot = [];
+
+			// Switcheroo the ::new function to detect the creation of an item.
+			// Then switcheroo the `drop` function of the item to instead push the item
+			// to our local `loot` array. This will be used to pass this item to the `onDropLoot`
+			// function as the _originalLoot.
+			local new = ::new;
+			::new = function( _script )
+			{
+				local item = new(_script);
+				if (::isKindOf(item, "item"))
+				{
+					item.drop_temp <- item.drop;
+					item.drop = @(_t) loot.push(this);
+				}
+				return item;
+			}
+
+			__original(_killer, _skill, _tile, _fatalityType);
+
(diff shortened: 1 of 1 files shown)


mod_reforged/hooks/experimental_modules/onDropLoot.nut
```diff
@@ -0,0 +1,60 @@
+// Original feature request: https://github.com/MSUTeam/MSU/issues/220
+
+// Goal:
+// - Add a new function to all actors which can be used to modify the existing loot they drop on death, or to add new loot.
+
+::Reforged.HooksMod.hook("scripts/entity/tactical/actor", function(q) {
+	q.onDropLoot <- function( _originalLoot, _killer, _skill, _tile, _fatalityType )
```
LordMidas, line 7: A problem is when we have let's say 2 mods who both want to drop everything from the `_originalLoot` except `item.my_item`. So they both hook this function and iterate over the `_originalLoot` and drop everything manually except that item type. This will cause that item to be dropped twice.

Instead of this perhaps it'd be better to have a function like `getLoot` which is passed the `_originalLoot` and then mods can hook/modify that function to add/remove things to the loot. Then finally the loot is dropped by an internal function. Using this system the above example is solved because both mods modify the `_loot` array and remove `item.some_item` from it. They don't care about manually dropping the rest and leave that up to the framework.

mod_reforged/hooks/experimental_modules/onDropLoot.nut
```diff
@@ -0,0 +1,65 @@
...
+			// to our local `loot` array. This will be used to pass this item to the `prepareLoot`
+			// function as the _loot.
+			local new = ::new;
+			::new = function( _script )
+			{
+				local item = new(_script);
+				if (::isKindOf(item, "item"))
+				{
+					item.drop_temp <- item.drop;
+					item.drop = @(_t) loot.push(this);
```
Suor, line 40:
```suggestion
					item.drop = @(_t) loot.push(_t);
```
probably

mod_reforged/hooks/experimental_modules/onDropLoot.nut
```diff
@@ -0,0 +1,65 @@
...
+			__original(_killer, _skill, _tile, _fatalityType);
+
+			// Revert the switcheroo on ::new
+			::new = new;
+
+			// Revert the switcheroo on the drop function of all dropped items
+			foreach (item in loot)
+			{
+				item.drop = item.drop_temp;
+				delete item.drop_temp;
```
Suor, line 54:
```suggestion
				item.drop = delete item.drop_temp;
```
Can write this way

Suor:
BTW, why do you use late hook? In my experience this lowers modability/compatiblity with other mods.

LordMidas:
> BTW, why do you use late hook? In my experience this lowers modability/compatiblity with other mods.

Because I want to wrap it after changes from other mods (as this module is intended to be part of a standard library e.g. MSU or Modular Vanilla). If I do it in Normal hook then it will not properly wrap around changes from mods that queue after it. Our wrapping doesn't change the function's functionality, but rather extends it in a safe way without altering behavior, therefore it needs to be Late so that it can wrap over other people's functionality changes.

mod_reforged/hooks/experimental_modules/onDropLoot.nut
LordMidas: Interesting, thank you.

LordMidas:
This is not necessary as of vanilla 1.5.1.4 as vanilla implemented their own system for dropping loot.

Suor:
Only vanilla system is not used for most enemies

PR #475, opened 2024-08-07
refactor: add skills of all items to their tooltip

LordMidas:


LordMidas:
I tested it and it seems to work quite nicely so I'll merge it in for the next build.

PR #485, opened 2024-08-13
refactor: add tip of the day for unlimited NPC ammunition

Darxo:


LordMidas (approved):
I pushed a commit to make the diff a bit cleaner. Squash before merging.

PR #486, opened 2024-08-14
refactor: add new art for swordstaff and named swordstaff

LordMidas:
New art for swordstaves I created by kitbashing vanilla art and then editing and polishing.

Also fixes the issue with swordstaff not rotating properly during Spearwall. This is fixed by adding `f2="-50"` to the sprite metadata, which somehow seems to have something to do with sprite rotation.


Darxo:
Imo the blade in the inventory icon for the sword staff should be facing top left.
- This is directly in line with how the weapon looks when equipped to a brother and how it looks when used in battle by allies
- Doing this for all two-handed weapons will help differentiate them better from one-handed ones when looking through the inventory

PR #492, opened 2024-08-26
refactor: improve modability of the split_shield skill

Darxo:
This chance allows it submods to easily adjust the shielddamage calculation aswell as turning off the shield damage part or the fatigue damage part.


LordMidas:
There is already a PR for this purpose #409 with some discussion and changes already happened there. Is there a reason for this new PR?

Darxo:
I forgot about that PR completely. I will close this then and move new changes over

PR #493, opened 2024-09-02
Refactor chance to hit head concept

Darxo:
image


mod_reforged/hooks/config/strings.nut
```diff
@@ -576,7 +576,7 @@ local vanillaDescriptions = [
 					"Skills build up " + ::MSU.Text.colorNegative("25%") + " less Fatigue.",
 					"[Lash|Skill+lash_skill] and [Hail|Skill+hail_skill] ignore the defense bonus of shields.",
 					"Gain the [From all Sides|Perk+perk_rf_from_all_sides] perk.",
-					"Pound ignores an additional " + ::MSU.Text.colorPositive("+10%") + " of armor on head hits.",
+					"Pound ignores an additional " + ::MSU.Text.colorPositive("+10%") + " of armor on [head hits|Concept.ChanceToHitHead].",
```
Uberbagel, line 579: I think your concept description doesn't always match the use case because of the first sentence and this is one of those cases. I think renaming the concept to Headshots or Critical Hits would be preferable with that first sentence removed. So it would read:

A headshot deals + 50% Hitpoints damage. The default chance to hit the head is 25% but this can be modified by perks or specific attacks. Some enemies lack a head, resulting in all attacks hitting the body.

mod_reforged/hooks/config/strings.nut
```diff
@@ -833,7 +833,7 @@ local vanillaDescriptions = [
 			Effects = [{
 				Type = ::UPD.EffectType.Passive,
 				Description = [
-					"Hits to the head no longer cause critical damage to this character, which also lowers the risk of sustaining debilitating head [injuries|Concept.Injury] significantly."
+					"[Hits to the head|Concept.ChanceToHitHead] no longer cause critical damage to this character, which also lowers the risk of sustaining debilitating head [injuries|Concept.Injury] significantly."
```
Uberbagel, line 836: Another instance.

mod_reforged/hooks/config/strings.nut
```diff
@@ -1139,8 +1139,8 @@ foreach (vanillaDesc in vanillaDescriptions)
 		Effects = [{
 			Type = ::UPD.EffectType.Passive,
 			Description = [
-				"Hits to the head with one-handed maces [daze|Skill+dazed_effect] the target for 1 [turn|Concept.Turn] and if the target is already [dazed,|Skill+dazed_effect] [stun|Skill+stunned_effect] for one [turn|Concept.Turn] instead.",
-				"Hits to the head with two-handed maces [stun|Skill+stunned_effect] the target for 1 [turn.|Concept.Turn]"
+				"[Hits to the head|Concept.ChanceToHitHead] with one-handed maces [daze|Skill+dazed_effect] the target for 1 [turn|Concept.Turn] and if the target is already [dazed,|Skill+dazed_effect] [stun|Skill+stunned_effect] for one [turn|Concept.Turn] instead.",
+				"[Hits to the head|Concept.ChanceToHitHead] with two-handed maces [stun|Skill+stunned_effect] the target for 1 [turn.|Concept.Turn]"
```
Uberbagel, line 1143: Another instance.

mod_reforged/hooks/config/strings.nut
```diff
@@ -1151,15 +1151,15 @@ foreach (vanillaDesc in vanillaDescriptions)
 			{
 				Type = ::UPD.EffectType.Passive,
 				Description = [
-					"Hits to the head which inflict at least 1 damage to [Hitpoints|Concept.Hitpoints] will instantly kill a target that has less than " + ::MSU.Text.colorNegative("33%") + " [Hitpoints|Concept.Hitpoints] remaining after the hit.",
+					"[Hits to the head|Concept.ChanceToHitHead] which inflict at least 1 damage to [Hitpoints|Concept.Hitpoints] will instantly kill a target that has less than " + ::MSU.Text.colorNegative("33%") + " [Hitpoints|Concept.Hitpoints] remaining after the hit.",
```
Uberbagel, line 1154: Another instance.

mod_reforged/hooks/config/strings.nut
```diff
@@ -1445,7 +1445,7 @@ foreach (vanillaDesc in vanillaDescriptions)
 			Type = ::UPD.EffectType.Passive,
 			Description = [
 				"Successful hits from Flails progressively reduce the target's [Melee Defense|Concept.MeleeDefense] and [Ranged Defense|Concept.RangeDefense] by a stacking " + ::MSU.Text.colorNegative(-5) + " for one [turn.|Concept.Turn]",
-				"The effect is doubled if the attacks hit the head."
+				"The effect is doubled if the attacks [chance to hit the head|Concept.ChanceToHitHead]."
```
Uberbagel, line 1448: Another instance plus 'chance to' in the description shouldn't be added.

Uberbagel (changes requested):

mod_reforged/hooks/config/strings.nut
Darxo: `ChanceToHitHead` is a vanilla concept that is normally only shown when you hover over the headshot chance bar in the stats section of a brother.
It has this text normally: 
> "The base percentage chance to hit a target\'s head for increased damage. The final chance can be modified by the skill used."
But that text is very barebones so I replaced it.

Unless we want multiple concepts around headshots, we need to find one that fits most/all situations.
I agree, the first sentence looks out of place sometimes. But it is pretty useful in one situation; to explain the status bar.

Uberbagel: Can you use the sentence I wanted to remove only for the status bar?

PR #494, opened 2024-09-02
refactor: improve code and design of bandits

Uberbagel:
equip throwing weapons to bag
give vandal, raider and highwayman quick hands
remove longsword from killer

Code on fast bandits can probably be further improved by adding to a single list for weapon rolls depending on getting throwing net or not, but I don't recall the code for this and don't know where a reference is.
image

### DIFF
--- a/scripts/entity/tactical/enemies/rf_bandit_bandit.nut
+++ b/scripts/entity/tactical/enemies/rf_bandit_bandit.nut
@@ -81,7 +81,6 @@ this.rf_bandit_bandit <- ::inherit("scripts/entity/tactical/human", {
 					"scripts/items/weapons/rf_poleflail",
 					"scripts/items/weapons/pike",
 					"scripts/items/weapons/spetum",
-					"scripts/items/weapons/rf_two_handed_falchion",
 					"scripts/items/weapons/warbrand"
 				]);
 			}
--- a/scripts/entity/tactical/enemies/rf_bandit_bandit.nut
+++ b/scripts/entity/tactical/enemies/rf_bandit_bandit.nut
@@ -61,7 +61,7 @@ this.rf_bandit_bandit <- ::inherit("scripts/entity/tactical/human", {
 
 	function assignRandomEquipment()
 	{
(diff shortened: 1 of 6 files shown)


scripts/entity/tactical/enemies/rf_bandit_bandit.nut
```diff
@@ -1,6 +1,8 @@
 this.rf_bandit_bandit <- ::inherit("scripts/entity/tactical/human", {
 	m = {
-		MyVariant = 0 // 1 for regular throwing weapons, 0 for not or throwing spear.
+		HasNet = 0, // 33% chance
+		IsRegularThrower = 0, // 50% chance
+		IsSpearThrower = 0 // 25% chance. only rolled if not regular thrower.
```
Darxo, line 5: I much prefer treating these variables as booleans instead for clarify of intent.
So have them be `false` by default.

And then change the roll-code to
this.m.IsSpearThrower = ::Math.rand(0, 3) == 0;

Darxo: Alternatively, if it is really your intent to capture the rolls, then you should rename the variables by having them include `Roll` at the start or end, instead of `has` and `is`

LordMidas: I agree with Darxo. So the rolling could be like:
```
this.m.IsRegularThrower = ::Math.rand(1, 2) == 2; // this is a 50% chance.
```
So if the roll is 2 then `this.m.RegularThrower` will be `true` otherwise it'll be `false`. And so on with the other fields.

Uberbagel: This branch was tested in-game both before and after the boolean change.

LordMidas:
I've pushed some improvements to the code and have also included improvements to the fast bandits code as you mentioned in the OP.

Uberbagel:
> I've pushed some improvements to the code and have also included improvements to the fast bandits code as you mentioned in the OP.
Excellent, thank you

LordMidas (approved):

LordMidas (approved):

PR #503, opened 2024-09-08
Improve racial tooltips

Darxo:
This PR does
- Add hyperlinks for existing tooltip lines of racial effects
- Add new hyperlinked tooltip line for morale immunity to racial effects

I confirmed the new hyperlinks in the following screenshots working correctly.
I haven't checked the other entities, where I only added hyperlinks.

image

image

image

image

### DIFF
--- a/scripts/skills/racial/rf_zombie_racial.nut
+++ b/scripts/skills/racial/rf_zombie_racial.nut
@@ -76,7 +76,7 @@ this.rf_zombie_racial <- ::inherit("scripts/skills/skill", {
 			{
 				id = 24,
 				type = "text",
-				icon = "ui/icons/special.png",
+				icon = "ui/icons/morale.png",
 				text = ::Reforged.Mod.Tooltips.parseString("Unaffected by [Morale|Concept.Morale]")
 			}
 		]);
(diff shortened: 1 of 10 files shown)


scripts/skills/racial/rf_zombie_racial.nut
```diff
@@ -76,7 +76,7 @@ this.rf_zombie_racial <- ::inherit("scripts/skills/skill", {
 			{
 				id = 24,
 				type = "text",
-				icon = "ui/icons/special.png",
+				icon = "ui/icons/morale.png",
 				text = ::Reforged.Mod.Tooltips.parseString("Unaffected by [Morale|Concept.Morale]")
```
LordMidas, line 80: Might as well change this line to `Not affected by` to match the other tooltips. I remember I had originally preferred to write `Unaffected by` because it is slightly better to convey the meaning that this entity doesn't use Morale instead of "it has morale but is not affected by its the effects of morale state". But I guess both wordings are overall not too big of a deal - but we should use the same wording for all these entities.

LordMidas (approved):

scripts/skills/racial/rf_zombie_racial.nut
Darxo: Referenced in https://github.com/Battle-Modders/mod-reforged/issues/505

PR #504, opened 2024-09-08
refactor: give t2+ tough bandits vigorous assault

Uberbagel:


Uberbagel:
Thought of this as I was playing but I think it's biased because I'm playing band of poachers who do fantastically well against 2h threats. Closing without merge.

PR #508, opened 2024-09-09
refactor: allow throw spear to target shield and inflict fatigue

LordMidas:
Now always hits shield if target has shield (just like it used to) and inflicts damage to the shield (vanilla default is 26 damage) and inflicts 40 Fatigue (Reforged mechanic).


Darxo:
This is how the tooltip looks now:

image

Darxo:
I pushed a tooltip improvement and fix
image

Darxo (approved):

Uberbagel (approved):

PR #509, opened 2024-09-09
refactor: rework the .css style for large buttons in dialog screens

Darxo:
Reduce text size for large buttons to 17 px (down from 18px) 
Form ellipsis for the text if it is too long to show 
Make text respect image (if present) and only take up remaining space

# Before this change
image

# After this change
image


TaroEld (approved):
Seems to work just fine. With "overflow: auto" on the label (and reverting the ellipsis stuff), one could actually have text scrolling for text... But there'd be no visual indicator.

LordMidas (approved):
This is Taro's domain, so if he says ok then sure.

PR #511, opened 2024-09-10
feat: add tip of the day for hotkey of perk group highlighting

Darxo:
I believe the hotkey was actually introduced in the dynamic perks framework.
So maybe we should:
- prefix the tip with **Dynamic Perks:**
- add this tip of the day instead from within the dynamic perks mod


LordMidas:
- Should be implemented in the Dynamic Perks Framework.
- Need feedback on this idea: The name of the key should ideally be pulled at runtime from the keybinds instead of hardcoding it as Shift.

Darxo:
This has been implemented into dynamic perks mod

PR #515, opened 2024-09-13
balance: adjust nomad spawns

Uberbagel:
reduce strength and cost of nomad outlaw
set startingresourcemin of nomad leader to 250

### DIFF
--- a/mod_reforged/hooks/config/spawnlist_master.nut
+++ b/mod_reforged/hooks/config/spawnlist_master.nut
@@ -13,10 +13,6 @@
 ::Const.World.Spawn.Troops.Wardog.Cost = 6; // vanilla 8
 ::Const.World.Spawn.Troops.ArmoredWardog.Strength = 8; // vanilla 9
 ::Const.World.Spawn.Troops.ArmoredWardog.Cost = 8; // vanilla 8
-::Const.World.Spawn.Troops.BanditThug.Strength = 9;
-::Const.World.Spawn.Troops.BanditThug.Cost = 9;
-::Const.World.Spawn.Troops.BanditMarksmanLOW.Strength = 12;
-::Const.World.Spawn.Troops.BanditMarksmanLOW.Cost = 12;
 ::Const.World.Spawn.Troops.BanditMarksman.Strength = 30; // vanilla 15
 ::Const.World.Spawn.Troops.BanditMarksman.Cost = 25; // vanilla 15
 ::Const.World.Spawn.Troops.BanditRaiderWolf.Strength = 30; // vanilla 25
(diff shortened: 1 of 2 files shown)


mod_reforged/hooks/config/spawnlist_master.nut
```diff
@@ -47,6 +47,11 @@
 ::Const.World.Spawn.Troops.Greatsword.TitleList <- ::Const.Strings.RF_ZweihanderTitles;
 ::Const.World.Spawn.Troops.Knight.Strength = 45;
 
+//Nomads
+::Const.World.Spawn.Troops.NomadOutlaw.Strength = 20;
+::Const.World.Spawn.Troops.NomadOutlaw.Strength = 20;
```
Darxo, line 52: Two times the same line

Darxo (changes requested):

mod_reforged/hooks/config/spawnlist_master.nut
Uberbagel: should've noticed that...

Darxo (approved):

mod_reforged/hooks/config/spawnlist_master.nut
```diff
@@ -47,6 +47,11 @@
 ::Const.World.Spawn.Troops.Greatsword.TitleList <- ::Const.Strings.RF_ZweihanderTitles;
 ::Const.World.Spawn.Troops.Knight.Strength = 45;
 
+//Nomads
+::Const.World.Spawn.Troops.NomadOutlaw.Strength = 20;
```
LordMidas, line 51: Please mention the vanilla cost/strength as a comment next to these changes.

mod_reforged/hooks/config/spawnlist_master.nut
LordMidas, line 1: Would be nice to have an overview of the design intention behind these changes. My inference is that this will make Outlaws appear sooner, and Leaders appear later. But isn't the south meant to spawn leaders earlier? This seems to go against that design of that faction.

LordMidas (changes requested):

mod_reforged/hooks/config/spawnlist_master.nut
Uberbagel: Your inference is correct. I didn't know/remember that leaders spawning earlier was our design intention for Nomads. Nomad Leaders aren't any weaker than Bandit Leaders. They would be comparable if not a little better than Bandit Leaders if not for their armor/helmet being able to lowroll. I'm drop their starting resource min to 200 instead of 250, so they will come earlier than bandit leaders and match this design of Nomad Leaders spawning earlier, but I think going below 200 would be a mistake. Bandit Leaders are 250 for comparison.

Uberbagel: And the Outlaw changes are necessary because currently they start showing up way too late and i did a strength comparison with bandits and I think 20 is reasonable for them.

LordMidas: We had a VC where we compared the spawns to bandits and we realized that Nomad Leaders spawn in roughly similar parties as Bandit party compositions. But we decided to go with 200 cost on Nomad Leaders to give them a bit of thematic difference. Also the Outlaw cost change seems reasonable considering their strength and upgrading in Dynamic Spawns so let's see how it plays it.

LordMidas (approved):

PR #516, opened 2024-09-15
feat: assign default reach depending on weapon type for modded weapons

Darxo:
When someone uses a vanilla mod, which adds new weapons, those will usually all have 1 Reach as that was our current default.
This branch employs simple weapon type and weapon range rules to assign much better reach values for unknown weapons.

### DIFF
--- a/mod_reforged/hooks/items/weapons/weapon.nut
+++ b/mod_reforged/hooks/items/weapons/weapon.nut
@@ -1,5 +1,16 @@
+::Reforged.HooksMod.hookTree("scripts/items/weapons/weapon", function(q) {
+	q.create = @(__original) function()
+	{
+		__original();
+		if (this.m.Reach == null)
+		{
+			this.assignDefaultReach();
+		}
+	}
+});
+
 ::Reforged.HooksMod.hook("scripts/items/weapons/weapon", function(q) {
-	q.m.Reach <- 1;
+	q.m.Reach <- null;
 
 	q.getTooltip = @(__original) function()
 	{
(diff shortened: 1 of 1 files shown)


mod_reforged/hooks/items/weapons/weapon.nut
```diff
@@ -1,5 +1,16 @@
+::Reforged.HooksMod.hookTree("scripts/items/weapons/weapon", function(q) {
+	q.create = @(__original) function()
+	{
+		__original();
+		if (this.m.Reach == null)
+		{
+			this.assignDefaultReach();
+		}
+	}
+});
```
Enduriel, line 10: I think reach should default to a value and then be adjustable within individual create functions, so someone could do something like

```squirrel
this.m.Reach -= 1
```

mod_reforged/hooks/items/weapons/weapon.nut
```diff
@@ -57,8 +68,99 @@
...
+				}
+				else if (this.isWeaponType(::Const.Items.WeaponType.Spear))
+				{
+					this.m.Reach = 5;
+				}
+				else if (this.isWeaponType(::Const.Items.WeaponType.Sword))
+				{
+					this.m.Reach = 4;
+				}
+			}
```
Enduriel, line 115: there should be an `else` here for 1 handed melee weapons

mod_reforged/hooks/items/weapons/weapon.nut
```diff
@@ -57,8 +68,99 @@
...
+				}
+				else if (this.isWeaponType(::Const.Items.WeaponType.Spear))
+				{
+					this.m.Reach = (this.m.RangeMax == 2) ? 6 : 5;
+				}
+				else if (this.isWeaponType(::Const.Items.WeaponType.Sword))
+				{
+					this.m.Reach = (this.m.RangeMax == 2) ? 6 : 6;
+				}
+			}
```
Enduriel, line 154: there should be an else here for 2 handed melee weapons

Enduriel (changes requested):
I think this would be a useful feature, I was also wondering if this would be possible to implement using something like itemtables so that we don't have the ugly if/else chain somehow, but I'm not sure. If you do think of something I think that would be a lot more readable and easier to maintain.

mod_reforged/hooks/items/weapons/weapon.nut
Darxo: Best I can think of for this case is to set the Reach to a negative value by default instead of `null`. For example -20
That way we can respect mods, which do simple addition with it.

But who would do such operations on an item, which they don't even know about?
And if they did it on a itemtype basis, they would also use a `hookTree` which should fire after `Reforged`, since that mod builds on Reforged

mod_reforged/hooks/items/weapons/weapon.nut
```diff
@@ -57,8 +68,99 @@
 		}
 	}
 
+// New Functions
 	q.getReach <- function()
 	{
 		return this.m.Reach;
 	}
+
+	q.assignDefaultReach <- function()
```
LordMidas, line 77: The logic in this function needs to be improved in a way that hybrid weapons are handled better. E.g. currently it will set a Dagger/Sword to 1 reach because it checks for Dagger first even though it might be a weapon similar to the Ancient Sword in Reforged.

Alternative logic could be:
- Start with 0 Reach.
- Each detected weapon type adds its own Reach value. E.g. Dagger adds 1, Axe adds 3, Sword adds 4, Spear adds 5 and so on.
- Then divide the total Reach by the total number of Weapon Types detected rounded to the nearest integer.
- Being Two-Handed adds 1 Reach if it isn't already at 5 Reach by now.
- Having a MaxRange of 2 adds 1 Reach if it isn't already at 6 Reach by now.

So in this case a Dagger/Sword will end up with 1+4 = 5 / 2 = 2.5 = 3 Reach. An Axe/Sword will end up with 3 + 4 = 7 / 2 = 3.5 = 4 Reach. An Axe/Hammer will end up with 3 + 3 = 6 / 2 = 3 Reach. And so on.

LordMidas (changes requested):

mod_reforged/hooks/items/weapons/weapon.nut
LordMidas: I can see the case for `null` or starting at 0.

mod_reforged/hooks/items/weapons/weapon.nut
Darxo: I would like to pass this advanced default reach algorithm implementation onto you or someone else. Maybe do it in another branch?

LordMidas: Ok. I'll implement it in a separate branch.

LordMidas: My implementation is now up in #526

LordMidas:
Closed in favor of #526

PR #517, opened 2024-09-15
fix: ensure that the throw net skill comes from an offhand item

LordMidas:
For Trip Artist and Kingfisher. Because these perks are meant to work with a net equipped in the offhand.

### DIFF
--- a/scripts/skills/perks/perk_rf_kingfisher.nut
+++ b/scripts/skills/perks/perk_rf_kingfisher.nut
@@ -19,7 +19,18 @@ this.perk_rf_kingfisher <- ::inherit("scripts/skills/skill", {
 	{
 		if (this.m.IsForceEnabled) return true;
 
-		return this.getContainer().hasSkill("actives.throw_net"); // Ignores hidden skills
+		// Ensure that the actor has an offhand item with the throw_net skill
+		local offhandItem = this.getContainer().getActor().getOffhandItem();
+		if (offhandItem == null)
+			return false;
+
+		foreach (skill in offhandItem.getSkills())
+		{
+			if (skill.getID() == "actives.throw_net" && !skill.isHidden())
+				return true;
+		}
+
+		return false;
 	}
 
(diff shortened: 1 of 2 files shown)


scripts/skills/perks/perk_rf_kingfisher.nut
```diff
@@ -19,7 +19,18 @@ this.perk_rf_kingfisher <- ::inherit("scripts/skills/skill", {
...
 
-		return this.getContainer().hasSkill("actives.throw_net"); // Ignores hidden skills
+		// Ensure that the actor has an offhand item with the throw_net skill
+		local offhandItem = this.getContainer().getActor().getOffhandItem();
+		if (offhandItem == null)
+			return false;
+
+		foreach (skill in offhandItem.getSkills())
+		{
+			if (skill.getID() == "actives.throw_net" && !skill.isHidden())
```
Darxo, line 29: Just for some context:
The compatibility issue stems from [a mod](https://www.nexusmods.com/battlebrothers/mods/69) which leaves nets equipped in the offhand after use, and disables (greys out) their ability.
That was the original reason for us to revisit this code.

Unfortunately we can't differentiate the reason for the net being unusable. So the weird behavior with that mod will remain. Until the owner bothers to hide the skill, instead of just disabling it.

Darxo (approved):
Your method is an improvement, even though it won't fix the reported behavior. But that's not a big problem

scripts/skills/perks/perk_rf_kingfisher.nut
LordMidas: Isn't the item number limit for Offhand Slot == 1? If yes, then how can that mod keep an item in that slot and still allow the player to equip some other item in that slot? That seems like a bad/dangerous move by that mod.

Darxo: Here is an image on how it looks ingame.
The used net is treated like an empty throwing weapon. It's still a full item that you need to manually unequip/swap with other items

image

PR #518, opened 2024-09-15
Fix hybridization tooltip error and skill validity check

LordMidas:


Enduriel (approved):
I now realize that what it was trying to do before with the tooltips didn't actually make any sense

Darxo (approved):

PR #525, opened 2024-09-17
refactor: improve weapon master for hybrid weapons

LordMidas:
With this setup for 2-way hybrid weapons if you have a perk from one of the weapon types' perk groups you get the perk from the other weapon type's perk group - instead of the alphabetically first one. However, if you have a perk from a perk group other than this weapon's weapon types, then you get the perks from this weapon's alphabetically first weapon type's perk group.

The solution isn't so pretty to look at in code - but it seems to work. An alternative approach is to iterate over all weapon perk groups in one go and then check if this item also has that weapon type that belongs to that perk group but that isn't necessarily going to be beautiful either as we have to somehow translate perk groups to weapon types which isn't always 1:1 (e.g. crossbow perk group applies to two weapon types Crossbow and Firearm).

So if you have better suggestions, I'd be happy to hear them out, otherwise we go with this.

### DIFF
--- a/scripts/skills/perks/perk_rf_weapon_master.nut
+++ b/scripts/skills/perks/perk_rf_weapon_master.nut
@@ -47,71 +47,106 @@ this.perk_rf_weapon_master <- ::inherit("scripts/skills/skill", {
 			return false;
 		}
 
-		local hasFirstPerk = false;
-		local hasSecondPerk = false;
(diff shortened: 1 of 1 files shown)


Darxo (approved):
Im just waving this through. 
It's a bit too complicated to [want to] wrap my head around it.

PR #526, opened 2024-09-18
feat: add automatic assignment of default reach for weapons

LordMidas:
This should cover all melee weapons added by mods (unless they use a WeaponType we are not aware of) and assign them an appropriate Reach value.

In my testing this produced very similar results for Reach for all weapons as to what we have manually set currently in Reforged.

### DIFF
--- a/mod_reforged/hooks/items/weapons/weapon.nut
+++ b/mod_reforged/hooks/items/weapons/weapon.nut
@@ -99,17 +99,17 @@
 		}
 		else
 		{
-			ret = ::Math.round(this.m.Reach / count);
+			ret = ::Math.round(ret / count);
 		}
 
 		if (this.isItemType(::Const.Items.ItemType.MeleeWeapon))
 		{
-			if (this.m.Reach < 5 && this.isItemType(::Const.Items.ItemType.TwoHanded))
+			if (ret < 5 && this.isItemType(::Const.Items.ItemType.TwoHanded))
 			{
 				ret = 5;
 			}
 
-			if (this.isAoE() || (this.getRangeMax() > 1 && this.m.Reach < 6))
+			if (this.isAoE() || (this.getRangeMax() > 1 && ret < 6))
 			{
 				ret += 1;
 			}
--- a/mod_reforged/hooks/items/weapons/weapon.nut
+++ b/mod_reforged/hooks/items/weapons/weapon.nut
@@ -2,7 +2,7 @@
 	q.create = @(__original) function()
 	{
 		__original();
-		if (this.m.Reach == 0 && this.isItemType(::Const.Items.ItemType.MeleeWeapon))
+		if (this.m.Reach == null)
 		{
 			this.assignDefaultReach();
 		}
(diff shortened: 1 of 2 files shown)


mod_reforged/hooks/items/weapons/weapon.nut
```diff
@@ -1,5 +1,16 @@
...
+		if (this.m.Reach == 0 && this.isItemType(::Const.Items.ItemType.MeleeWeapon))
+		{
+			this.assignDefaultReach();
+		}
+	}
+});
+
 ::Reforged.HooksMod.hook("scripts/items/weapons/weapon", function(q) {
-	q.m.Reach <- 1;
+	q.m.Reach <- 0;
```
Darxo, line 13: 0 is a valid reach value. Some mod might want to assign this reach to a melee weapon, but that would now be overwritten by assignDefaultReach

Darxo: I suggest `null` here.
If any mod builds acknowledges Reforgeds existance by adjusting Reach values it should also take the extra step of ordering itself after Reforged

LordMidas: I see pros and cons of both. If we go with `null` then we have to also assign `0` reach manually for ranged weapons in our `assignDefaultReach` function. If we go with `0` then the modder who wants 0 reach for a melee weapon could theoretically overwrite `assignDefaultReach` with an empty function in that weapon's file.

I am leaning more towards `null` personally. Or basically I am leaning more towards a non-zero value. So another idea could be to do `-1`. Then if someone sets it to `0` then it's fine and works out of the box. But the advantage of going with `-1` is that no one is thrown off by an exception due to `arithmetic operation between integer and null` if he tries to +/- the default reach. But there are edge cases with that as well where if someone modifies that base value then it may clash with what this function is trying to detect.

I'm happy to go with `null` for now. Let's see if @Enduriel has any other thoughts.

Enduriel: I actually prefer darxos idea of like -100. So that you can assume values >=0 are not meant to be changed and values <0 are meant to still be adjusted but this allows you to do things like this.m.Range -= 1

Darxo: Either way works for me. `null` seems a bit better to understand when [re]reading the code. And the negative number makes some reach manipulation by submods a bit easier

mod_reforged/hooks/items/weapons/weapon.nut
```diff
@@ -1,5 +1,17 @@
...
+		{
+			this.m.Reach += 999 + this.getDefaultReach();
+			this.m.Reach = ::Math.max(0, this.m.Reach);
+		}
+	}
+});
+
 ::Reforged.HooksMod.hook("scripts/items/weapons/weapon", function(q) {
-	q.m.Reach <- 1;
+	q.m.Reach <- -999;
```
Darxo, line 14: this could use a comment like 
`// arbitrary negative default value so we preserve simple addition and subtraction operations made by mods`

Darxo (approved):

Darxo (approved):

PR #533, opened 2024-09-21
fix: enemy-only perk devastating strikes missing a tooltip

Darxo:
It currently only appears on humans: Executioner, King, Highwayman, Killer

image

status_effect_06


Uberbagel:
I prefer the old reach advantage icon over this one. With either icon, I support this change as a temporary improvement until we make a more fitting icon ourselves.

LordMidas (approved):
I've added new art.

Uberbagel (approved):

PR #534, opened 2024-09-22
refactor(balance): reduce idealsize of barbarian parties by 20%

Darxo:
This is one method of addressing the barbarian spawn behavior

In Essence this change reduces the idealsize of barbarian parties by 20%.

However it also partially fixes a bug in the dynamic spawn framework. Because currently EVERY party has an idealsize of 12 (or 14/16 for manhunter/militia) and then 50% more for locations.
Idealsize was meant to be at most that number, but at least how many brothers the player actually has.


Uberbagel:
My understanding was that IdealSize is defaulted to 6 as shown in party.nut in DSF. And then add 50% for locations as you say.
image

Darxo:
@Uberbagel 
These three entries are mislreading and should probably be changed to `null` by default to clarify this.
They are not meant to be set by the party definition and instead get initialized using the member `DefaultResources` and function `getIdealSize` respectively.

image

Uberbagel:
I may have knew that and forgotten but that definitely changes things. Will think about our options on how to proceed.

Uberbagel:
> However it also partially fixes a bug in the dynamic spawn framework. Because currently EVERY party has an idealsize of 12 (or 14/16 for manhunter/militia) and then 50% more for locations.
We will be fixing this bug in the DSF framework I presume? So it correctly scales with player party size with a min or 6 and max of 12?

Uberbagel:
I'm pretty certain even with that bug fixed most factions will need custom ideal size rules like we did with bandits. I'll get on that.

LordMidas:
Yeah we'll fix the bug. For now you can simulate the fixed bug by setting `::DynamicSpawns.Const.MainMenuIdealSize = 7` in your main menu before you do party generation simulations. You can replace the 7 with any number you want the "default" IdealSize to be.

Darxo:
The barbarian spawn behavior has changed since this PR and are no longer a big issue

PR #536, opened 2024-09-23
refactor: replace morale progressbar on turn sequence bar with reach

Darxo:
When Taro did the same for the character screen he opted to send new Reach entries via the data structure to js.
For this change I decided to instead replace the existing morale values as I found that easier to implement and unlikely to cause issues.

This is how it looks now with this PR:
image

Here is how it lookes after doing Bagels issue
image

Here Reach looks in the character screen (before this PR):
image

### DIFF
--- a/mod_reforged/hooks/ui/screens/tactical/modules/turn_sequence_bar/turn_sequence_bar.nut
+++ b/mod_reforged/hooks/ui/screens/tactical/modules/turn_sequence_bar/turn_sequence_bar.nut
@@ -23,6 +23,22 @@
 		__original();
 	}
 
+	q.convertEntityToUIData = @(__original) function( _entity, isLastEntity = false )
+	{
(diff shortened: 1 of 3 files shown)


TaroEld (approved):
There's a higher chance for conflicts with this method but I suppose the odds are still rather slim

Darxo:
Instead of the word `Reach` I could also display offensive and defensive reach ignore here in brackets, like on the combat tooltip.
If so we also need to consider doing the same in the stats panel of the character screen

Uberbagel:
> Instead of the word `Reach` I could also display offensive and defensive reach ignore here in brackets, like on the combat tooltip. If so we also need to consider doing the same in the stats panel of the character screen

I prefer this approach. If not, I would still remove 'Reach' as I think it's unnecessary and adds clutter.

Uberbagel (approved):

PR #539, opened 2024-09-25
Feat reach base modifier values

Darxo:
Just like for the other attributes in the characterscreen before, this PR aims to add information about base Reach, weapon Reach, Reach Modifier and Offensive/Defensive Reachignore to the tooltip for Reach.

- For that, we need to send the entityID as an argument into the bindTooltip.
- I also had to remove an old, seemingly redundant bindTooltip for Reach, that would have otherwise overwritten my adjust bindTooltip with the entityID.
- Finally I hooked the MSU function `onQueryMSUTooltipData` because tooltips like that for Reach pass through that event, where I can then add my additional info to them

I have not tested whether the Reach Modifier field works, but the other four seem to work fine.

image

### DIFF
--- a/ui/mods/mod_reforged/js_hooks/screens/character/modules/character_screen_left_panel/character_screen_stats_module.js
+++ b/ui/mods/mod_reforged/js_hooks/screens/character/modules/character_screen_left_panel/character_screen_stats_module.js
@@ -19,13 +19,13 @@ CharacterScreenStatsModule.prototype.setProgressbarValues = function (_data)
 		}
 		else
 		{
-			_value.Row.bindTooltip({ entityId: entityID, contentType: 'ui-element', elementId: _value.TooltipId });
+			_value.Row.bindTooltip({ contentType: 'ui-element', elementId: _value.TooltipId, entityId: entityID });
 		}
 	});
 
 	$.each(this.mMiddleStatsRows, function (_key, _value)
 	{
-		_value.Row.bindTooltip({ entityId: entityID, contentType: 'ui-element', elementId: _value.TooltipId });
+		_value.Row.bindTooltip({ contentType: 'ui-element', elementId: _value.TooltipId, entityId: entityID });
 	});
 }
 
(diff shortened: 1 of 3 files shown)


ui/mods/mod_reforged/js_hooks/screens/character/modules/character_screen_left_panel/character_screen_stats_module.js
```diff
@@ -15,12 +15,12 @@ CharacterScreenStatsModule.prototype.setProgressbarValues = function (_data)
 	{
 		if (_value.TooltipId == "character-stats.Morale")
 		{
-			_value.Row.bindTooltip({ contentType: 'msu-generic', modId: Reforged.ID, elementId: "Concept.Reach" });
+			_value.Row.bindTooltip({ entityId: entityID, contentType: 'msu-generic', elementId: "Concept.Reach", modId: Reforged.ID });
```
TaroEld, line 18: Just move this around to to have the same order:
`{ contentType: 'msu-generic', modId: Reforged.ID, elementId: "Concept.Reach",  entityId: entityID}`

TaroEld (approved):
Approved after that little cosmetic thing

ui/mods/mod_reforged/js_hooks/screens/character/modules/character_screen_left_panel/character_screen_stats_module.js
Darxo: But the `else` directly below it also has `entityId: entityID` as the first argument. Should I change that too?

TaroEld: I think usually it's type, modid, elementid, extra content, so yeah, you can change those too. Thanks

Uberbagel:
This is very nice

TaroEld (approved):

PR #540, opened 2024-09-27
Uber spawn balance

Uberbagel:
This branch is not ready to be merged yet but I wanted to pull you both in to look it over early if you want. Additionally pending is check Barbarian Spawns after the Frost Unhold spawn issue is resolved and discussing what we want to do with orc spawns being too full of morale-breaking orc young.

### DIFF
--- a/scripts/mods/mod_reforged/dynamic_spawns/parties/barbarian_parties.nut
+++ b/scripts/mods/mod_reforged/dynamic_spawns/parties/barbarian_parties.nut
@@ -1,3 +1,5 @@
+// still need to figure out frost unhold spawns and adjust spawns after that resource value
+
 local parties = [
 	{
 		ID = "Barbarians",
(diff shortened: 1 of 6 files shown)


scripts/mods/mod_reforged/dynamic_spawns/parties/barbarian_parties.nut
```diff
@@ -21,9 +45,31 @@ local parties = [
 		VisibilityMult = 1.0,
 		VisionMult = 1.0,
 		UnitBlockDefs = [
-			{ BaseID = "UnitBlock.RF.BarbarianHunterFrontline", RatioMin = 0.60, RatioMax = 1.0, DeterminesFigure = true },
+			{ BaseID = "UnitBlock.RF.BarbarianFrontline", RatioMin = 0.60, RatioMax = 1.0, DeterminesFigure = true },
```
Darxo, line 48: Is it intended that you break with vanilla spawn behavior here?
`BarbarianHunters` only spawn thralls and dogs in vanilla

scripts/mods/mod_reforged/dynamic_spawns/parties/barbarian_parties.nut
```diff
@@ -9,9 +11,31 @@ local parties = [
...
-			{ BaseID = "UnitBlock.RF.BarbarianBeastmaster", RatioMin = 0.00, RatioMax = 0.10, ReqPartySize = 5 }		// Vanilla: Start spawning in armies of 7+ (singular case) but more like 9+
+			{ BaseID = "UnitBlock.RF.BarbarianDog", RatioMin = 0.00, RatioMax = 1.00, },		// Vanilla: Start spawning in armies of 6+
+			// want to add for dogs: if dogs in party >= 3 then reduce spawn chance by x (set RatioMax to 1.00 if we do this)
+			{ BaseID = "UnitBlock.RF.BarbarianBeastmaster", RatioMin = 0.00, RatioMax = 0.10, ReqPartySize = 5,  }		// Vanilla: Start spawning in armies of 7+ (singular case) but more like 9+
 		]
+
+		function generateIdealSize( _isLocation = false )
+		{
+			local startingResources = this.getSpawnProcess().getStartingResources();
+			if (startingResources >= 300)
```
Darxo, line 22: Considering how often you repeat the same three thresholds (300, 216, 164) we should define them globally somewhere so that we have them in one place. And submods can easier adjust them all at once

Something like `::Reforged.Const.DynamicSpawn.IdealSize.High`
@LordMidas

scripts/mods/mod_reforged/dynamic_spawns/parties/barbarian_parties.nut
Uberbagel: Yes, because our parties scale longer than vanilla I thought this would be preferable to massive thrall parties.

Darxo: How about we keep BarbarianHunterFrontline block but add new units to that?
That way the uniqueness of a BarbarianHunters party can still be preserved. It's weird to see fully armored Chosen moving around as "Hunters"

PR #542, opened 2024-10-01
docs: add vanilla values in comments

Uberbagel:


LordMidas (approved):
I didn't double check each value so I assume they are correct.

PR #551, opened 2024-10-04
Remove hooks on inherited backgrounds

LordMidas:
- Fixes the issue that shepherd southern gets the bonuses twice.
- I've taken the opportunity to remove hooks on other southern backgrounds where identical changes were present in our hook on the parent background. While in these cases it was benign, but it's still good to remove these.
- Please ensure that everything that I've removed is correct (i.e. identical hook exists on parent background) and let me know if I've missed any. 

@Uberbagel I haven't touched the 3 different slave backgrounds in this PR. Please take a look at them in Reforged and notice that they have slightly different multipliers for perk groups. Let me know if that is intended.


Darxo (approved):

Uberbagel:
> @Uberbagel I haven't touched the 3 different slave backgrounds in this PR. Please take a look at them in Reforged and notice that they have slightly different multipliers for perk groups. Let me know if that is intended.

Three different Slaves are intended, yes.

Uberbagel (changes requested):
Missed Southern Peddler

Uberbagel (approved):

PR #552, opened 2024-10-06
fix: properly resume generator function in hook on throw net behavior

LordMidas:


mod_reforged/hooks/ai/tactical/behaviors/ai_attack_throw_net.nut
```diff
@@ -4,6 +4,18 @@
...
+		{
+			local thread = ret;
+			do
+			{
+				ret = resume thread;
+			}
+			while (ret == null);
+		}
+
 		if (this.m.Skill != null && ret != ::Const.AI.Behavior.Score.Zero && ::Math.rand(1, 100) < 75 && _entity.getSkills().hasSkill("perk.rf_kingfisher"))
```
Darxo, line 19: Its not very performant to let the onEvaluate run through completely only to discard the result at the end on a random roll. Instead that random roll and the other checks here could be done once after the very first `resume`.

Example
```
q.onEvaluate = @(__original) function( _entity )
{
	local ret = __original(_entity);
	// It is a generator in vanilla, but we add the typeof check for maximum
	// compatibility if another mod changed the original function to not be a generator.
	if (typeof ret == "generator")
	{
		local thread = ret;
		ret = resume thread;
		if (this.m.Skill != null && ret != ::Const.AI.Behavior.Score.Zero && ::Math.rand(1, 100) < 75 && _entity.getSkills().hasSkill("perk.rf_kingfisher"))
			ret = ::Const.AI.Behavior.Score.Zero;
		while (ret == null);
		{
			yield null;
			ret = resume thread;
		}
	}
	else
	{
		if (this.m.Skill != null && ret != ::Const.AI.Behavior.Score.Zero && ::Math.rand(1, 100) < 75 && _entity.getSkills().hasSkill("perk.rf_kingfisher"))
			ret = ::Const.AI.Behavior.Score.Zero;
	}
	
	return ret;
```

mod_reforged/hooks/ai/tactical/behaviors/ai_attack_throw_net.nut
```diff
@@ -4,6 +4,18 @@
...
 	{
 		local ret = __original(_entity);
+		// It is a generator in vanilla, but we add the typeof check for maximum
+		// compatibility if another mod changed the original function to not be a generator.
+		if (typeof ret == "generator")
+		{
+			local thread = ret;
+			do
+			{
+				ret = resume thread;
```
Darxo, line 14: You do not call any `yield null` which means onEvaluate is now no longer a generator.
I think that will cause a runtime error in vanilla, because vanilla tries to call `resume` on the first value which is returned by `onEvaluate`

mod_reforged/hooks/ai/tactical/behaviors/ai_attack_throw_net.nut
```diff
@@ -4,6 +4,18 @@
 	q.onEvaluate = @(__original) function( _entity )
 	{
 		local ret = __original(_entity);
+		// It is a generator in vanilla, but we add the typeof check for maximum
+		// compatibility if another mod changed the original function to not be a generator.
```
Darxo, line 8: Are you sure, vanilla does not simply crash, when `onEvaluate` is no longer a generator? They call `resume` on its return value

Darxo (changes requested):

mod_reforged/hooks/ai/tactical/behaviors/ai_attack_throw_net.nut
Darxo, line 8: I don't think we should try to catch bad hooking by other mods and make them compatible.
If someone hooks a generator we should assume that it remains a generator afterwards

LordMidas:
Closed in favor of #619

PR #555, opened 2024-10-06
refactor: rework opportunist to reduce ap cost instead of ap recovery

LordMidas:
- Instead of halving the AP cost of the first 2 throwing attacks in the battle, it now halves the cost of all throwing attacks but only during the first round of battle.
- Instead of recovering 4 AP after stepping on a corpse, it now makes the next throwing attack cost no AP. (Still reduces its Fatigue cost by 50% like before).

The first point is somewhat of a sidegrade as you can now potentially unleash 4 throwing attacks during the first turn if you want if you don't move from your position. But it makes more sense thematically to me as an "opportunistic" playstyle and doesn't suffer from the weird situation we have currently where you are fatigued out after 4 rounds and still have the halved AP cost for the throwing attacks.

The second point reduces the silly chain of corpse hopping that can lead to extremely long turns as you keep recovering Action Points and may have other perks which reduce the cost of your attacks.

### DIFF
--- a/mod_reforged/hooks/config/strings.nut
+++ b/mod_reforged/hooks/config/strings.nut
@@ -1652,8 +1652,8 @@ foreach (vanillaDesc in vanillaDescriptions)
 		Effects = [{
 			Type = ::UPD.EffectType.Passive,
 			Description = [
-				"The first 2 throwing attacks during a combat have their [Action Point|Concept.ActionPoints] costs " + ::MSU.Text.colorPositive("halved") + ".",
-				"When using a throwing weapon which uses ammo, every time you stand over an enemy\'s corpse during your [turn,|Concept.Turn] gain " + ::MSU.Text.colorPositive(1) + " ammo and restore " + ::MSU.Text.colorPositive(4) + " [Action Points.|Concept.ActionPoints] Afterward, the next throwing attack has its [Fatigue|Concept.Fatigue] cost " + ::MSU.Text.colorPositive("halved") + ".",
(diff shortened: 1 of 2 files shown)


Uberbagel:
> The first point is somewhat of a sidegrade as you can now potentially unleash 4 throwing attacks during the first turn if you want if you don't move from your position. But it makes more sense thematically to me as an "opportunistic" playstyle and doesn't suffer from the weird situation we have currently where you are fatigued out after 4 rounds and still have the halved AP cost for the throwing attacks.

When does this weird situation happen? Only when the player doesn't utilize their 2 free throws and saves them for later?

LordMidas:
Yes. A guy could be attacking in melee but then use throwing attacks 4 rounds later and benefit from "opportunist". Which seems weird.

Uberbagel:
> Yes. A guy could be attacking in melee but then use throwing attacks 4 rounds later and benefit from "opportunist". Which seems weird.

It is a bit weird, but I think chances are small this will actually happen. I think the large majority of players will prefer to get that extra damage out early where it has a greater chance of positively impacting the fight.

I think half AP during the first round is both too restrictive and too strong at the same time. The player will only benefit from it if the enemy charges. Then if they do, that's potentially 4 attacks of value vs. 3 attacks from the current design. The enemy will only benefit from it if they have ranged advantage and the player charges. 

Perhaps a middle ground is half the cost of the first two throws for the first two rounds in combat.

LordMidas:
Yeah let's go with the restriction of first 2 rounds for now.

PR #556, opened 2024-10-06
balance: remove bullseye from ifrits

Uberbagel:


LordMidas (approved):

LordMidas:
Change is fine but the line should be commented out instead of deleted.

PR #560, opened 2024-10-13
feat: add mod setting to control visibility of blueprints in taxidermist

Darxo:
**Always** treats all Blueprints in the game as if you've already crafted them once before
**One Ingredient Available** treats all Blueprints in the game, for which you have at least one Ingredient, as if you've already crafted them once before
**All Ingredient Available** treats all Blueprints in the game, as if you've never crafted them once before
**Vanilla** applies the Vanilla filtering behavior

This hook preserves any overwritten Qualified-behavior like that of Snake Oil. 
Those recipes will not show up, no matter what setting is active, if those overwritten rules forbid it.

### DIFF
--- a/mod_reforged/hooks/crafting/crafting_manager.nut
+++ b/mod_reforged/hooks/crafting/crafting_manager.nut
@@ -1,0 +1,79 @@
+::Reforged.HooksMod.hook("scripts/crafting/crafting_manager", function(q) {
+	q.getQualifiedBlueprints = @(__original) function()
+	{
+		local settingValue = ::Reforged.Mod.ModSettings.getSetting("CraftingBlueprintVisibility").getValue();
+		switch (settingValue)
+		{
+			case "Vanilla":
+			{
+				return __original();
+			}
+			case "Always":
+			{
+				local ret = [];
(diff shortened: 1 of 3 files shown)


Uberbagel (approved):

mod_reforged/hooks/crafting/crafting_manager.nut
```diff
@@ -0,0 +1,75 @@
...
+						ret.push(b);
+					}
+					b.m.TimesCrafted = oldTimesCrafted;
+				}
+				return ret;
+			}
+			case "Vanilla":		// Do nothing, because this is equal to the catch-all return at the very end
+		}
+
+		return __original();
```
TaroEld, line 59: Move this return statement into the Vanilla case, and move it up top as the first option. This makes it clearer that on the "Vanilla" setting, we just run the original function.
Also add a default: case where you throw with a message that the case was not recognized. This is a failsave against future regressions.

TaroEld (changes requested):

TaroEld (approved):
I'd still like to see a default {throw "Unknown argument"} case.

PR #561, opened 2024-10-13
feat: attribute projection setting and attribute base range display

Darxo:
This PR does two things
- Add a Mod Setting for the already existing "Attribute Projection" tooltip in character backgrounds. With this the player can turn it on and off
- Add a new Base Attribute Range tooltip for backgrounds showing the minimum and maximum range of each Attribute. After hiring this will also show the current base attribute. In the Mod Settings the display can be changed.

## Mod Settings
image

### During Hiring
image

### After Hiring / Character Screen
image

### DIFF
--- a/mod_reforged/hooks/skills/backgrounds/character_background.nut
+++ b/mod_reforged/hooks/skills/backgrounds/character_background.nut
@@ -60,11 +60,15 @@
 	{
 		local ret = __original();
 
-		local player = this.getContainer().getActor();
-		if (::Const.XP.MaxLevelWithPerkpoints - player.getLevel() + player.getLevelUps() > 0)
+		if (::Reforged.Mod.ModSettings.getSetting("CharacterScreen_ShowAttributeProjection").getValue())
 		{
-			ret.extend(this.getProjectedAttributesTooltip());
+			local player = this.getContainer().getActor();
+			if (::Const.XP.MaxLevelWithPerkpoints - player.getLevel() + player.getLevelUps() > 0)
+			{
+				ret.extend(this.getProjectedAttributesTooltip());
+			}
 		}
+
 		return ret;
 	}
 })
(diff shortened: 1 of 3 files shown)


Uberbagel:
I'm not seeing the text for Paragon levels anywhere in this commit. Is that because you took that screenshot while using Hardened? This PR is great, but we've been using Veteran levels and I think that fits the game far better than Paragon levels.

Darxo:
> I'm not seeing the text for Paragon levels anywhere in this commit. Is that because you took that screenshot while using Hardened? This PR is great, but we've been using Veteran levels and I think that fits the game far better than Paragon levels.

"Paragon Levels" are a thing from my **Little Kingdom** origin mod. 
They have nothing to do with this PR and ideally I would have removed that mod before making the screenshots.

mod_reforged/hooks/skills/backgrounds/character_background.nut
```diff
@@ -53,18 +65,80 @@
...
+		local showBaseAttributeRangeRegular = ::Reforged.Mod.ModSettings.getSetting("CharacterScreen_ShowBaseAttributeRangesRegular").getValue();
+		if (showBaseAttributeRangeRegular == "Always")
+		{
+			ret.extend(this.getBaseAttributesTooltip(false));
+		}
+		else if (showBaseAttributeRangeRegular == "Only New Recruits")
 		{
-			ret.extend(this.getProjectedAttributesTooltip());
+			local player = this.getContainer().getActor();
+			if (player.getLevel() == player.getLevelUps() + 1)	// This condition is an approximation of "not yet having spent any level-up"
```
TaroEld, line 127: This isn't great; why not just show it whenever they have level ups available?

TaroEld (changes requested):
Mostly I'm wondering if the second option should just be "whenever levelups are available", the current version seems jank. If you really want it for only recruits, you should set a flag at hiring with the current level, I think.

Uberbagel (approved):

mod_reforged/hooks/skills/backgrounds/character_background.nut
Darxo: That goes against the advertised "Only New Recruits" option. 
Suddently these ranges would regularly show up on non-new recruits as well.

TaroEld (approved):

PR #562, opened 2024-10-14
fix(vanilla): missing isNull check during querySortedByItems

Darxo:
This fixes glitchy behavior when swapping equipped bandages This also fixes a rare crash during battle, when opening the inventory

### DIFF
--- a/mod_reforged/hooks/skills/skill_container.nut
+++ b/mod_reforged/hooks/skills/skill_container.nut
@@ -31,4 +31,21 @@
 	{
 		this.callSkillsFunctionWhenAlive("onSkillsUpdated", null, false);
 	}
+
+	// Vanilla Fix: A missing 'isNull' in Vanillas implementation causes crashes and glitchy behavior in some situations
+	q.querySortedByItems = @(__original) function( _filter, _notFilter = 0 )
+	{
+		// Instead of overwriting the original function, we clean the WeakTableRefs in every skill, that are no longer valid anyways
+		foreach (skill in this.m.Skills)
+		{
+			if (::MSU.isNull(skill.getItem()))
+			{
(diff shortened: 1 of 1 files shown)


mod_reforged/hooks/skills/skill_container.nut
```diff
@@ -31,4 +31,22 @@
...
 		this.callSkillsFunctionWhenAlive("onSkillsUpdated", null, false);
 	}
+
+	// Vanilla Fix: A missing 'isNull' in Vanillas implementation causes crashes and glitchy behavior in some situations
+	q.querySortedByItems = @(__original) function( _filter, _notFilter = 0 )
+	{
+		// Instead of overwriting the original function, we clean the WeakTableRefs in every skill, that are no longer valid anyways
+		foreach (skill in this.m.Skills)
+		{
+			if (skill.getItem() == null) continue;
```
TaroEld, line 41: This line is unnecessary, MSU.isNull will check that anyways

TaroEld (approved):

mod_reforged/hooks/skills/skill_container.nut
LordMidas, line 1: Are we sure this is actually a vanilla bug and not a bug caused by `mod_stack_based_skills`? Have you reproduced this bug in vanilla?

Darxo: I have not reproduced it in vanilla, especially the bandage swapping, which is the most consistent way of forcing it.

I'm sure this is a vanilla bug and `mod_stack_based_skills` is just the first mod to put this to light
It's not our fault that Vanilla uses its own WeakTableRefs but doesn't properly deal with situation of them containing a null reference.

PR #567, opened 2024-10-18
fix: add safety check for skill containers before spinning the flail

Darxo:
In a recent bug report the user crashed when a Fallen Hero spun the flail.
According to the log it looks like the Fallen Hero died to Riposte/Rebuke while spinning the flail.

image

If a dying actor instantly uncoupled from its skill container and/or deconstructs its actor object then that would explain, why we crash during the `onVerifyTarget` function.
That function assumes this skill to be attached to a skill container and that container to be attached to an actor.

image

My solution is to check for the coupling of the container before spinning the flail

### DIFF
--- a/scripts/skills/perks/perk_rf_flail_spinner.nut
+++ b/scripts/skills/perks/perk_rf_flail_spinner.nut
@@ -87,7 +87,11 @@ this.perk_rf_flail_spinner <- ::inherit("scripts/skills/skill", {
 	{
 		if (this.isEnabled() && _skill.isAttack() && _skill.m.IsWeaponSkill)
 		{
-			this.spinFlail(_skill, _targetTile);
+			// Our character might have perished from something like a counter attack
+			if (!::MSU.isNull(this.getContainer()) && !::MSU.isNull(this.getContainer().getActor()))
+			{
+				this.spinFlail(_skill, _targetTile);
+			}
 		}
 	}
 });
--- a/scripts/skills/perks/perk_rf_flail_spinner.nut
+++ b/scripts/skills/perks/perk_rf_flail_spinner.nut
@@ -44,7 +44,7 @@ this.perk_rf_flail_spinner <- ::inherit("scripts/skills/skill", {
 				this.getContainer().setBusy(true);
 				::Time.scheduleEvent(::TimeUnit.Virtual, 300, function ( perk )
 				{
-					if (targetEntity.isAlive())
+					if (user.isAlive() && targetEntity.isAlive())
 					{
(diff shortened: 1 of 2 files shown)


scripts/skills/perks/perk_rf_flail_spinner.nut
```diff
@@ -87,7 +87,11 @@ this.perk_rf_flail_spinner <- ::inherit("scripts/skills/skill", {
 	{
 		if (this.isEnabled() && _skill.isAttack() && _skill.m.IsWeaponSkill)
 		{
-			this.spinFlail(_skill, _targetTile);
+			// Our character might have perished from something like a counter attack
+			if (!::MSU.isNull(this.getContainer()) && !::MSU.isNull(this.getContainer().getActor()))
+			{
+				this.spinFlail(_skill, _targetTile);
+			}
```
LordMidas, line 94: This might be too early for this check. I think this should be moved into the scheduled event inside `spinFlail` because a counter-attack usually happens with a delay.

LordMidas (changes requested):

scripts/skills/perks/perk_rf_whirling_death.nut
```diff
@@ -144,7 +144,7 @@ this.perk_rf_whirling_death <- ::inherit("scripts/skills/skill", {
 			this.getContainer().setBusy(true);
 			::Time.scheduleEvent(::TimeUnit.Virtual, 100, function ( _perk )
 			{
-				if (target.isAlive())
+				if (user.isAlive() && target.isAlive())
```
Darxo, line 147: What do you make of the vanilla check `!actor.isDying()`, which I constantly see alongside the `isAlive` check?
Is it redundant, or just not needed in this particular case?
I dont know the difference/nuance between those two checks

LordMidas: I have not looked into it in maximum detail but as far as I looked the `IsDying = true` is set during `actor.kill` function slightly before `IsAlive = false`. So, it seems it is only relevant for certain skill events e.g. `onDeath` where `IsDying` will be true but `IsAlive` will be false.

For our use case here I used to do both `isAlive` and `isDying` checks following some vanilla style, but eventually I realized that the doing `isAlive` is enough. In my opinion the `isDying` check would be redundant and I have so far never encountered any issue in skipping it in skill_container events.

Vanilla also only checks for `target.isAlive()` in the delayed attacks from `cascade_skill.onUse`, so maybe for these cases the `isAlive` check is enough.

LordMidas (approved):

PR #568, opened 2024-10-21
refactor: play additional delayed sfx, on a multi hit with hail/cascade

Darxo:
The purpose of this change is to 
- give the player (who plays with audio) a bit more feedback about how many attacks did hit
- bring back a little bit of the original flavour of those skills, by making them seem to be multi-hits

I tried playing an sfx for each additionl hit, but they did completely overlap when the game briefly lags after killing someone. So I went with a single sfx as "good enough"


Darxo:
When playtesting this change felt weird. I will change it to draft for now

PR #570, opened 2024-10-26
refactor: implement maximum stacksize of 1 for retribution

Darxo:
I tested this ingame and it looks to work correctly.
Description of the perk is also adjusted.


TaroEld (commented):
Sorry, why not do this the clean way and remove the whole concept of "Stacks" from this skill if the max is 1 anyways?

PR #571, opened 2024-10-26
refactor: replace perma shieldwall on phalanx with shieldwall discount

Darxo:
I tested this ingame and the descriptions and effect seem to work correctly

image

Here the tooltip line spawned on **Shieldwall**
image

### DIFF
--- a/mod_reforged/hooks/config/strings.nut
+++ b/mod_reforged/hooks/config/strings.nut
@@ -1676,7 +1676,7 @@ foreach (vanillaDesc in vanillaDescriptions)
 			Type = ::UPD.EffectType.Passive,
 			Description = [
 				"Gain " + ::MSU.Text.colorPositive("+1") + " [Reach|Concept.Reach] per adjacent ally also equipped with a shield. Does not work with bucklers.",
-				"[Shieldwall|Skill+shieldwall_effect] does not expire at the start of your [turn|Concept.Turn] if an adjacent ally is also using [Shieldwall.|Skill+shieldwall_effect]"
+				"[Shieldwall|Skill+shieldwall] costs " + ::MSU.Text.colorPositive("50%") + " less [Action Points|Concept.ActionPoints] (up to a minimum of " + ::MSU.Text.colorPositive(2) + ") and " + ::MSU.Text.colorPositive("50%") + " less [Fatigue|Concept.Fatigue] if an adjacent ally has the [Shieldwall Effect.|Skill+shieldwall_effect]"
 			]
 		}]
 	}),
--- a/mod_reforged/hooks/config/strings.nut
+++ b/mod_reforged/hooks/config/strings.nut
@@ -1676,7 +1676,7 @@ foreach (vanillaDesc in vanillaDescriptions)
 			Type = ::UPD.EffectType.Passive,
 			Description = [
 				"Gain " + ::MSU.Text.colorPositive("+1") + " [Reach|Concept.Reach] per adjacent ally also equipped with a shield. Does not work with bucklers.",
-				"[Shieldwall|Skill+shieldwall] costs " + ::MSU.Text.colorPositive("50%") + " less [Action Points|Concept.ActionPoints] (up to a minimum of " + ::MSU.Text.colorPositive(2) + ") and " + ::MSU.Text.colorPositive("50%") + " less [Fatigue|Concept.Fatigue] if an adjacent ally has the [Shieldwall Effect.|Skill+shieldwall_effect]"
+				"[Shieldwall|Skill+shieldwall] costs " + ::MSU.Text.colorPositive("50%") + " less [Action Points|Concept.ActionPoints] (to a minimum of " + ::MSU.Text.colorPositive(2) + ") and " + ::MSU.Text.colorPositive("50%") + " less [Fatigue|Concept.Fatigue] if an adjacent ally has the [Shieldwall Effect.|Skill+shieldwall_effect]"
 			]
 		}]
 	}),
--- a/scripts/skills/perks/perk_rf_phalanx.nut
+++ b/scripts/skills/perks/perk_rf_phalanx.nut
@@ -122,7 +122,7 @@ this.perk_rf_phalanx <- ::inherit("scripts/skills/skill", {
 				id = 15,
 				type = "text",
 				icon = "ui/icons/special.png",
-				text = ::Reforged.Mod.Tooltips.parseString("Costs " + ::MSU.Text.colorizeMultWithText(this.m.ShieldWallActionPointsMult, {InvertColor = true}) + " [Action Points|Concept.ActionPoints] (to a minimum of " + ::MSU.Text.colorPositive(this.m.ShieldWallActionPointsMininum) + ") and " + ::MSU.Text.colorizeMultWithText(this.m.ShieldWallFatigueMult, {InvertColor = true}) + " [Fatigue|Concept.Fatigue]")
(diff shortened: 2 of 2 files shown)


mod_reforged/hooks/config/strings.nut
```diff
@@ -1676,7 +1676,7 @@ foreach (vanillaDesc in vanillaDescriptions)
 			Type = ::UPD.EffectType.Passive,
 			Description = [
 				"Gain " + ::MSU.Text.colorPositive("+1") + " [Reach|Concept.Reach] per adjacent ally also equipped with a shield. Does not work with bucklers.",
-				"[Shieldwall|Skill+shieldwall_effect] does not expire at the start of your [turn|Concept.Turn] if an adjacent ally is also using [Shieldwall.|Skill+shieldwall_effect]"
+				"[Shieldwall|Skill+shieldwall] costs " + ::MSU.Text.colorPositive("50%") + " less [Action Points|Concept.ActionPoints] (up to a minimum of " + ::MSU.Text.colorPositive(2) + ") and " + ::MSU.Text.colorPositive("50%") + " less [Fatigue|Concept.Fatigue] if an adjacent ally has the [Shieldwall Effect.|Skill+shieldwall_effect]"
```
Uberbagel, line 1679: I think this description should say:
_Shieldwall costs 50% less Action Points (to a minimum of 2) and 50% less Fatigue if an adjacent ally is also using Shieldwall._
- 'up to a minimum of 2' is improper because we're scaling down, not up. It's also unnecessary, so just remove it is best imo.
- 'if an adjacent ally has the Shieldwall Effect' sounds way too mechanical imo and we should stick with the original 'is also using Shieldwall'.

scripts/skills/perks/perk_rf_phalanx.nut
```diff
@@ -32,7 +36,7 @@ this.perk_rf_phalanx <- ::inherit("scripts/skills/skill", {
 				id = 11,
 				type = "text",
 				icon = "ui/icons/special.png",
-				text = ::Reforged.Mod.Tooltips.parseString("[Shieldwall|Skill+shieldwall_effect] will not expire at the start of your [turn|Concept.Turn] as you are next to an ally with [Shieldwall|Skill+shieldwall_effect]")
+				text = ::Reforged.Mod.Tooltips.parseString("[Shieldwall|Skill+shieldwall] costs " + ::MSU.Text.colorizeMultWithText(this.m.ShieldWallActionPointsMult, {InvertColor = true}) + " [Action Points|Concept.ActionPoints] (up to a minimum of " + ::MSU.Text.colorPositive(this.m.ShieldWallActionPointsMininum) + ") and " + ::MSU.Text.colorizeMultWithText(this.m.ShieldWallFatigueMult, {InvertColor = true}) + " [Fatigue|Concept.Fatigue]")
```
Uberbagel, line 39: Same comment as before about 'up to a minimum'.

scripts/skills/perks/perk_rf_phalanx.nut
```diff
@@ -114,13 +116,13 @@ this.perk_rf_phalanx <- ::inherit("scripts/skills/skill", {
...
 	{
-		if (_skill.getID() == "effects.shieldwall" && this.hasAdjacentShieldwall())
+		if (_skill.getID() == "actives.shieldwall" && this.hasAdjacentShieldwall())
 		{
 			_tooltip.push({
 				id = 15,
 				type = "text",
 				icon = "ui/icons/special.png",
-				text = ::Reforged.Mod.Tooltips.parseString("Will not expire as you have [Phalanx|Perk+perk_rf_phalanx] and are next to an ally with [Shieldwall|Skill+shieldwall_effect]")
+				text = ::Reforged.Mod.Tooltips.parseString("Costs " + ::MSU.Text.colorizeMultWithText(this.m.ShieldWallActionPointsMult, {InvertColor = true}) + " [Action Points|Concept.ActionPoints] (up to a minimum of " + ::MSU.Text.colorPositive(this.m.ShieldWallActionPointsMininum) + ") and " + ::MSU.Text.colorizeMultWithText(this.m.ShieldWallFatigueMult, {InvertColor = true}) + " [Fatigue|Concept.Fatigue]")
```
Uberbagel, line 125: Same comment as before about 'up to a minimum'.

Uberbagel (changes requested):

mod_reforged/hooks/config/strings.nut
Darxo: I need a second person to also tell me that this is "too mechanical", I don't see it

mod_reforged/hooks/config/strings.nut
```diff
@@ -1676,7 +1676,7 @@ foreach (vanillaDesc in vanillaDescriptions)
 			Type = ::UPD.EffectType.Passive,
 			Description = [
 				"Gain " + ::MSU.Text.colorPositive("+1") + " [Reach|Concept.Reach] per adjacent ally also equipped with a shield. Does not work with bucklers.",
-				"[Shieldwall|Skill+shieldwall_effect] does not expire at the start of your [turn|Concept.Turn] if an adjacent ally is also using [Shieldwall.|Skill+shieldwall_effect]"
+				"[Shieldwall|Skill+shieldwall] costs " + ::MSU.Text.colorPositive("50%") + " less [Action Points|Concept.ActionPoints] (to a minimum of " + ::MSU.Text.colorPositive(2) + ") and " + ::MSU.Text.colorPositive("50%") + " less [Fatigue|Concept.Fatigue] if an adjacent ally has the [Shieldwall Effect.|Skill+shieldwall_effect]"
 			]
```
TaroEld, line 1680: Optimally these texts would pull their values from the actual skill so it doesn't get outdated, but I don't know how feasible that is with the  current setup.

TaroEld: In fact I see down below that the text is duplicated; there should be one source of truth.

scripts/skills/perks/perk_rf_phalanx.nut
```diff
@@ -114,13 +116,13 @@ this.perk_rf_phalanx <- ::inherit("scripts/skills/skill", {
...
 	{
-		if (_skill.getID() == "effects.shieldwall" && this.hasAdjacentShieldwall())
+		if (_skill.getID() == "actives.shieldwall" && this.hasAdjacentShieldwall())
 		{
 			_tooltip.push({
 				id = 15,
 				type = "text",
 				icon = "ui/icons/special.png",
-				text = ::Reforged.Mod.Tooltips.parseString("Will not expire as you have [Phalanx|Perk+perk_rf_phalanx] and are next to an ally with [Shieldwall|Skill+shieldwall_effect]")
+				text = ::Reforged.Mod.Tooltips.parseString("Costs " + ::MSU.Text.colorizeMultWithText(this.m.ShieldWallActionPointsMult, {InvertColor = true}) + " [Action Points|Concept.ActionPoints] (to a minimum of " + ::MSU.Text.colorPositive(this.m.ShieldWallActionPointsMininum) + ") and " + ::MSU.Text.colorizeMultWithText(this.m.ShieldWallFatigueMult, {InvertColor = true}) + " [Fatigue|Concept.Fatigue]")
```
TaroEld, line 125: This should probably indicate where the effect comes from (like the former text)

TaroEld (approved):
Duplicating descriptions with hard-coded values is something that needs to get looked at at some point.

mod_reforged/hooks/config/strings.nut
Darxo: For `strings.nut` we don't have a standardized solution yet. It has always been hard-coded in here

PR #572, opened 2024-10-26
Refactor starting brother weapon group

Darxo:
The first commit is just refactoring, by encapsulating the weapon group identification logic, into a helper function.

The second commit will guarantee that all starting brothers from origins have at least one weapon group that works with their equipped weapon.
This approach makes sure that most custom perk tree changes by that origin are preserved.

### DIFF
--- a/mod_reforged/hooks/misc.nut
+++ b/mod_reforged/hooks/misc.nut
@@ -1,6 +1,41 @@
+::Reforged.Const <- {};
+
+::Reforged.Const.WeaponTypeToPerkGroupsMap <- {};
+
+// This needs to be adjusted if mods introduce or change weapon types and perk groups for those types
+::Reforged.Const.WeaponTypeToPerkGroupsMap[::Const.Items.WeaponType.Axe] <- "pg.rf_axe";
+::Reforged.Const.WeaponTypeToPerkGroupsMap[::Const.Items.WeaponType.Bow] <- "pg.rf_bow";
+::Reforged.Const.WeaponTypeToPerkGroupsMap[::Const.Items.WeaponType.Cleaver] <- "pg.rf_cleaver";
+::Reforged.Const.WeaponTypeToPerkGroupsMap[::Const.Items.WeaponType.Dagger] <- "pg.rf_dagger";
+::Reforged.Const.WeaponTypeToPerkGroupsMap[::Const.Items.WeaponType.Flail] <- "pg.rf_flail";
+::Reforged.Const.WeaponTypeToPerkGroupsMap[::Const.Items.WeaponType.Hammer] <- "pg.rf_hammer";
+::Reforged.Const.WeaponTypeToPerkGroupsMap[::Const.Items.WeaponType.Mace] <- "pg.rf_mace";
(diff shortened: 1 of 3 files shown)


Uberbagel (approved):

mod_reforged/hooks/misc.nut
```diff
@@ -4,3 +4,65 @@
 	if (_function != null) _function(obj);
 	return obj;
 }
+
```
TaroEld, line 7: I would prefer this to be a static map of {WeaponType : pg.id} and then a loop of the weaponTypes of the weapon.

TaroEld:
```
::Reforged.WeaponTypeToPerkGroupsMap <- {
	::Const.Items.WeaponType.Axe = "pg.rf_axe",
	::Const.Items.WeaponType.Bow = "pg.rf_bow",
	::Const.Items.WeaponType.Cleaver = "pg.rf_cleaver",
	::Const.Items.WeaponType.Dagger = "pg.rf_dagger",
	::Const.Items.WeaponType.Flail = "pg.rf_flail",
	::Const.Items.WeaponType.Hammer = "pg.rf_hammer",
	::Const.Items.WeaponType.Mace = "pg.rf_mace",
	::Const.Items.WeaponType.Polearm = "pg.rf_polearm",
	::Const.Items.WeaponType.Spear = "pg.rf_spear",
	::Const.Items.WeaponType.Sword = "pg.rf_sword",
	::Const.Items.WeaponType.Throwing = "pg.rf_throwing",
	::Const.Items.WeaponType.Crossbow ="pg.rf_crossbow",
	::Const.Items.WeaponType.Firearm ="pg.rf_crossbow",
} 

::Reforged.getPerkGroupsOfWeapon <- function( _weapon )
{
	local ids = [];

	if (_weapon == null || !_weapon.isItemType(::Const.Items.ItemType.Weapon))
	{
		return ids;
	}
	foreach(weaponType in _weapon.getWeaponTypes())
	{
		ids.push(weaponType);
	}
	return ids;
}

```
I don't know if getWeaponTypes exists.

mod_reforged/hooks/misc.nut
```diff
@@ -4,3 +4,65 @@
...
+// Return a list with ids of all weapon perk groups belonging to the given _weapon
+::Reforged.getWeaponPerkGroups <- function( _weapon )
+{
+	local ids = [];
+
+	if (_weapon == null || !_weapon.isItemType(::Const.Items.ItemType.Weapon))
```
TaroEld, line 13: Not sure if we should allow passing null here. It's more convenient, but kind of delays issues.

TaroEld (changes requested):
If it's refactored, might as well do it all the way.

mod_reforged/hooks/misc.nut
Darxo: I remove that check and emphasize in the documentation that the caller needs to do that check

mod_reforged/hooks/misc.nut
Darxo: Good suggestion, I included them in the new commits

TaroEld (approved):

PR #589, opened 2024-11-15
balance: manually set vision of most ranged enemies to 8 like vanilla

Uberbagel:
Will hopefully fix ranged enemy ai charging ahead and getting caught


TaroEld:
These should be done right in the base properties, for example faction_bandits.nut. You're setting the values on the base properties anyhow. Also, as you're doing it after the currentProperties are set (this.m.CurrentProperties = clone b), the currentProperties won't have the vision setting.

PR #591, opened 2024-11-15
Bagel spawn balance

Uberbagel:


scripts/mods/mod_reforged/dynamic_spawns/parties/barbarian_parties.nut
```diff
@@ -16,14 +16,19 @@ local parties = [
...
 			{
 				return 10;
 			}
+			else if (startingResources >= 216)
+			{
+				return 8;
+			}
 			else
 			{
 				return 8;
```
TaroEld, line 34: That's a lot of magic numbers, what are they based on? Shouldn't that be based on some entity cost?
Also last two conditions return the same value.

scripts/mods/mod_reforged/dynamic_spawns/parties/undead_parties.nut
```diff
@@ -13,29 +13,21 @@ local parties = [
...
-					else if (startingResources >= 120)
-					{
-						return 7;
-					}
 					else
 					{
-						return 5;
+						return 10;
 					}
 				}
```
TaroEld, line 33: Bit weird how <350 has a higher size than >=350

PR #593, opened 2024-11-20
fix: error when fetching tooltip for strenght in numbers

Darxo:
Move isPlacedOnMap checks into the getter functions, when the getTile() checks are actually required

Via hyperlinks you can view this tooltip for actors not placed on a map. This will then cause an error, which is catched by a try block and causes the following logs:

image


TaroEld (approved):
werks

PR #600, opened 2025-01-12
new random origins

Darxo:
Unlike vanilla origins, these two do not spawn an ingame intro event at the start.
I couldnt be bothered writing something together and those are purely flavor anyways, which is hard to do if the backgrounds are random.

It is possible to get the same background or name multiple times. I think that is totally fine
I tested both origins briefly and verified that the adjusted stars are having the expected effect

Requires: https://github.com/Battle-Modders/mod_modular_vanilla/pull/21


Uberbagel (approved):
Nice work Darxo

TaroEld (approved):

PR #602, opened 2025-01-16
Fix faulty logInfo call ai_rf_kata_step

Suor:
Causes a crash if `::Const.AI.VerboseMode` is set to true.


Darxo (changes requested):
The commit message should start with `fix: `, instead of `Fix ` 

The fix itself is good

Darxo (approved):

PR #612, opened 2025-01-26
balance: reduce melee defense malus of net from vanilla 45% to 25%

LordMidas:
Because in Reforged being netted reduces Reach as well.


Uberbagel (approved):

Darxo (approved):
I added some checks to allow us or mods to easier turn off some attribute changes completely

PR #613, opened 2025-01-26
refactor: improve hook shield

LordMidas:
- Allow using on targets using Shieldwall. Removes Shieldwall in this case.
- Second use on the same target refunds its AP cost.
- Increase Fatigue Cost from 15 to 25.

### DIFF
--- a/scripts/skills/actives/rf_hook_shield_skill.nut
+++ b/scripts/skills/actives/rf_hook_shield_skill.nut
@@ -138,7 +138,7 @@ this.rf_hook_shield_skill <- ::inherit("scripts/skills/skill", {
 				if (shieldWall != null)
 				{
 					::Tactical.EventLog.log(::Const.UI.getColorizedEntityName(_user) + " hooks " + ::Const.UI.getColorizedEntityName(targetEntity) + "\'s shield removing their " + shieldWall.m.Name);
-					targetEntity.getSkills().removeByID("effects.shieldWall");
+					targetEntity.getSkills().remove(shieldWall);
 				}
 				else
 				{
--- a/scripts/skills/actives/rf_hook_shield_skill.nut
+++ b/scripts/skills/actives/rf_hook_shield_skill.nut
@@ -39,6 +39,7 @@ this.rf_hook_shield_skill <- ::inherit("scripts/skills/skill", {
 		this.m.IsActive = true;
 		this.m.IsTargeted = true;
 		this.m.IsIgnoredAsAOO = true;
+		this.m.IsTooCloseShown = true;
 		this.m.IsWeaponSkill = true;
 		this.m.ActionPointCost = 3;
 		this.m.FatigueCost = 25;
--- a/scripts/skills/actives/rf_hook_shield_skill.nut
+++ b/scripts/skills/actives/rf_hook_shield_skill.nut
@@ -1,6 +1,8 @@
 this.rf_hook_shield_skill <- ::inherit("scripts/skills/skill", {
 	m = {
-		ApplyAxeMastery = true
+		ApplyAxeMastery = true,
+		LastTargetID = null,
+		IsSpent = false
 	},
 	function isAxeMasteryApplied()
 	{
@@ -39,7 +41,7 @@ this.rf_hook_shield_skill <- ::inherit("scripts/skills/skill", {
 		this.m.IsIgnoredAsAOO = true;
 		this.m.IsWeaponSkill = true;
 		this.m.ActionPointCost = 3;
-		this.m.FatigueCost = 15;
+		this.m.FatigueCost = 25;
 		this.m.MinRange = 1;
 		this.m.MaxRange = 1;
 		this.m.AIBehaviorID = ::Const.AI.Behavior.ID.KnockOut;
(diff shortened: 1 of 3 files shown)


Uberbagel (approved):

scripts/skills/actives/rf_hook_shield_skill.nut
```diff
@@ -110,20 +112,38 @@ this.rf_hook_shield_skill <- ::inherit("scripts/skills/skill", {
...
+		return targetEntity.isArmedWithShield() && this.skill.onVerifyTarget(_originTile, _targetTile) && !targetEntity.getSkills().hasSkill("effects.rf_hooked_shield");
 	}
 
 	function onUse( _user, _targetTile )
 	{
 		local targetEntity = _targetTile.getEntity();
+
+		if (!this.m.IsSpent && targetEntity.getID() == this.m.LastTargetID)
+		{
+			_user.setActionPoints(::Math.min(_user.getActionPointsMax(), this.getActionPointCost()));
```
Darxo, line 124: `getActionPointCost` does not take into account the vanilla property `.getCurrentProperties().AdditionalActionPointCost`, see `function use` from the `skill.nut`

scripts/skills/actives/rf_hook_shield_skill.nut
```diff
@@ -53,13 +55,20 @@ this.rf_hook_shield_skill <- ::inherit("scripts/skills/skill", {
...
 			icon = "ui/icons/special.png",
-			text = ::Reforged.Mod.Tooltips.parseString("Applies the [Hooked Shield|Skill+rf_hooked_shield_effect] effect on the target")
+			text = ::Reforged.Mod.Tooltips.parseString("Removes [Shieldwall|Skill+shieldwall_effect] from the target. Otherwise applies the [Hooked Shield|Skill+rf_hooked_shield_effect] effect")
 		});
 
 		ret.push({
 			id = 11,
 			type = "text",
 			icon = "ui/icons/special.png",
+			text = ::Reforged.Mod.Tooltips.parseString("Once per [turn|Concept.Turn], when used against the same target twice, the second use refunds all of its [Action Point|Concept.ActionPoints] cost")
```
Darxo, line 65: the comma should move into the hyperlink, like this: `[turn,|Concept.Turn]`

scripts/skills/actives/rf_hook_shield_skill.nut
```diff
@@ -157,4 +177,20 @@ this.rf_hook_shield_skill <- ::inherit("scripts/skills/skill", {
...
 	}
+
+	function onTurnStart()
+	{
+		this.m.IsSpent = false;
+	}
+
+	function onGetHitFactors( _skill, _targetTile, _tooltip )
+	{
+		if (_skill != this || this.m.IsSpent || !_targetTile.IsOccupiedByActor || _targetTile.getID() != this.m.LastTargetID)
```
Darxo, line 188: `this.m.LastTargetID` contains the id of an entity but you compare it with the id of a tile here

Darxo (changes requested):

scripts/skills/actives/rf_hook_shield_skill.nut
LordMidas: I submitted a [bug report](https://battlebrothersgame.com/forums/topic/problems-with-getactionpointcost-and-similar-functions-in-skill-nut/) for that several years ago but Overhype took no action against it. So I have no idea whether that is intended behavior or not. In any case `getActionPointCost` is what we are supposed to use to get the AP cost of a skill - we don't really have a better choice. Unless of course we decide to change this vanilla behavior in Modular Vanilla as per my suggestion in the bug report.

LordMidas: An alternative approach we can use right now is use `onBeforeAnySkillExecuted` and `onAnySkillExecuted` to check for the AP cost of the skill.

scripts/skills/actives/rf_hook_shield_skill.nut
```diff
@@ -110,20 +113,38 @@ this.rf_hook_shield_skill <- ::inherit("scripts/skills/skill", {
...
 			this.spawnAttackEffect(_targetTile, ::Const.Tactical.AttackEffectSplitShield);
 			if (this.attackEntity(_user, targetEntity))
 			{
-				::Tactical.EventLog.log(::Const.UI.getColorizedEntityName(_user) + " hooks " + ::Const.UI.getColorizedEntityName(targetEntity) + "\'s shield");
-				targetEntity.getSkills().add(::new("scripts/skills/effects/rf_hooked_shield_effect"));
+				local shieldWall = targetEntity.getSkills().getSkillByID("effects.shieldwall");
+				if (shieldWall != null)
+				{
+					::Tactical.EventLog.log(::Const.UI.getColorizedEntityName(_user) + " hooks " + ::Const.UI.getColorizedEntityName(targetEntity) + "\'s shield removing their " + shieldWall.m.Name);
+					targetEntity.getSkills().removeByID("effects.shieldWall");
```
Darxo, line 141: You could use `(shieldWall.getID())` here or `.remove(shieldWall)`, to tie it closer to the already fetched shieldwall skill from before. Just in case we change it or allow more different shieldwall like effects to be the target

Darxo (approved):
With out without the comments its good to go

PR #614, opened 2025-01-26
feat: naked effect

LordMidas:
New effect that is present on all player characters.
- Resolve is reduced by 50% when not wearing body armor.
- Does not affect Barbarians or Wildmen.


Uberbagel (approved):

scripts/skills/special/rf_naked.nut
```diff
@@ -0,0 +1,38 @@
...
+	},
+	function create()
+	{
+		this.m.ID = "special.rf_naked";
+		this.m.Name = "Naked";
+		this.m.Description = "This character\'s private parts are fully exposed, making it a rather nervous situation to be in.";
+		this.m.Icon = "skills/rf_naked.png";
+		this.m.Type = ::Const.SkillType.Special | ::Const.SkillType.StatusEffect;
+		this.m.IsSerialized = false;
+		this.m.IsHidden = true;
```
Darxo, line 17: this line is redundant as you overwrite the isHidden function anyways

scripts/skills/special/rf_naked.nut
```diff
@@ -0,0 +1,38 @@
...
+		if (this.isEnabled())
+		{
+			_properties.BraveryMult *= this.m.BraveryMult;
+		}
+	}
+
+	function isEnabled()
+	{
+		local actor = this.getContainer().getActor();
+		return ::MSU.isKindOf(actor, "player") && ::MSU.isNull(actor.getBodyItem()) && this.m.ExcludedBackgrounds.find(actor.getBackground().getID()) == null)
```
Darxo: The `::MSU.isKindOf(actor, "player")` should be partially removed to allow this effect to work on anyone, if we or a mod decide to add it to other characters.
Right now it is even redundant because we only add this effect to player characters anyways.

I suggest do write the check like this:
```
	function isEnabled()
	{
		return ::MSU.isNull(actor.getBodyItem()) && (!::MSU.isKindOf(actor, "player") || this.m.ExcludedBackgrounds.find(actor.getBackground().getID()) == null);
	}
```

scripts/skills/special/rf_naked.nut
```diff
@@ -26,24 +26,29 @@ this.rf_naked <- ::inherit("scripts/skills/skill", {
...
+					text = obj.m.Name
+				};
+			});
 
 			ret.push({
 				id = 20,
 				type = "text",
 				icon = "ui/icons/special.png",
-				text = ::Reforged.Mod.Tooltips.parseString("The following backgrounds are unaffected: " + excluded)
+				text = "The following backgrounds are unaffected: "
```
Darxo, line 50: a comma is missing here at the end of the line

scripts/skills/special/rf_naked.nut
```diff
@@ -26,24 +26,29 @@ this.rf_naked <- ::inherit("scripts/skills/skill", {
...
-			foreach (bg in this.m.ExcludedBackgrounds)
-			{
-				excluded += format("[Img/gfx/%s|Skill+%s]", ::new("scripts/skills/backgrounds/" + bg).getIconColored(), bg);
-			}
+			local excluded = this.m.ExcludedBackgrounds.map(function( _filename) {
+				local obj = ::new("scripts/skills/backgrounds/" + _filename);
+				return {
+					id = 20,
+					type = "text",
+					icon = obj.getIconColored()
```
Darxo, line 41: comma is missing at the end of the line

Darxo (changes requested):

scripts/skills/special/rf_naked.nut
LordMidas: Alright I'll make the change. The condition would have to be a bit more involved in that case though, because `getBackground()` only exists in `player`. So I'd have to put that behind an `isKindOf` check.

Darxo: If we trust the interpreter to evaluate the first argument of || first, always, then we can trust that actor is a player, the moment that we check for `this.m.ExcludedBackgrounds.find(actor.getBackground().getID()) == null`

Anyways, your rework of this function is good

Darxo (approved):

scripts/skills/special/rf_naked.nut
LordMidas: Ah I didn't see that you had it in an `||` there. Yeah that would've worked.

PR #616, opened 2025-01-26
refactor: make flaming arrows a separate active skill

LordMidas:


Uberbagel (approved):

mod_reforged/hooks/config/strings.nut
```diff
@@ -1318,9 +1318,9 @@ foreach (vanillaDesc in vanillaDescriptions)
 		Fluff = "Burn them all!",
 		Requirement = "Bow",
 		Effects = [{
-			Type = ::UPD.EffectType.Passive,
+			Type = ::UPD.EffectType.Active,
 			Description = [
-				"A successful [Aimed Shot|Skill+aimed_shot] will now light the target tile on fire for 2 rounds and trigger an additional morale check for the target and adjacent enemies."
+				"Unlocks the [Flaming Arrow|Skill+rf_flaming_arrows_skill] skill with bows."
```
Darxo, line 1323: no need to mention bows here when they are already mentioned in the requirement for this whole perk. So it should be applied that this perks effect (adding a new skill) only works while a bow is equipped.

scripts/skills/actives/rf_flaming_arrows_skill.nut
```diff
@@ -0,0 +1,219 @@
+this.rf_flaming_arrows_skill <- ::inherit("scripts/skills/actives/aimed_shot", {
+	m = {
+		TargetTile = null,
+		RangeModifier = -2,
```
Darxo, line 4: Have you considered just adding this to `this.m.MaxRange` during the `create` function?
We inherit from aimed_shot so we already mimic any changes that mods might do to this value.

scripts/skills/actives/rf_flaming_arrows_skill.nut
```diff
@@ -0,0 +1,219 @@
...
+		this.m.MaxRange += this.m.RangeModifier;
+	}
+
+	function onUse( _user, _targetTile )
+	{
+		// One ammo is consumed during the subsequent aimed_shot.onUse, so this manual loop consumes the additional ammo
+		// Start the loop at 1 because we want to consume 1 less ammo as 1 is consumed by aimed shot function call.
+		for (local i = 1; i < this.m.AmmoConsumed; i++)
+		{
+			this.consumeAmmo();
```
Darxo, line 103: You could hook the `consumeAmmo` function instead and call the base function once per `this.m.AmmoConsumed`
Like this
```
q.consumeAmmo = @(__original) function()
{
	for (local i = 1; i <= this.m.AmmoConsumed; ++i)
	{
		__original();
	}
}
```

scripts/skills/actives/rf_flaming_arrows_skill.nut
```diff
@@ -0,0 +1,219 @@
...
+				{
+					if (tile.Properties.Effect != null)
+					{
+						::Tactical.Entities.removeTileEffect(tile);
+					}
+
+					tile.Properties.Effect = clone p;
+					local particles = [];
+
+					for( local i = 0; i < ::Const.Tactical.FireParticles.len(); i = ++i )
```
Darxo, line 194: formatting should be adjusted and the redundant `i = ` at the end removed

Darxo (changes requested):

scripts/skills/actives/rf_flaming_arrows_skill.nut
LordMidas: Unfortunately `this.m.MaxRange` is set by vanilla during `onAfterUpdate`. So we have to do this workaround where we adjust it at the end of `onAfterUpdate` by calling the vanilla `onAfterUpdate` of `aimed_shot`.

Darxo (approved):

PR #619, opened 2025-01-26
fix: throw net behavior with kingfisher

Suor:
The function was returning integer sporadically, which caused "trying resume a 'integer'" crash.

Also, we won't get score until we cycle through the original generator.


Darxo:
I believe this old PR is related and was supposed to fix this issue: #552

Suor:
There was performance issue stated, which is also a thing here and also out of scope. 

However, since checking skill and score doesn't change anything the check could be simply moved up.

LordMidas (approved):

PR #622, opened 2025-01-27
feat: always preserve dropped player weapons

Darxo:
In Vanilla: normal player weapons are never dropped if they have 11 or less condition unless you have the blacksmith retinue.
This PR will make it so weapons that were dropped from a player character always drop, even if they have 0 condition.

I have not yet tested this out ingame

We also need to verify what happens if you try to loot those 0 condition items again and equip them. Although this shouldnt be too buggy as that would also be possible with the blacksmith retinue in vanilla

### Blacksmith
image

### Simple Test with Lute
image

### DIFF
--- a/mod_reforged/hooks/items/weapons/weapon.nut
+++ b/mod_reforged/hooks/items/weapons/weapon.nut
@@ -71,6 +71,21 @@
 		}
 	}
 
+	q.isDroppedAsLoot = @(__original) function()
+	{
+		if (!this.item.isDroppedAsLoot())
+		{
+			return false;
+		}
+
+		if (this.m.LastEquippedByFaction == ::Const.Faction.Player || !::MSU.isNull(this.getContainer()) && ::MSU.isKindOf(this.getContainer().getActor(), "player"))
+		{
+			return true;	// Player Weapons now always drop guaranteed
+		}
+
+		return __original();
+	}
+
 	q.getReach <- function()
 	{
 		return this.m.Reach;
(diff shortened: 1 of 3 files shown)


mod_reforged/hooks/items/weapons/weapon.nut
```diff
@@ -71,6 +71,23 @@
...
+		{
+			return false;
+		}
+
+		local actor = this.getContainer().getActor();
+		// Same condition as vanilla has for isPlayer
+		if (this.m.LastEquippedByFaction == this.Const.Faction.Player || this.getContainer() != null && !::MSU.isNull(actor) && this.isKindOf(actor.get(), "player"))
+		{
+			return true;	// Player Weapons now always drop guranteed
+		}
```
LordMidas, line 86: You need to first check if container is null. Currently you check for that condition in the next `if` block but you already use `getActor` here assuming the container is not null. Also `null` check on actor is unnecessary as `isKindOf` will already return false if that is the case. So we can optimize this a bit.

```suggestion
		if (this.m.LastEquippedByFaction == ::Const.Faction.Player || !::MSU.isNull(this.getContainer()) && ::MSU.isKindOf(this.getContainer().getActor(), "player"))
		{
			return true;
		}
```

mod_reforged/hooks/items/weapons/weapon.nut
```diff
@@ -71,6 +71,23 @@
...
+
+		local ret = __original();
```
LordMidas, line 88: `return __original();`

LordMidas (changes requested):

mod_reforged/hooks/items/weapons/weapon.nut
```diff
@@ -71,6 +71,21 @@
...
+	q.isDroppedAsLoot = @(__original) function()
+	{
+		if (!this.item.isDroppedAsLoot())
+		{
+			return false;
+		}
+
+		if (this.m.LastEquippedByFaction == ::Const.Faction.Player || !::MSU.isNull(this.getContainer()) && ::MSU.isKindOf(this.getContainer().getActor(), "player"))
+		{
+			return true;	// Player Weapons now always drop guranteed
```
LordMidas, line 83: typo in comment: `guaranteed`

LordMidas (approved):

LordMidas (approved):

PR #623, opened 2025-01-27
fix: killer being inconsistent between kill and onDeath functions

LordMidas:
Because we were changing the killer in `onDeath` which is called from `kill`, it was causing `kill` to have a different killer than `onDeath`. In this commit's method, we change the killer during `kill` which has the advantage of keeping the killer consistent across both functions and also more clearly shows intent that we only consider player or player animal faction to be the killer if he did at least 50% damage.

Related bug report on Discord: https://discord.com/channels/1006908336991645757/1327651957657305099


LordMidas:
Closed in favor of #624

PR #625, opened 2025-01-28
fix: fallen heroes sometimes being forever stuck in place

Darxo:
This is a quick way to fix situations like these, where Fallen Heroes are stuck in place.
This is very likely because their AI thinks they can't do any action because of their low available Stamina and the increased travel cost from level 4 encumbrance; Despite them having a 0% fatigue multiplier on all actions.

Exhibit_A
Exhibit_B
Exhibit_C
Exhibit_D

### DIFF
--- a/scripts/skills/racial/rf_zombie_racial.nut
+++ b/scripts/skills/racial/rf_zombie_racial.nut
@@ -97,4 +97,9 @@ this.rf_zombie_racial <- ::inherit("scripts/skills/skill", {
 		baseProperties.FatigueEffectMult = this.m.FatigueEffectMult;
 		baseProperties.FatigueDealtPerHitMult += this.m.FatigueDealtPerHitMultModifier;	// In vanilla normal zombies don't have this bonus
 	}
+
+	function onUpdate( _properties )
+	{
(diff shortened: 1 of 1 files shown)


scripts/skills/racial/rf_zombie_racial.nut
```diff
@@ -97,4 +97,9 @@ this.rf_zombie_racial <- ::inherit("scripts/skills/skill", {
 		baseProperties.FatigueEffectMult = this.m.FatigueEffectMult;
 		baseProperties.FatigueDealtPerHitMult += this.m.FatigueDealtPerHitMultModifier;	// In vanilla normal zombies don't have this bonus
 	}
+
+	function onUpdate( _properties )
```
LordMidas, line 101: I think a cleaner solution is to remove Encumbrance from Undead units.

LordMidas: For this we can change how we add Encumbrance in `actor.onInit` and only add it if the entity isn't kind of `zombie` or `skeleton`. Alternatively, inside Encumbrance effect, we can make it not apply if `FatigueEffectMult == 0`.

Darxo: Okay, while I still see this bug returning even without encumbrance, im ok waiting until it happens.

PR #629, opened 2025-02-03
Fix command ai freeze

Darxo:
There are three `onVerifyTarget` in the old implementation. I couldnt replace the third one though as it simulates future tiles and isUsableOn always takes the skill owners current tile as the origin.

I tested this with a toggle of the before and after condition like this:
image

Then I pumpted the score for ai_command x100 and set its minimum range to 2 and maximum range to 1 = (impossible to use)

I did 3 (each) back&forth tests with both setups in a fight against 4 heralds
image

The marked herald in exactly this setup consistently froze the game with the old check.
After the check swap (see first screenshot), I got no more freezes here. Instead that guy tried to move elsewhere

This is not a complete test by all means but good enough for me


LordMidas (approved):
I don't fully understand why this fixes it, but if it does then let's use it. If there are new problems down the line we can take another look.

PR #630, opened 2025-02-03
refactor: hide decisive outside of combat

Darxo:
With the last release, we made it so decisive shows, no matter how many stacks you have, meant to help the player during combat to remember this perks restriction.
As a result it also shows outside of battle.

This PR adjusts the perk to always show during combat but never outside of it

### DIFF
--- a/scripts/skills/perks/perk_rf_decisive.nut
+++ b/scripts/skills/perks/perk_rf_decisive.nut
@@ -28,6 +28,12 @@ this.perk_rf_decisive <- ::inherit("scripts/skills/skill", {
 		return this.m.Stacks == 0 ? this.m.Name : this.m.Name + " (x" + this.m.Stacks + ")";
 	}
 
+	// During combat we always display this effect on player characters even with 0 stacks as a helpful reminder
+	function isHidden()
+	{
+		return !::Tactical.isActive() || this.m.Stacks == 0 && !this.getContainer().getActor().isPlayerControlled();
+	}
+
 	function getTooltip()
 	{
 		local ret = this.skill.getTooltip();


scripts/skills/perks/perk_rf_decisive.nut
```diff
@@ -28,6 +28,11 @@ this.perk_rf_decisive <- ::inherit("scripts/skills/skill", {
 		return this.m.Stacks == 0 ? this.m.Name : this.m.Name + " (x" + this.m.Stacks + ")";
 	}
 
+	function isHidden()
+	{
+		return !::MSU.Utils.hasState("tactical_state");
```
LordMidas, line 33: Why not use `return !this.getContainer().getActor().isPlacedOnMap();`? I think that is preferable and more in line with what we have been using for other skills. We can also take this opportunity to hide it from AI characters so I'd say hide it from non-player controlled characters too.

LordMidas (changes requested):

scripts/skills/perks/perk_rf_decisive.nut
Darxo: I added a commit to hide this effect on enemies who don't yet have any stacks

LordMidas: When doing nested tooltips for vanilla skills e.g. `spearwall` we ran into crashes when a nested tooltip for Spearwall was triggered within combat. Because that tooltip uses `::Tactical.isActive()` to determine if combat is happening and within that check it was checking zone of control of the actor. Because the dummy player isn't placed on map, this caused crash because it had no tile. If that tooltip was using `isPlacedOnMap()` instead then there would be no crash. Vanilla condition for reference:
```squirrel
if (this.Tactical.isActive() && !this.getContainer().getActor().getCurrentProperties().IsSpecializedInSpears && this.getContainer().getActor().getTile().hasZoneOfControlOtherThan(this.getContainer().getActor().getAlliedFactions()))
```

Therefore, I have a preference for `isPlacedOnMap` as a safer check. Why do you say it is a misuse of that check to determine are we in combat? Because as far as I know the only time an actor can be placed on map is when a combat is happening. I kinda agree that "technically" checking for tactical state is a more "correct" method, but it has a potential pitfall as described above. I don't think it'll cause any issue during `isHidden` though, so maybe in this case it doesn't matter. Perhaps you can replace it with `::Tactical.isActive()` instead of the MSU function?

What are your thoughts overall considering this?

Darxo: When we are gating something that can cause crashes, like Spearwall or many other positional skills then absolutely we need to use `isPlacedOnMap`.
Therefor whenever I see an `isPlacedOnMap` check being used in code, I should be able to infer, that positioning is the key reasoning for this check.

But that's not happening in `Decisive`. This PRs change wants to hide information outside of battle and `isPlacedOnMap` is only like 99% correct for this.
Moving characters and swallowed characters are the 1% exceptions. Mods might add more exceptions and might make it able to view tooltips of swallowed, or otherwise phased out brothers.

LordMidas: Sounds good. Then I think we should go with `::Tactical.isActive()` in this case (unless I am missing something which makes the MSU function better).

LordMidas (approved):

PR #631, opened 2025-02-04
fix: remove wrong figure from skeleton heavy bodyguard unit

Darxo:


scripts/mods/mod_reforged/dynamic_spawns/units/undead_units.nut
```diff
@@ -24,8 +24,7 @@ local units = [
 	},
 	{
 		ID = "Unit.RF.SkeletonHeavyBodyguard",
-		Troop = "SkeletonHeavyBodyguard",
-		Figure = "figure_zombie_03"
```
LordMidas, line 28: I think replace it with `figure_skeleton_03` because that's the honor guard figure. Also I have noticed that that figure is missing from `Unit.RF.SkeletonHeavy` so I'd say add that there too.

LordMidas (changes requested):

LordMidas (approved):
Looks good now. Commit message should be adapted to something along the lines of `fix: use correct figure for skeleton heavy units`.

PR #632, opened 2025-02-08
refactor: move all actual effects from PP to realized potential

Darxo:
- Promised Potential now has a 100% chance to succeed at level 11 if the user has the **Player Character** perk
- Promised Potential perk description has been simplified a lot by moving all the effects into the **Realized Potential Description**

I verified that this change works by setting the WillSucceed chance to 0 and printing the WillSucceed chance right before it is swapped from the PlayerCharacter trait.
Then I loaded a Random Solo with a pauper starter and confirmed that he succeeded at level 11

image

image

**Player Character** is technically a trait but I decided to just call it like this

### Old Image from initial PR
image

### DIFF
--- a/mod_reforged/hooks/config/strings.nut
+++ b/mod_reforged/hooks/config/strings.nut
@@ -1706,7 +1706,8 @@ foreach (vanillaDesc in vanillaDescriptions)
 			Type = ::UPD.EffectType.OneTimeEffect,
 			Description = [
 				"Upon reaching level 11, this perk has a " + ::MSU.Text.colorPositive("50%") + " chance of being replaced with [Realized Potential|Perk+perk_rf_realized_potential].",
-				"If unsuccessful, this perk is replaced by [Failed Potential|Perk+perk_rf_failed_potential] which does nothing."
+				"If unsuccessful, this perk is replaced by [Failed Potential|Perk+perk_rf_failed_potential] which does nothing.",
+				"A [Player Character|Skill+player_character_trait] has a " + ::MSU.Text.colorPositive("100%") + " chance to succeed."
 			]
 		}],
 		Footer = ::MSU.Text.colorNegative("This perk cannot be picked after you have spent a perk point elsewhere. This perk cannot be refunded.")
--- a/mod_reforged/hooks/config/strings.nut
+++ b/mod_reforged/hooks/config/strings.nut
@@ -1705,7 +1705,7 @@ foreach (vanillaDesc in vanillaDescriptions)
 		Effects = [{
 			Type = ::UPD.EffectType.OneTimeEffect,
 			Description = [
-				"Upon reaching level 11, this perk has a " + ::MSU.Text.colorPositive("50%") + " chance of being replaced with [Realized Potential|Perk+perk_rf_realized_potential] which will double this character\'s salary, increase all attributes by " + ::MSU.Text.colorPositive("+15") + ", unlock new perk groups and refund all perk points, including the one spent on this perk.",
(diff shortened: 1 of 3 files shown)


Uberbagel:
We have an inconsistency with perk effect text. Sometimes the text ends with a period and other times it doesn't. Most often it ends without a period from my recollection.

I think what makes the most sense is perk _descriptions_ should end in periods whereas _effects_ should always end without.

scripts/skills/perks/perk_rf_promised_potential.nut
```diff
@@ -32,6 +32,11 @@ this.perk_rf_promised_potential <- ::inherit("scripts/skills/skill", {
 			local perkTree = actor.getPerkTree();
 			perkTree.removePerk(this.getID());
 
+			if (this.getContainer().hasSkill("trait.player"))
+			{
+				this.m.WillSucceed = true;
+			}
```
LordMidas, line 38: This should be a separate commit.

mod_reforged/hooks/config/strings.nut
```diff
@@ -1731,7 +1732,32 @@ foreach (vanillaDesc in vanillaDescriptions)
...
+					"Gain " + ::MSU.Text.colorPositive("1") + " new random Shared Perk Group.",
+					"Gain " + ::MSU.Text.colorPositive("2") + " new random Weapon Perk Groups.",
+					"Gain " + ::MSU.Text.colorPositive("+15") + " [Melee Skill.|Concept.MeleeSkill]",
+					"Gain " + ::MSU.Text.colorPositive("+15") + " [Melee Defense.|Concept.MeleeDefense]",
+					"Gain " + ::MSU.Text.colorPositive("+15") + " [Ranged Skill.|Concept.RangedSkill]",
+					"Gain " + ::MSU.Text.colorPositive("+15") + " [Ranged Defense.|Concept.RangedDefense]",
+					"Gain " + ::MSU.Text.colorPositive("+15") + " [Hitpoints.|Concept.Hitpoints]",
+					"Gain " + ::MSU.Text.colorPositive("+15") + " [Maximum Fatigue.|Concept.Stamina]",
+					"Gain " + ::MSU.Text.colorPositive("+15") + " [Initiative.|Concept.Initiative]",
+					"Gain " + ::MSU.Text.colorPositive("+15") + " [Resolve.|Concept.Bravery]"
```
LordMidas, line 1757: All of these should be combined into a single bullet point: `All attributes are increased by +15`.

mod_reforged/hooks/config/strings.nut
```diff
@@ -1731,7 +1732,32 @@ foreach (vanillaDesc in vanillaDescriptions)
...
+		Effects = [
+			{
+				Type = ::UPD.EffectType.Passive,
+				Description = [
+					"Require " + ::MSU.Text.colorNegative("100%") + " more daily wage.",
+				]
+			},
+			{
+				Type = ::UPD.EffectType.OneTimeEffect,
+				Description = [
```
LordMidas, line 1745: As part of this change, the logic inside `onUpdateLevel` of `perk_rf_promised_potential` must be moved to `onAdded` of `perk_rf_realized_potential` with the `if (!this.m.IsNew)` check. Then all the promised potential perk does is add the realized potential perk and the latter handles all the changes to the character, as described in its tooltip.

LordMidas (changes requested):

Uberbagel (approved):
Approved pending whatever outcome is determined regarding fullstops for perk effects.

LordMidas (approved):

PR #635, opened 2025-02-08
refactor: improve rush of battle stacks management

LordMidas:
Stacks gained during each turn are tracked separately and expire at the end of their next turn.

### DIFF
--- a/scripts/skills/perks/perk_rf_the_rush_of_battle.nut
+++ b/scripts/skills/perks/perk_rf_the_rush_of_battle.nut
@@ -1,6 +1,9 @@
 this.perk_rf_the_rush_of_battle <- ::inherit("scripts/skills/skill", {
 	m = {
-		Stacks = 0,
+		// We use two pools because each pool is tracked separately and expires at the end of the next turn
+		StackPool1 = 0,
+		StackPool2 = 0,
+		CurrPool = true, // Which pool is currently gaining stacks. true is pool 1 and false is pool 2. Flipped onTurnEnd.
 		BonusPerStack = 5,
 		MaxStacks = 5
 	},
@@ -17,7 +20,8 @@ this.perk_rf_the_rush_of_battle <- ::inherit("scripts/skills/skill", {
 
 	function getName()
 	{
-		return this.m.Stacks > 1 ? this.m.Name + " (x" + this.m.Stacks + ")" : this.m.Name;
+		local stacks = this.getStacks();
+		return stacks > 1 ? this.m.Name + " (x" + stacks + ")" : this.m.Name;
 	}
 
 	function getTooltip()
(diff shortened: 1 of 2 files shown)


Uberbagel (approved):

scripts/skills/perks/perk_rf_the_rush_of_battle.nut
```diff
@@ -1,6 +1,9 @@
 this.perk_rf_the_rush_of_battle <- ::inherit("scripts/skills/skill", {
 	m = {
-		Stacks = 0,
+		// We use two pools because each pool is tracked separately and expires at the end of the next turn
+		StackPool1 = 0,
+		StackPool2 = 0,
+		CurrPool = true, // Which pool is currently gaining stacks. true is pool 1 and false is pool 2. Flipped onTurnEnd.
```
Darxo, line 6: Have you considered an approach where `StackPool1` is always the pool that expires next, and `StackPool2` is the one that expires one turn later?
`addStacks` always adds to `StackPool2`
`onTurnStart()` copies `StackPool2` into `StackPool1` and then sets `StackPool2` to 0

Then there is no need for `CurrPool` flip flop

Darxo (changes requested):

scripts/skills/perks/perk_rf_the_rush_of_battle.nut
LordMidas: That is a great suggestion, thanks. Will do it like that.

scripts/skills/perks/perk_rf_the_rush_of_battle.nut
```diff
@@ -88,13 +86,12 @@ this.perk_rf_the_rush_of_battle <- ::inherit("scripts/skills/skill", {
 
 	function addStacks( _num )
 	{
-		local pool = this.m.CurrPool ? "StackPool1" : "StackPool2";
-		this.m[pool] = ::Math.min(this.m.MaxStacks, this.m[pool] + _num);
+		this.m.NewStacks = ::Math.min(this.m.MaxStacks, this.m.NewStacks + 1);
```
Darxo, line 89: `MaxStacks`s meaning is weird during `addStacks`: So with a MaxStacks of 5, you can have a total of 10 internal stacks, 5 can expire soon and 5 will expire a little bit later. 

I don't think there is a need to enforce MaxStacks during adding, when we enforce it during `getStacks`.

scripts/skills/perks/perk_rf_the_rush_of_battle.nut
```diff
@@ -124,28 +121,23 @@ this.perk_rf_the_rush_of_battle <- ::inherit("scripts/skills/skill", {
...
-			this.m.StackPool1 = 0;
-		}
+		this.m.OldStacks = 0;
 	}
 
 	function onTurnStart()
 	{
-		// Any stacks gained before the start of the first turn should be in the first pool
+		// Any stacks gained before the start of the first turn should be considered NewStacks
 		if (::Time.getRound() != 1)
```
Darxo, line 130: what is this check for?
If you get hit before your first turn starts, I dont see why the effect of `Rush of Battle` should be artificially extended by 1 round. As far as I see, the effect would exactly like advertised even during round 1

scripts/skills/perks/perk_rf_the_rush_of_battle.nut
LordMidas: It helps with the calculation in `getTooltip` where we show how many stacks will be lost at the end of the turn.

scripts/skills/perks/perk_rf_the_rush_of_battle.nut
LordMidas: I wouldn't call it extending by 1 round. All that is happening is that stacks gained before the start of your first turn are considered the same as gaining them during your first turn - simply because your turn hasn't even started in the battle yet.

The way you propose is "technically" more correct, but I think it is probably OK to give the character some benefit in this regard. Otherwise any stacks gained before starting your first turn will expire at the end of your first turn, which might feel weird for players.

@Uberbagel What are your thoughts?

Uberbagel: I agree on giving the character the extra benefit. Rush of Battle has been considered weak and frustrating to use for awhile now. I'd rather go a little too strong than a little too weak for this rework.

PR #641, opened 2025-02-11
refactor: move most long_reach logic out of actor.nut into the perk

Darxo:


LordMidas (approved):
Looks good. I pushed a commit with some adjustments which you can squash into yours.

PR #642, opened 2025-02-13
fix: fix line break consistency

Uberbagel:


mod_reforged/hooks/config/character_names.nut
LordMidas, line 1: The commit message type should be `style:` because it is a purely whitespace change with no effect on code.

LordMidas (changes requested):

LordMidas (approved):

PR #643, opened 2025-02-14
refactor: rework phalanx

LordMidas:
- Require Melee Piercing Attack and either a Shield or Two-Handed Weapon.
- Gain +1 Reach per adjacent ally who fulfills these requirements.
- Ignore the surround malus from enemies up to the number of adjacent allies who fulfill these requirements.

### DIFF
--- a/scripts/skills/perks/perk_rf_phalanx.nut
+++ b/scripts/skills/perks/perk_rf_phalanx.nut
@@ -144,6 +144,6 @@ this.perk_rf_phalanx <- ::inherit("scripts/skills/skill", {
 
 	function isEnabled()
 	{
-		return this.isActorValid(_actor);
+		return this.isActorValid(this.getContainer().getActor());
 	}
 });
--- a/scripts/skills/perks/perk_rf_phalanx.nut
+++ b/scripts/skills/perks/perk_rf_phalanx.nut
@@ -116,15 +116,13 @@ this.perk_rf_phalanx <- ::inherit("scripts/skills/skill", {
 
 	function isActorValid( _actor )
 	{
-		local actor = this.getContainer().getActor();
-
-		local weapon = actor.getMainhandItem();
+		local weapon = _actor.getMainhandItem();
 		if (weapon == null)
 			return false;
 
 		if (weapon.isItemType(::Const.Items.ItemType.OneHanded))
 		{
-			if (!actor.isArmedWithShield() || actor.getOffhandItem().getID().find("buckler") != null)
+			if (!_actor.isArmedWithShield() || _actor.getOffhandItem().getID().find("buckler") != null)
 				return false;
 
 			if (this.m.ShieldIgnoresDamageTypeRequirement)
--- a/scripts/skills/perks/perk_rf_phalanx.nut
+++ b/scripts/skills/perks/perk_rf_phalanx.nut
@@ -1,8 +1,6 @@
 this.perk_rf_phalanx <- ::inherit("scripts/skills/skill", {
 	m = {
-		RequiredDamageType = ::Const.Damage.DamageType.Piercing,
-		ReachModifier = 1, // per valid ally
-		StartSurroundCountAtModifier = 1 // per valid ally
+		RequiredDamageType = ::Const.Damage.DamageType.Piercing
 	},
 	function create()
 	{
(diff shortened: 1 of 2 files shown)


scripts/skills/perks/perk_rf_phalanx.nut
```diff
@@ -45,45 +51,40 @@ this.perk_rf_phalanx <- ::inherit("scripts/skills/skill", {
 
 	function onAnySkillUsed( _skill, _targetEntity, _properties )
 	{
-		if (_targetEntity != null && _skill.isAttack() && !_skill.isRanged())
+		if (_skill.isAttack() && !_skill.isRanged() && (this.m.RequiredDamageType == null || _skill.getDamageType().contains(this.m.RequiredDamageType)))
 		{
-			_properties.Reach += this.getCount();
+			_properties.Reach += this.getAllyCount();
```
Darxo, line 56: This does not currently use your new `ReachModifier` variable.
I recommend adding a new `getReachModifier()` function, which utilizes `getAllyCount` internally.

Then you can use that also for the `isHidden` and `getTooltip` implementations

scripts/skills/perks/perk_rf_phalanx.nut
```diff
@@ -45,45 +51,40 @@ this.perk_rf_phalanx <- ::inherit("scripts/skills/skill", {
...
 		}
 	}
 
-	function onAfterUpdate( _properties )
+	function onUpdate( _properties )
 	{
-		if (this.hasAdjacentShieldwall())
+		if (this.isActorValid(this.getContainer().getActor()))
```
Darxo, line 70: I suggest adding a new `isEnabled()` function, which just calls `this.isActorValid(this.getContainer().getActor())` internally.
It does not safe much currently except making the code a bit more readable and it follows our growing standardization in skills, where such a function usually mirrors the requirement line in perks.

But in the near future we might divorce again the conditions for the perk being enabled on the user, and the conditions for an adjacent ally being eligable for counting. Then it will come in handy

scripts/skills/perks/perk_rf_phalanx.nut
```diff
@@ -45,45 +51,40 @@ this.perk_rf_phalanx <- ::inherit("scripts/skills/skill", {
...
 		}
 	}
 
-	function getCount()
+	function getAllyCount()
 	{
 		local ret = 0;
 		local actor = this.getContainer().getActor();
-		if (!actor.isPlacedOnMap() || !actor.isArmedWithShield() || actor.getOffhandItem().getID().find("buckler") != null)
+		if (!actor.isPlacedOnMap() || !this.isActorValid(actor))
```
Darxo, line 80: I don't think we should check `!this.isActorValid(actor)` in here. With that check the `getAllyCount` function would do more than you would expect from it.
We might want to display the effect you would get, if you were to swap back to a valid weapon setup or become un-disarmed (if that was be to added as a condition)

scripts/skills/perks/perk_rf_phalanx.nut
```diff
@@ -92,56 +93,24 @@ this.perk_rf_phalanx <- ::inherit("scripts/skills/skill", {
...
+		local actor = this.getContainer().getActor();
+		if (actor.isArmedWithShield())
 		{
-			_tooltip.push({
-				id = 15,
-				type = "text",
-				icon = "ui/icons/special.png",
-				text = ::Reforged.Mod.Tooltips.parseString("Costs " + ::MSU.Text.colorizeMultWithText(this.m.ShieldWallActionPointsMult, {InvertColor = true}) + " [Action Points|Concept.ActionPoints] (to a minimum of " + ::MSU.Text.colorPositive(this.m.ShieldWallActionPointsMininum) + ") and " + ::MSU.Text.colorizeMultWithText(this.m.ShieldWallFatigueMult, {InvertColor = true}) + " [Fatigue|Concept.Fatigue] because of " + ::Reforged.NestedTooltips.getNestedSkillName(this))
-			});
+			return actor.getOffhandItem().getID().find("buckler") == null;
```
Darxo, line 101: You don't check for the melee piercing attack in this path. As far as I see, you would enable this perk when wielding your fists with a shield or any weapon really

Darxo (changes requested):

scripts/skills/perks/perk_rf_phalanx.nut
LordMidas: I've changed it as you have suggested. It is the most "robust" implementation but I'd like to ask is it really worth it? Now we have 2x the performance hit in `isHidden()`. I'd say it is safe to assume that this perk works always with ally count by virtue of its design, so the `ReachModifier` and `StartSurroundCountAtModifier` being member variables and `isHidden()` only checking for `getAllyCount()` seems clean enough and more efficient to me.

Uberbagel (approved):

Darxo (approved):

PR #644, opened 2025-02-14
feat: add greatsword helm without faction ribbons

LordMidas:
Uses existing unused vanilla art.


Darxo (changes requested):
This should wait until we actually use it. 
Otherwise it might get forgotten or reverted without ever being used

LordMidas:
> This should wait until we actually use it. 
> Otherwise it might get forgotten or reverted without ever being used

It is going to be used immediately for swordmaster enemies who are currently using the faction version with a ribbon which is misleading for them.

Darxo (approved):

Uberbagel (approved):
I will make the commit to add this to swordmasters

PR #648, opened 2025-02-14
fix(vanilla): decapitate tooltip not accounting for all damage modifiers

LordMidas:


mod_reforged/hooks/skills/actives/decapitate.nut
```diff
@@ -4,4 +4,45 @@
...
+	{
+		if (_skill == this)
+		{
+			// This is so that the tooltip shows the intended values when viewing the tooltip of the skill
+			if (_targetEntity == null)
+			{
+				// So that the max HP damage appears x2 just like in vanilla
+				_properties.DamageTooltipMaxMult *= 2.0;
+				// The above property will make the armor damage appear x2 as well, so we halve it manually
+				_properties.DamageArmorMult *= 0.5;
```
Darxo, line 39: the minimum armor damage you deal might now be inaccurate, as its getting halved

mod_reforged/hooks/skills/actives/decapitate.nut
```diff
@@ -4,4 +4,45 @@
...
+			}
+			// Else clause is vanilla behavior
+			else
+			{
+				_properties.DamageRegularMult += 1.0 - _targetEntity.getHitpoints() / (_targetEntity.getHitpointsMax() * 1.0);
```
Darxo, line 44: I know this is vanilla implementation.
I don't like how they change the multiplier in an additive way. We should fix that in Reforged.
Not least just to make our tooltip more accurate. I would go as far as to say that vanilla tooltip is inaccurate, as soon as you have another source of damage multiplier

Darxo (changes requested):

Darxo:
The same bug also exists in
- crush_armor
- demolish_armor_skill
- ghoul_claws
These skills also manually display damage numbers in their tooltip, but base them off of currentProperties, which can sometimes be inaccurate

mod_reforged/hooks/skills/actives/decapitate.nut
LordMidas: Ah yes, you're right. I was hoping I could do this using some clever techniques and not have to rewrite the entire calculation like vanilla does in `getTooltip` but it seems there is no easy way around this.

LordMidas:
Closed in favor of the [fix](https://github.com/Battle-Modders/mod_modular_vanilla/pull/25) implemented in Modular Vanilla.

PR #649, opened 2025-02-15
refactor: clarify inspiring presence tooltip about same faction allies

LordMidas:
Helps clarify misunderstandings like this one: https://discord.com/channels/1006908336991645757/1337779290984546407/1340105413940609054

### DIFF
--- a/mod_reforged/hooks/skills/perks/perk_inspiring_presence.nut
+++ b/mod_reforged/hooks/skills/perks/perk_inspiring_presence.nut
@@ -24,6 +24,12 @@
 			icon = "ui/icons/special.png",
 			text = ::Reforged.Mod.Tooltips.parseString("Allies who start their [turn|Concept.Turn] adjacent to this character when they are [engaged|Concept.ZoneOfControl] in melee or are adjacent to an ally [engaged|Concept.ZoneOfControl] in melee will gain the [Feeling Inspired|Skill+rf_inspiring_presence_buff_effect] effect")
 		});
+		ret.push({
+			id = 11,
+			type = "text",
+			icon = "ui/icons/warning.png",
+			text = ::Reforged.Mod.Tooltips.parseString("Only same faction members are considered allies for this [perk|Concept.Perk]")
+		});
 		return ret;
 	}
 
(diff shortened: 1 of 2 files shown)


mod_reforged/hooks/skills/perks/perk_inspiring_presence.nut
```diff
@@ -24,6 +24,12 @@
 			icon = "ui/icons/special.png",
 			text = ::Reforged.Mod.Tooltips.parseString("Allies who start their [turn|Concept.Turn] adjacent to this character when they are [engaged|Concept.ZoneOfControl] in melee or are adjacent to an ally [engaged|Concept.ZoneOfControl] in melee will gain the [Feeling Inspired|Skill+rf_inspiring_presence_buff_effect] effect")
 		});
+		ret.push({
+			id = 11,
+			type = "text",
+			icon = "ui/icons/warning.png",
+			text = ::Reforged.Mod.Tooltips.parseString("Only same faction members are considered allies for this [perk|Concept.Perk]")
```
Darxo, line 31: This is the first time I see us using `faction member` as a term.
I like it as it is less confusing when these effects appear on enemies, but it differs from how you say `member of your company` in this same perks perk description

Do we want to switch to saying `ally from your faction` or `Only members of your faction are considered allies for this`?

Uberbagel: I agree company should be replaced with faction whichever variant we go with.

Uberbagel (approved):

mod_reforged/hooks/skills/perks/perk_inspiring_presence.nut
LordMidas: I prefer company in the perk tooltip, but I'll change it to faction.

LordMidas: However, we do use `company` quite a bit in a lot of other places in perk tooltips and it seems to fit and flow a lot better in the game's theme. I chose `faction` for the status effect as it is an in-combat thing and is present on enemies too, but for the perk tooltip I still prefer company.

Darxo (approved):

PR #650, opened 2025-02-16
fix(vanilla): scenarios without intro event locking up the game

Darxo:
Vanilla calls `setAutoPause(true)` at the start of `startNewCampaign` but never sets it to `false` again.
That does not cause problems in vanilla because every dialog or other window that might pop-up will always call a `this.setAutoPause(false);`, when they are closed and vanilla always spawns an intro event for every origin at the start of a new campaign.

### DIFF
--- a/mod_reforged/hooks/states/world_state.nut
+++ b/mod_reforged/hooks/states/world_state.nut
@@ -1,4 +1,10 @@
 ::Reforged.HooksMod.hook("scripts/states/world_state", function(q) {
+	q.startNewCampaign = @(__original) function()
+	{
+		__original();
+		this.setAutoPause(false);	// VanillaFix: Fix scenarios without intro event locking up the game
+	}
+
 	// Add functionality to allow using more vars in troop names e.g. for champions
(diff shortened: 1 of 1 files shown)


mod_reforged/hooks/states/world_state.nut
```diff
@@ -1,4 +1,10 @@
 ::Reforged.HooksMod.hook("scripts/states/world_state", function(q) {
+	q.startNewCampaign = @(__original) function()
+	{
+		__original();
+		this.setAutoPause(false);	// Vanilla Fix: Fix scenarios without intro event locking up the game
```
LordMidas, line 5: For vanilla fixes please label the comment as `VanillaFix` without space so we can always search for that term.

LordMidas (approved):

mod_reforged/hooks/states/world_state.nut
Darxo: We are using "Vanilla Fix" with a space more than we use it without one

LordMidas: Then all of those should be changed to without space.

LordMidas: I've changed all the occurrences in the code to `VanillaFix`.

PR #657, opened 2025-02-17
refactor: improve moddability and require dagger and piercing attack

LordMidas:
In Swift Stabs perk.

### DIFF
--- a/scripts/skills/perks/perk_rf_swift_stabs.nut
+++ b/scripts/skills/perks/perk_rf_swift_stabs.nut
@@ -4,7 +4,7 @@ this.perk_rf_swift_stabs <- ::inherit("scripts/skills/skill", {
 		RequiredDamageType = ::Const.Damage.DamageType.Piercing,
 		ActionPointCostModifier = -2,
 		ActionPointCostMin = 2,
-		IsSpent = true
+		IsInEffect = false
 	},
 	function create()
 	{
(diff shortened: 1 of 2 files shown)


Uberbagel (approved):

scripts/skills/perks/perk_rf_swift_stabs.nut
```diff
@@ -1,5 +1,9 @@
 this.perk_rf_swift_stabs <- ::inherit("scripts/skills/skill", {
 	m = {
+		RequiredWeaponType = ::Const.Items.WeaponType.Dagger,
+		RequiredDamageType = ::Const.Damage.DamageType.Piercing,
+		ActionPointCostModifier = -2,
+		ActionPointCostMin = 2,
 		IsSpent = true
```
Darxo, line 7: I would like a short explanation and/or renaming of this variable. 
Usually `IsSpent` is used for a perk that is only used once per turn.
Here it seems to be used to control whether this perk is in effect or not, which confuses me when trying to understand this code. And you seem to spend and un-spend this perk multiple times per turn

Darxo (approved):
You decide, whether and how you want to resolve my remaining comment

scripts/skills/perks/perk_rf_swift_stabs.nut
LordMidas: I'll change it to `IsInEffect`

PR #658, opened 2025-02-18
refactor: open combo to all skills and swap tier with calculated strikes

LordMidas:


Uberbagel (approved):

scripts/skills/perks/perk_rf_combo.nut
```diff
@@ -47,47 +49,44 @@ this.perk_rf_combo <- ::inherit("scripts/skills/skill", {
...
 	{
 		local actor = this.getContainer().getActor();
-		local usedSkillID = this.m.UsedSkillID;
-		if (actor.isPreviewing() && actor.getPreviewSkill() != null)
-		{
-			usedSkillID = actor.getPreviewSkill().getID();
-		}
-
-		if (usedSkillID != "")
+		if (this.m.IsInEffect || actor.isPreviewing() && actor.getPreviewSkill() != null)
```
Darxo, line 65: The if is missing a few checks.
Simply speaking, when 
- this perk `IsInEffect`, we apply a discount, unless the player is previewing a skill, because after that skill usage, this would no longer be in effect
- this perk is not `InEffect`, but the player is previewing a valid skill, then we show the discount
```suggestion
		if ((this.m.IsInEffect && (!actor.isPreviewing() || actor.getPreviewSkill() == null)) || !this.m.IsInEffect && actor.isPreviewing() && actor.getPreviewSkill() != null && actor.getPreviewSkill().getActionPointCost() != 0)
```

Darxo (changes requested):

Darxo (approved):

PR #671, opened 2025-02-20
fix: double grip free swap working opposite than intended

LordMidas:
The intention was to allow swapping to daggers in the bag to be free, but it was working the opposite way because vanilla recalculates and pays the item action cost AFTER the item action is complete. By that time the dagger is no longer in the bag so `item.isInBag()` returning false.

An ideal solution would be to add LastSlotType to items which holds information about the last slot they were equipped in (in addition to the CurrentSlotType that they have) and that could be used here.

But the change in this commit is intended to provide a middle-ground design while we don't have that tech.


mod_reforged/hooks/skills/special/double_grip.nut
```diff
@@ -3,6 +3,7 @@
 	// because we go through weapon types alphabetically and choose the first applicable type
 	q.m.CurrWeaponType <- null;
 	q.m.MeleeDamageMult_Dagger <- 1.2;
+	q.m.IsFreeSwapSpent <- true;
```
Darxo, line 6: The default value should be `false` to emphasize. `false` is what this variable has most of the time and which is the normal case. `true` is the temporary exception

Darxo (approved):

Uberbagel (approved):

PR #680, opened 2025-03-01
Apply Damage Resistances from Racial skills using DamageType probability of the skill

LordMidas:
Additionally the resistances are now also taken into account in `onBeingAttacked` which ensures that AI properly takes into account the resistances when calculating expected damage (from `skill.getExpectedDamage`).

### DIFF
--- a/mod_reforged/hooks/skills/racial/alp_racial.nut
+++ b/mod_reforged/hooks/skills/racial/alp_racial.nut
@@ -87,6 +87,27 @@
 		baseProperties.IsImmuneToPoison = true;
 	}
 
+	q.onBeingAttacked = @() function( _attacker, _skill, _properties )
+	{
+		local d = _skill.getDamageType();
+		if (d.contains(::Const.Damage.DamageType.Blunt))
+		{
+			if (_skill.isRanged())
+				_properties.DamageReceivedRegularMult *= 0.33;
+		}
+		else if (d.contains(::Const.Damage.DamageType.Piercing))
+		{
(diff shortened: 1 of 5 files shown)


Darxo:
I prefer
- more centralized functions for doing all the probability->partly damage reduction management. Right now there seems to be duplicate code everywhere
- not having to deal with damage type probabilities in the first place
  - I feel like it was never a conscious decision by us (reforged team) that we want/need damage type probabilities in the first place. It is just baggage, that came with MSU's decision to marry damage types closely with injury types
  
Im not very motivated to look into or understand this code. If damage type probability is to stay then I want it very far abstracted away, out of my sight. I dont want to have to deal with it, whenever I implement any damagetype related damage reduction

LordMidas:
> Right now there seems to be duplicate code everywhere

Each racial has its own custom reductions depending on the damage type, so centralizing it somehow isn't that straightforward or worth it in my opinion. Did you have some solution in mind?

> not having to deal with damage type probabilities in the first place

They are a thing that exists though. I think working with them can improve accuracy of both tooltips as well as design. A skill inflicting mixed damage e.g. 50% piercing and 50% cutting is something we can leverage to create better design. In any case, the damage reduction calculations need to be improved to account for mixed damage skills. This is important if, for example, someone wants to predict an "expected" damage resistance of an entity to a certain attack type and then show this in the hit factors tooltip. Or similarly, if the AI wants to predict expected damage via `skill.getExpectedDamage`.

PR #683, opened 2025-03-03
fix(vanilla): goblin wolfriders getting damage modified by wolf bite

LordMidas:
For "balancing" them we add Fast Adaptation and Double Strike perks to them.


Darxo (approved):

Darxo:
Does this bug also plague vanilla or are they circumventing it, e.g. by always using bite first?

LordMidas:
> Does this bug also plague vanilla or are they circumventing it, e.g. by always using bite first?

To the best of my knowledge it exists in vanilla and they are not circumventing it in any way.

Darxo (approved):

PR #684, opened 2025-03-03
feat: add mod setting to make non-targeted skills behave as targeted

LordMidas:
Causes all skills to behave as targeted skills so that non-targeted skills such as Shieldwall, Rally the Troops etc. require you to click them to enable the skill preview mode and then use those skills on your character. This has the benefit of allowing you to preview the costs of those skills.


Darxo (approved):
I dont see myself using this feature and people coming from vanilla will be used to instant-activation skills, so with it being `off` by default I dont mind the addition

TaroEld (approved):

LordMidas:
At the suggestion of Taro, I converted it to a keybind which is incredibly more usable than a mod setting.

PR #685, opened 2025-03-03
fix: kata step costs not being incrementally changeable by other skills

LordMidas:


Darxo (commented):
How does this work outside of combat? Will skills retain discounts from the last tile you were on, when the combat ends, until you start your next combat?

LordMidas:
> How does this work outside of combat? Will skills retain discounts from the last tile you were on, when the combat ends, until you start your next combat?

Yes but that doesn't affect anything because:
- The `getCostString` function is overwritten for Kata Step and Gain Ground to return a manual string when not placed on map. So the actual cost is ignored completely in this case.
- Move Under Cover is never available outside of combat.

If we really want to be pristine with this we can reset the costs back to original value `onCombatFinished`, but I just left that out as it wasn't necessary.

LordMidas:
Actually I realized that there was an issue that the base values of the skills before adding were not being respected. I've pushed a new commit that fixes that and by necessity also resets the costs back to their original values `onCombatFinished` - so that issue should be taken care of as well.

scripts/skills/actives/rf_kata_step_skill.nut
```diff
@@ -112,6 +115,12 @@ this.rf_kata_step_skill <- ::inherit("scripts/skills/skill", {
 		return ret;
 	}
 
+	function onAdded()
+	{
+		this.m.BaseActionPointCost = this.m.ActionPointCost;
+		this.m.BaseFatigueCost = this.m.FatigueCost;
```
Darxo, line 121:
```suggestion
		this.m.__BaseActionPointCost = this.m.ActionPointCost;
		this.m.__BaseFatigueCost = this.m.FatigueCost;
```

scripts/skills/actives/rf_move_under_cover_skill.nut
```diff
@@ -1,5 +1,8 @@
 this.rf_move_under_cover_skill <- ::inherit("scripts/skills/skill", {
-	m = {},
+	m = {
+		__BaseActionPointCost = 0,
+		__BaseFatigueCost = 0
```
Darxo, line 4: These two member are never set in this script, unlike with what you did in kata step `onAdded`

scripts/skills/actives/rf_gain_ground_skill.nut
```diff
@@ -68,6 +71,12 @@ this.rf_gain_ground_skill <- ::inherit("scripts/skills/skill", {
 		return ret;
 	}
 
+	function onAdded()
+	{
+		this.m.BaseActionPointCost = this.m.ActionPointCost;
+		this.m.BaseFatigueCost = this.m.FatigueCost;
```
Darxo, line 77:
```suggestion
		this.m.__BaseActionPointCost = this.m.ActionPointCost;
		this.m.__BaseFatigueCost = this.m.FatigueCost;
```

Darxo (changes requested):

PR #686, opened 2025-03-03
refactor: add tooltip about rooting to uproot skill from uproot zoc aoo

LordMidas:


mod_reforged/hooks/skills/actives/uproot_zoc_skill.nut
```diff
@@ -51,4 +51,18 @@
...
+	q.onQueryTooltip = @(__original) function( _skill, _tooltip )
+	{
+		// The ZOC skill is not shown in the actor's tactical tooltip in Reforged. So we add the "rooting" behavior to the tooltip of the active uproot skill.
+		if (_skill.getID() == "actives.uproot")
+		{
+			_tooltip.push({
+				id = 100,
+				type = "text",
+				icon = "ui/icons/special.png",
+				text = ::Reforged.Mod.Tooltips.parseString("Will [root|Skill+rooted_effect] the target upon a successful [attack of opportunity|Concept.ZoneOfControl]")
```
Darxo, line 64: While this is not technically correct (because this way the player might think, the ZoC attack also is a 3 tile long attack), I see it still as an improvement

Darxo (approved):

mod_reforged/hooks/skills/actives/uproot_zoc_skill.nut
LordMidas: Sorry, I don't understand. The tooltip is added to the main attack and it specifically mentions attack of opportunity which only happens within ZoC. I don't see the capacity for confusion for the player. Did I misunderstand something?

PR #695, opened 2025-03-11
fix: small schrat causing morale checks when killing itself with root

LordMidas:


mod_reforged/hooks/entity/tactical/enemies/schrat_small.nut
```diff
@@ -41,4 +41,23 @@
...
 	}
+
+	// Return 0 during the call from onOtherActorDeath when using my own skill to kill myself so that my death
+	// in this case does not trigger morale checks because vanilla returns early in onOtherActorDeath if `getXPValue() <= 1`
+	q.getXPValue = @(__original) function()
+	{
+		local infos = ::getstackinfos(2);
+		if (split(infos.src, "/").top() == "actor.nut")
+		{
+			// Just checking for infos.func is not enough because hooked functions appear as "unknown"
```
Darxo, line 52: On top of that function being "unknown", it might also not even be "onOtherActorDeath".
Instead the stack, we are viewing, could be that of a hooked `getXPValue` function, if it was hooked twice by active mods.

Darxo (commented):
Have you considered, doing a switcheroo of the schrat XP, before and after it kills itself during schrat small root skill execution?

image

LordMidas:
> Have you considered, doing a switcheroo of the schrat XP, before and after it kills itself during schrat small root skill execution?

Yes, I already mentioned this in the Discord thread. This has the drawback (in Reforged) that any damage you dealt to these small schrats before they kill themselves will no longer give you any XP from them, as the XP will be 0 during his death. Because in Reforged XP is gained via `onDeath` of actors.

mod_reforged/hooks/entity/tactical/enemies/schrat_small.nut
LordMidas: I see, yeah that's a concern. Although, in this particular case I think it's very unlikely that someone will hook `getXPValue()` of `schrat_small`. Maybe they will do a `hookTree` on `actor` for `getXPValue()` but that also seems low probability to me.

We can wait and try to think of more elegant solutions.

Darxo:
I see no issue whatsoever of setting the XP gained from small Schrats to 0 in the first place
I dont like how it would otherwise be an optimal strategy to purposefully spawn small schrats to maximize XP gain in those fights

PR #700, opened 2025-03-13
refactor: BanditMarksmanLOW UnitID into BanditPoacher

Darxo:
Initially we probably went with `BanditMarksmanLOW`, because that is the name that vanilla gave to Poachers in the `spawnlist_master`.

But that name is confusing because there is also the `bandit_marksman_low` in vanilla, which is a scaled down version of the marksman, but which we decided to not use in Reforged. Someone looking into Reforged code might see `BanditMarksmanLOW` in our Unit_Blocks, and associate it with `bandit_marksman_low.nut` script, which we also hook in Reforged

This change is purely to improve readability of our code, when we create more unitblocks using poachers


mod_reforged_AfterHooks/dynamic_spawns/unit_blocks/bandit_blocks.nut
```diff
@@ -33,7 +33,7 @@ local unitBlocks = [
 		ID = "UnitBlock.RF.BanditRanged",
 		DeterminesFigure = true,
 		UnitDefs = [
-			{ BaseID = "Unit.RF.BanditMarksmanLOW" },
+			{ BaseID = "Unit.RF.BanditPoacher" },
```
LordMidas, line 36: Overall I think this may be _fine_. However, I am a bit afraid because currently we follow a very nice convention which also enables a nice level of "search" function through the codebase i.e. `Unit.RF.TroopID` convention. This change will break it. What are your thoughts about that?

Darxo: Okay makes sense to not break that.
I thought about a smaller scale alternative:
- Adding a comment mentioning Brigand Poacher after the definition of this unit
```
	{
		ID = "Unit.RF.BanditMarksmanLOW",	// Brigand Poacher
		Troop = "BanditMarksmanLOW",
		Figure = "figure_bandit_01", // Vanilla bandit poacher figure
	},
```

PR #705, opened 2025-03-28
refactor: remove vanilla mention of bad terrain in hitfactors

Darxo:
Vanilla doesnt differentiate different types of bad terrain. It doesnt know whether this specific bad terrain actually influences your hitchance.
Also with our expanded tooltips, it is much easier to see the effects that the bad terrain has on you.
Currently the only bad terrain there is is swamp, which applies its effects continuously, being visible in the attributes.
*edit1:* Swamp also doesn't affect ranged skill, and yet it shows up as a "negative hitfactor", when you aim with a ranged skill onto an enemy while you stand in swamp

So I deem these entries as bloat/misleading

### DIFF
--- a/mod_reforged/hooks/skills/skill.nut
+++ b/mod_reforged/hooks/skills/skill.nut
@@ -30,6 +30,8 @@
 				case "Resistance against ranged weapons":	// This was never followed through for anything other than skeletons
 				case "Resistance against piercing attacks":	// This was never followed through for anything other than skeletons
 				case "Nighttime":			// Already displayed on the stats of the character and in the effects section
+				case "On bad terrain":		// Terrain effects usually have their effects viewable in the actor tooltip
+				case "Target on bad terrain":	// Terrain effects usually have their effects viewable in the actor tooltip
 					ret.remove(index);
 					break;
 


Uberbagel (approved):
I think it's nice to have a clear visual indicator that the bad terrain is where the lowering of attributes is coming from. I don't care too strongly either way though, so if it's bloating the tooltip too much or if Midas also wants to remove it, it's fine by me.

mod_reforged/hooks/skills/skill.nut
```diff
@@ -31,6 +31,8 @@
 				case "Resistance against ranged weapons":	// This was never followed through for anything other than skeletons
 				case "Resistance against piercing attacks":	// This was never followed through for anything other than skeletons
 				case "Nighttime":			// Already displayed on the stats of the character and in the effects section
+				case "On bad terrain":		// Terrain effects usually have every present effects viewable in the actor tooltip
+				case "Target on bad terrain":	// Terrain effects usually have every present effects viewable in the actor tooltip
```
LordMidas, line 35: reword it as `usually have their effects viewable` instead of "usually have every present effects viewable"

LordMidas (approved):

PR #706, opened 2025-03-28
feat: add artwork for empty bolas

Darxo:
The art was made by Von Krolok
The images themselves don't have any attribution at all, I assume as this art was trivial and uses vanilla art, thats not necessary

## Screenshot
image


Uberbagel (approved):

LordMidas (changes requested):
The art files still need attribution because even if they are simply an edit of vanilla files, someone still put in the effort to make the edit.

LordMidas (approved):

PR #712, opened 2025-04-01
Refactor remove named names

Darxo:
I noticed that "Reach" is used in 4 different named weapons as a name, so I reduced it to only be used for Bows.
I also removed "Bardiche" and "Voulge" as possible names

### DIFF
--- a/mod_reforged/hooks/config/item_names.nut
+++ b/mod_reforged/hooks/config/item_names.nut
@@ -1,4 +1,9 @@
 // Vanilla Adjustments
+::MSU.Array.removeValues(::Const.Strings.BillNames, [
+	"Bardiche",	// There is already a vanilla weapon with this name
+	"Voulge"	// There is already a reforged weapon with this name
+]);
+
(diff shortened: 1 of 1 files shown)


Uberbagel (commented):
Bardiche and Voulge removal is good.

Reach being duplicated I don't mind so much. I think Reach as a name is kind of lackluster anyways. Sometimes it can be cool and other times it can be lame.

mod_reforged/hooks/config/item_names.nut
```diff
@@ -1,5 +1,12 @@
 // Vanilla Adjustments
+::MSU.Array.removeValues(::Const.Strings.BillNames, [
+	"Reach",	// This is re-used multiple times. We restrict it to just Bows
```
LordMidas, line 3: Bardiche and Voulge removal are fine. No need to remove Reach from any of the names.

LordMidas (changes requested):

Uberbagel (approved):

PR #715, opened 2025-04-04
doc: mark Brigand Poacher in unit definition as it differs from ID

Darxo:


LordMidas (approved):
Looks good. The commit type for this is `docs:` not `doc:`, so please adjust that in the commit message and PR title before merging.

PR #717, opened 2025-04-05
feat: replace power shot with a new perk steady brace

LordMidas:


scripts/skills/perks/perk_rf_steady_brace.nut
```diff
@@ -0,0 +1,112 @@
...
+	{
+		local ret = this.skill.getTooltip();
+		local weapon = this.getContainer().getActor().getMainhandItem();
+		if (weapon.isWeaponType(::Const.Items.WeaponType.Crossbow))
+		{
+			ret.push({
+				id = 10,
+				type = "text",
+				icon = "ui/icons/action_points.png",
+				text = ::Reforged.Mod.Tooltips.parseString("Crossbow shots have " + ::MSU.Text.colorizeMult(1.0 + this.m.DirectDamageAddModifier, {AddSign = true}) + " armor penetration")
```
Darxo, line 45: I would use the term `attack` here or `ranged attack` here as we use those everywhere else and they directly translate to the two conditions we check in code: `isAttack` and `isRanged`.
`ranged attack` is more accurate but not really needed as crossbows and firearms have no melee attack yet.
```suggestion
				text = ::Reforged.Mod.Tooltips.parseString("Crossbow attacks have " + ::MSU.Text.colorizeMult(1.0 + this.m.DirectDamageAddModifier, {AddSign = true}) + " armor penetration")
```

scripts/skills/perks/perk_rf_steady_brace.nut
```diff
@@ -0,0 +1,112 @@
...
+			});
+		}
+		else if (weapon.isWeaponType(::Const.Items.WeaponType.Firearm))
+		{
+			ret.push({
+				id = 11,
+				type = "text",
+				icon = "ui/icons/action_points.png",
+				text = ::Reforged.Mod.Tooltips.parseString("Firearm shots have " + ::MSU.Text.colorizeValue(this.m.RangedSkillModifier, {AddSign = true, AddPercent = true}) + " chance to hit")
```
Darxo, line 54:
```suggestion
				text = ::Reforged.Mod.Tooltips.parseString("Firearm attacks have " + ::MSU.Text.colorizeValue(this.m.RangedSkillModifier, {AddSign = true, AddPercent = true}) + " chance to hit")
```

scripts/skills/perks/perk_rf_steady_brace.nut
```diff
@@ -0,0 +1,112 @@
...
+
+		local weapon = this.getContainer().getActor().getMainhandItem();
+		if (weapon == null)
+			return true;
+
+		foreach (s in weapon.getSkills())
+		{
+			if (this.isSkillValid(s))
+				return false;
+		}
```
Darxo, line 30: This could be extracted into an `isEnabled` function

scripts/skills/perks/perk_rf_steady_brace.nut
```diff
@@ -0,0 +1,112 @@
...
+				icon = "ui/icons/action_points.png",
+				text = ::Reforged.Mod.Tooltips.parseString("Firearm shots have " + ::MSU.Text.colorizeValue(this.m.RangedSkillModifier, {AddSign = true, AddPercent = true}) + " chance to hit")
+			});
+		}
+
+		ret.push({
+			id = 20,
+			type = "text",
+			icon = "ui/icons/warning.png",
+			text = ::Reforged.Mod.Tooltips.parseString("Will expire upon moving or changing your weapon")
```
Darxo, line 62: `changing your weapon` is not accurate, as this effect is also removed when you swap ammunition or move items between bag and ground, in its current implementation

Furthermore `changing` is a new term to me. Previously we used the term `swapping`

I suggest to use `swapping items` here or changing the implementation of `onPayForItemAction`

Darxo (changes requested):

scripts/skills/perks/perk_rf_steady_brace.nut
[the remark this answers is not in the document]
LordMidas: I just realized that even though the perk description says that it only applies to the first attack, the implementation doesn't actually enforce that. So I will adjust the description for now because making it expire after the first attack is rather complicated. And as these weapons _mostly_ only attack once per turn, it's not a big deal.

Darxo (approved):

PR #719, opened 2025-04-05
feat: add custom framework for golem dual weapons and skills

LordMidas:
- Dual weapons hold an instance of each weapon.
- Dual weapons now get assigned appropriate stats and weapon types based on the two weapons they are based off of.
- Golem skills use one of the instances of the weapons inside the dual weapon for its damage values.

### DIFF
--- a/mod_reforged/hooks_helper.nut
+++ b/mod_reforged/hooks_helper.nut
@@ -1,6 +1,8 @@
 ::Reforged.HooksHelper <- {
 	function dualWeapon( q )
 	{
+		// These are strings which are the script paths to the weapons in this dual wield weapon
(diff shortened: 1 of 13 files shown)


mod_reforged/hooks_helper.nut
```diff
@@ -0,0 +1,82 @@
+::Reforged.HooksHelper <- {
+	function dualWeapon( q )
+	{
+		q.m.Weapon1 <- null;
+		q.m.Weapon2 <- null;
```
Darxo, line 5: There should be a comment here describing what these variables need to contain in order for this weapon to work properly.
> Strings? Instanitated Weapons? Can they be empty?

Darxo (approved):
The code is alright. do the documenting comments as you see fit

PR #720, opened 2025-04-05
feat: drop loot only if doing at least half damage to an entity

LordMidas:


Darxo:
I dont see where you make sure that the items from the item_container drop according to our 50% rule

Darxo:
I suggest to build this whole 50% drop feature about a new helper function like this:
```
	q.isLootAssignedToPlayer <- function( _killer )
	{
		// Player controlled characters always drop loot, no matter who kills them
		if (this.isPlayerControlled()) return true;

		// Allies never drop loot for the player, even if the player did most of the damage to them
		if (this.isAlliedWithPlayer()) return false;

		local playerRelevantDamage = 0.0;
		if (::Const.Faction.Player in this.m.RF_DamageReceived)
			playerRelevantDamage += this.m.RF_DamageReceived[::Const.Faction.Player].Total;
		if (::Const.Faction.PlayerAnimals in this.m.RF_DamageReceived)
			playerRelevantDamage +=  this.m.RF_DamageReceived[::Const.Faction.PlayerAnimals].Total;

		// If player + player animals did at least 50% of total damage to this actor, they gain the loot , we set the _killer to null to ensure that the loot properly drops from this actor.
		// This is because vanilla drops loot if _killer is null or belongs to Player or PlayerAnimals faction.
		return (playerRelevantDamage / this.m.RF_DamageReceived.Total >= 0.5)
	}
```
This can now be used in two ways:
- redirect all vanilla/modded loot droppings
- use this function in new entities so simplify the drop condition

mod_reforged/hooks/entity/tactical/actor.nut
```diff
@@ -291,5 +296,29 @@
...
 });
+
+::Reforged.QueueBucket.Late.push(function() {
+	::Reforged.HooksMod.hookTree("scripts/entity/tactical/actor", function(q) {
+		q.getLootForTile = @(__original) function( _killer, _loot )
+		{
+			// Change _killer to null to guarantee loot drop as vanilla checks for killer being null or belonging to Player or PlayerAnimals faction.
+			// Warning: this may be problematic for a mod that hooks this function and expects _killer to be the actual killer.
+			// Note: We do this in Late bucket so that mods that queue their hookTree after Reforged still receive the null _killer.
+			return this.m.RF_CanDropLoot ? __original(null, _loot) : [];
```
Darxo, line 313: This method deletes loot which was otherwise meant to always drop, no matter what.
A theoretical example is the **Kraken**, although its items are currently dropped directly during the `onDeath` function

Darxo, line 313: Here is an alternative approach, that makes sure to always fail the generic checks, if the loot is not assigned to the player:
Note: it uses a helper function, that I posted as a generic comment
```
if (this.isLootAssignedToPlayer(_killer))
{
	return __original(null, _loot);	// We pass null, because that will always pass the generic vanilla conditions for dropping loot
}
else
{
	local dummyNonPlayer = ::MSU.getDummyPlayer();
	local dummyPlayerFaction = dummyNonPlayer.getFaction();
	dummyNonPlayer.setFaction(::Const.Faction.None);	// We set the faction to none, so we fail all generic vanilla conditions for dropping loot

	// Some loot will still be dropped, but that is then usually by design of that specific enemy/mod
	local ret = __original(dummyNonPlayer, _loot);

	dummyNonPlayer.setFaction(dummyPlayerFaction);	// Revert the dummy faction, because other mods might expect it to be of player faction

	return ret;
}
```

Darxo (changes requested):

mod_reforged/hooks/entity/tactical/actor.nut
```diff
@@ -310,15 +330,23 @@
 			// Change _killer to null to guarantee loot drop as vanilla checks for killer being null or belonging to Player or PlayerAnimals faction.
 			// Warning: this may be problematic for a mod that hooks this function and expects _killer to be the actual killer.
 			// Note: We do this in Late bucket so that mods that queue their hookTree after Reforged still receive the null _killer.
-			return this.m.RF_CanDropLoot ? __original(null, _loot) : [];
-		}
+			if (_killer == null || this.RF_canDropLootForPlayer(_killer))
```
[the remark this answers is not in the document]
LordMidas: Good catch. I'll remove the `_killer == null` check. However, I don't understand this part of your message:

> As a result, we need to use a dummy entity, when the _killer is null, to pass to the __original function.
I suggest using the ::MSU.getDummyPlayer()

The original function does not expect `_killer` to never be null, so it should be safe to just pass `null` there. No need for a dummy entity. Unless I'm missing something?

mod_reforged/hooks/entity/tactical/actor.nut
```diff
@@ -291,5 +318,35 @@
...
+			// Warning: this may be problematic for a mod that hooks this function and expects _killer to be the actual killer.
+			// Note: We do this in Late bucket so that mods that queue their hookTree after Reforged still receive the null _killer.
+			if (this.RF_canDropLootForPlayer(_killer))
+			{
+				return __original(null, _loot);
+			}
+
+			// Switcheroo getFaction to return None so that the generic vanilla faction check
+			// for loot fails i.e. _killer being of faction Player or PlayerAnimals.
+			local getFaction = _killer.getFaction;
```
Darxo, line 340: `_killer` might be `null` at this point.

We must find another dummy entity, to pass into `__original` with `none` as the faction, if we want to be able to turn off loot in every case

I recommend the dummy player

Darxo (changes requested):

Darxo (approved):

PR #722, opened 2025-04-06
feat: add new shield bash skill unlocked from line breaker perk

LordMidas:
Remove the addition of Stagger to knock back from Line Breaker.

Shield Bash has +25% chance to hit, does no damage, but applies Dazed on a hit.


scripts/skills/perks/perk_rf_line_breaker.nut
```diff
@@ -15,24 +15,20 @@ this.perk_rf_line_breaker <- ::inherit("scripts/skills/skill", {
...
 	function onRemoved()
 	{
 		this.getContainer().removeByID("actives.rf_line_breaker");
 	}
 
-	function onTargetHit( _skill, _targetEntity, _bodyPart, _damageInflictedHitpoints, _damageInflictedArmor )
+	function onEquip( _item )
 	{
-		if (_skill.getID() == "actives.knock_back" && _targetEntity.isAlive() && !_targetEntity.isDying())
+		if (::MSU.isKindOf(_item, "shield"))
```
Darxo, line 29: how does `isKindOf` work exactly?
Like what if there is a shield, that does not have `shield` in its classname?
And what about items, which do have `shield` in their name, but are not a shield? E.g. "Shieldbreaker"

scripts/skills/actives/rf_shield_bash_skill.nut
```diff
@@ -0,0 +1,102 @@
...
+			return false;
+
+		targetEntity.getSkills().add(::new("scripts/skills/effects/dazed_effect"));
+
+		if (!_user.isHiddenToPlayer() || _targetTile.IsVisibleForPlayer)
+		{
+			::Tactical.EventLog.log(::Const.UI.getColorizedEntityName(_user) + " has dazed " + ::Const.UI.getColorizedEntityName(targetEntity));
+		}
+
+		::Sound.play(::MSU.Array.rand(this.m.SoundOnHit), ::Const.Sound.Volume.Skill, targetEntity.getPos());
```
Darxo, line 80: I dont think we need this as `attackEntity` already spawns sound effects

scripts/skills/actives/rf_shield_bash_skill.nut
```diff
@@ -0,0 +1,102 @@
...
+
+		return true;
+	}
+
+	function onAfterUpdate( _properties )
+	{
+		if (_properties.IsProficientWithShieldSkills)
+		{
+			this.m.FatigueCostMult *= ::Const.Combat.WeaponSpecFatigueMult;
```
Darxo, line 89: No other shield skill has this and we don't mention this in the tooltip of shield expert

Darxo (changes requested):

scripts/skills/perks/perk_rf_line_breaker.nut
LordMidas: It means that `_item` must be an instance of the BB class `shield` or a descendant of it.

scripts/skills/actives/rf_shield_bash_skill.nut
LordMidas: These are different sound effects. `attackEntity` spawns the sound effect of the attack e.g. the woosh sound. These are sounds on hit which are only played on a successful hit.

scripts/skills/actives/rf_shield_bash_skill.nut
LordMidas: All shield skills have this in vanilla. This is from the oath of fortification.

Darxo (approved):

PR #725, opened 2025-04-06
Uber enemy balance

Uberbagel:


scripts/entity/tactical/enemies/rf_bandit_highwayman.nut
```diff
@@ -98,13 +98,13 @@ this.rf_bandit_highwayman <- ::inherit("scripts/entity/tactical/human", {
...
 
-				if (::Math.rand(1, 100) <= ::Reforged.Config.ArmorAttachmentChance.Tier4)
+				if (::Math.rand(1, 100) <= ::Reforged.Config.ArmorAttachmentChance.Tier2)
 				{
 					local armorAttachment = ::Reforged.ItemTable.ArmorAttachmentNorthern.roll({
 						Apply = function ( _script, _weight )
 						{
 							local conditionModifier = ::ItemTables.ItemInfoByScript[_script].ConditionModifier;
-							if (conditionModifier > 40) return 0.0;
+							if (conditionModifier < 30 || conditionModifier > 30) return 0.0;
```
LordMidas, line 107: Maybe it was meant to be `> 40`? Because right now it is < 30 and > 30 is 0 which means he can only get exactly 30.

mod_reforged/hooks/entity/tactical/humans/swordmaster.nut
```diff
@@ -71,11 +71,18 @@
...
 			if (::Math.rand(1, 100) <= ::Reforged.Config.ArmorAttachmentChance.Tier3)
 			{
-				local armorAttachment = ::MSU.Class.WeightedContainer([
-					[1, "scripts/items/armor_upgrades/direwolf_pelt_upgrade"],
-					[1, "scripts/items/armor_upgrades/leather_shoulderguards_upgrade"],
-					[1, "scripts/items/armor_upgrades/double_mail_upgrade"]
-				]).roll();
+				if (this.m.MyArmorVariant == 0) // light armor
+				{
+					local armorAttachment = ::MSU.Class.WeightedContainer([
```
LordMidas, line 76: `armorAttachment` variable has to be declared above this `if` scope because it is used outside it further down.

LordMidas (changes requested):

scripts/entity/tactical/enemies/rf_bandit_highwayman.nut
Uberbagel: No that's correct. Those are the only appropriate attachments given the durability of their armor and me not wanting them to have access to the 40s

mod_reforged/hooks/entity/tactical/humans/swordmaster.nut
Uberbagel: Thank you

scripts/entity/tactical/enemies/rf_bandit_highwayman.nut
LordMidas: In that case I'd change it to `< 30 || >= 40` as that is a better representation of the level you want them to be at and is more compatible with mods adding their own items e.g. at 35. Otherwise that entire condition is just `conditionModifier != 30`.

PR #727, opened 2025-04-06
Uber text improvements

Uberbagel:


mod_reforged/hooks/config/strings.nut
```diff
@@ -1162,7 +1162,7 @@ foreach (vanillaDesc in vanillaDescriptions)
 	}),
 	RF_DentArmor = ::UPD.getDescription({
 		Fluff = "\'Can\'t fight if they can\'t walk.\'",
-		Requirement = "Blunt Damage",
+		Requirement = "Blunt Damage with Hammer",
```
LordMidas, line 1165: This is not true. The Dent Armor perk does not require a Hammer. This perk is not relevant for players as they get the Dented Armor effect applied differently (via Hammer Mastery). This perk is for enemies only.

LordMidas (changes requested):

LordMidas (approved):

PR #734, opened 2025-04-09
refactor: update conflicting mods

LordMidas:


Uberbagel (approved):

Darxo (approved):
We should think about managing the mod conflict logic into a separate file, where we have more space for structuring purposes.
In the preload script they will take up a lot of space as time goes on

PR #737, opened 2025-04-10
balance: spear mastery free attack and spearwall AP reduction

LordMidas:
Keeps the identity of the free attack on spears while making it a little more gated to use for balance. Primarily it prevents 2-tile spears e.g. Warfork, Spetum, Swordstaff from walking 2 tiles and still attacking twice.

Also remove Spearwall AP cost reduction.

### DIFF
--- a/mod_reforged/hooks/config/strings.nut
+++ b/mod_reforged/hooks/config/strings.nut
@@ -658,7 +658,7 @@ local vanillaDescriptions = [
 					"Skills build up " + ::MSU.Text.colorPositive("25%") + " less [Fatigue.|Concept.Fatigue]",
 					"The [Action Point|Concept.ActionPoints] cost of [Spearwall|Skill+spearwall] is reduced by " + ::MSU.Text.colorPositive("1") + ".",
 					"[Spearwall|Skill+spearwall] is no longer disabled once an opponent manages to overcome it. Instead, [Spearwall|Skill+spearwall] can still be used and continues to give free attacks on any further opponent attempting to enter the [Zone of Control|Concept.ZoneOfControl]",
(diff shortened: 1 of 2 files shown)


mod_reforged/hooks/config/strings.nut
```diff
@@ -656,9 +656,8 @@ local vanillaDescriptions = [
 				Type = ::UPD.EffectType.Passive,
 				Description = [
 					"Skills build up " + ::MSU.Text.colorPositive("25%") + " less [Fatigue.|Concept.Fatigue]",
-					"The [Action Point|Concept.ActionPoints] cost of [Spearwall|Skill+spearwall] is reduced by " + ::MSU.Text.colorPositive("1") + ".",
```
Darxo, line 659: This is not mentioned in the commit message. Was this put in by accident?
If not, please give me few words to understand this change. And ideally this change should be part of the commit message

mod_reforged/hooks/skills/perks/perk_mastery_spear.nut
```diff
@@ -82,6 +80,14 @@
 		}
 	}
 
+	q.onMovementStarted = @(__original) function( _tile, _numTiles )
+	{
+		this.m.TilesMoved += _numTiles;
```
Darxo, line 85: You never reset TilesMoved to 0

Darxo (changes requested):

mod_reforged/hooks/config/strings.nut
LordMidas: That got included in that by mistake. I've pulled it out into a separate commit.

Darxo (approved):

Uberbagel (approved):

Darxo (approved):

PR #738, opened 2025-04-10
balance: only get crowded by enemies with lower reach than you

LordMidas:
This accomplishes two things:
- From a realism perspective it makes sense that others with longer weapons like to remain at a longer distance from you, so they aren't really "crowding" you the same way someone with a shorter weapon would by closing the gap with you.
- From a gameplay perspective it encourages the use of longer weapons, even in melee, against beasts and enemies who out-reach you. This helps the player have a tool to overcome the Reach Disadvantage, by using longer weapons, and also sounds great thematically when fighting bigger beasts.


Uberbagel (approved):
Just to be sure, this does not take into account reach negation from shields, correct?

LordMidas:
> Just to be sure, this does not take into account reach negation from shields, correct?

Yes, it only compares the current Reach value of both actors.

PR #749, opened 2025-04-18
Helper functions to seed random

LordMidas:


mod_reforged/config/math/math.nut
```diff
@@ -31,13 +31,81 @@
 		return (newDistance < oldDistance) ? _newValue : _oldValue;
 	}
 
-	function seededRand( _seed, _min, _max )
+	function seededRand( _min, _max, ... )
```
Darxo, line 34: An alternative name for this function could be `RandWithFixedSeed`
As of right now it sounds very similar to seedRandom

Darxo (approved):

PR #751, opened 2025-04-19
refactor: restrict overlay icon for bleed effect to one per skill usage

Darxo:
Right now a bleed overlay icon is spawned on top of the receiver for every stack of bleed they receive. This is most notable with `gouge` which adds 3 stacks by default and 5 stacks if you have cleaver mastery and mauler.

This PR uses the vanila `SkillCounter` to keep track of whether we already posted the overlay icon


LordMidas (changes requested):
I think the icon spam is relevant for the player to see at a glance how many bleeds were applied in a single attack. I tried an alternative approach where we speed up icons from subsequent bleeds in a single attack. Please see #753

Darxo:
Ok, lets go with your fix. I will close this PR

PR #760, opened 2025-04-27
refactor: show neutral towns in relation screen with recent changes

Darxo:
Vanilla decided to not display all factions at all times in your relation screen. Probably because that would be too much scrolling.
Civilian factions (every non-military settlement comes with its own one) are hidden, if you have exactly 50 relation with them (which is the default/neutral)

In some situations however this can be confusing:
- If you completed a contract with a settlement that you previously had 40 relation with, it will now bow at 50 relation and no longer show up in the relation screen. When to try to see how that contract impacted your relation you wont be able to see it
- Same it true for when one of your actions reduced the relation of a settlement down to exactly 50

With this change neutral village factions now always show in the relation ship screen if their relation ship changed with you in the last 10 days (not counting decay).
Or in other words: If they have a relationship change entry with a flavor description, then they will not be hidden.

Like for example in this screenshot (using the fix)
image

In the same playthrough I also have other villages with 50 relationship but which dont appear in the relationship screen. That does check out because I probably havent done anything for them in the last 10 days.

### DIFF
--- a/mod_reforged/hooks/factions/faction.nut
+++ b/mod_reforged/hooks/factions/faction.nut
@@ -1,0 +1,7 @@
+::Reforged.HooksMod.hook("scripts/factions/faction", function (q) {
+	// Overwrite, because we only want to hide certain settlements at 50 relation, if they didn't have a recent relationship change with the player
+	q.isHidden = @() function()
+	{
+		return this.m.IsHidden || (this.m.IsHiddenIfNeutral && this.m.PlayerRelation == 50 && this.m.PlayerRelationChanges.len() == 0);
+	}
+});


mod_reforged/hooks/factions/faction.nut
```diff
@@ -0,0 +1,7 @@
+::Reforged.HooksMod.hook("scripts/factions/faction", function (q) {
+	// Overwrite, because we only want to hide certain settlements at 50 relation, if they didn't have a recent relationship change with the player
+	q.isHidden = @() function()
+	{
+		return this.m.IsHidden || (this.m.IsHiddenIfNeutral && this.m.PlayerRelation == 50 && this.m.PlayerRelationChanges.len() == 0);
```
LordMidas, line 5: Looks good to me but I have a question. Vanilla here does a `Math.abs(this.m.PlayerRelation - 50.0) == 0.0` comparison whereas you directly check for `this.m.PlayerRelation == 50`. Do we know if there is a good reason vanilla did that `abs` thing and if there are any pitfalls with this new method? Btw the return type of `::Math.abs` is integer.

Darxo: I don't understand vanillas way of doing the check. 
Maybe this was a leftover from a refactoring from an earlier version.

LordMidas (approved):

mod_reforged/hooks/factions/faction.nut
LordMidas: Ok let's go with your method. If anything strange happens we will see.

Uberbagel (approved):

PR #761, opened 2025-04-27
refactor: mention zombie ability to resurrect slain humans as zombies

Darxo:
Zombie have an undocumented ability, build in their `zombie.nut` script (therefor affecting everything inheriting this), which causes 
- any humans and any player characters slain (last hitted) by them to resurrect in 2 - 3 turns
- this only affects those that actually can be resurrected
  - struck down players wont resurrect
  - headless characters will also not resurrect this way
  
This PR will mention this effect in the zombie_racial. Although it might happen that at some point an enemy receives the `zombie_racial`, but isnt inheriting from `zombie.nut`. Then we have to revisit this built-in effect

Here the tooltip how it looks now (after first adjustment)
image

### DIFF
--- a/scripts/skills/racial/rf_zombie_racial.nut
+++ b/scripts/skills/racial/rf_zombie_racial.nut
@@ -47,6 +47,13 @@ this.rf_zombie_racial <- ::inherit("scripts/skills/skill", {
 			});
 		}
 
+		ret.push({
+			id = 12,
+			type = "text",
+			icon = "ui/orientation/zombie_01_orientation.png",
+			text = "Humans killed by you will rise as Wiedergangers after some time",	// This is implemented in onActorKilled of zombie.nut
+		});
+
 		ret.extend([
 			{
 				id = 20,
--- a/scripts/skills/racial/rf_zombie_racial.nut
+++ b/scripts/skills/racial/rf_zombie_racial.nut
@@ -23,7 +23,7 @@ this.rf_zombie_racial <- ::inherit("scripts/skills/skill", {
 				id = 10,
 				type = "text",
 				icon = "ui/icons/fatigue.png",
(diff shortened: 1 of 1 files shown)


scripts/skills/racial/rf_zombie_racial.nut
```diff
@@ -47,6 +47,14 @@ this.rf_zombie_racial <- ::inherit("scripts/skills/skill", {
 			});
 		}
 
+		ret.push({
+			id = 12,
+			type = "text",
+			icon = "ui/icons/obituary.png",	// skull icon
+			text = "Resurrect Humans that you kill as a zombie, if possible",	// This is implemented in onActorKilled of zombie.nut
```
LordMidas, line 54: I'd reword this as: `Humans killed by you will rise as Wiedergangers after some time`. The `if possible` part is unnecessary because headless wiedergangers also don't resurrect and struck down characters are not killed.

LordMidas (changes requested):

LordMidas (approved):

scripts/skills/racial/rf_zombie_racial.nut
```diff
@@ -47,6 +47,13 @@ this.rf_zombie_racial <- ::inherit("scripts/skills/skill", {
 			});
 		}
 
+		ret.push({
+			id = 12,
+			type = "text",
+			icon = "ui/icons/obituary.png",	// skull icon
```
LordMidas, line 53: Ok looking at the screenshot now, perhaps the zombie orientation icon would be better?

Darxo: image

LordMidas (approved):

Uberbagel (approved):

PR #762, opened 2025-04-28
fix: unstoppable not resetting to 0 upon getting stunned or rooted

LordMidas:
Because the IsStunned and IsRooted properties are set during `onUpdate` of the effects that cause them, SkillOrder becomes important if we are checking for those properties in `onUpdate`. The stunned_effect may be present but if it updates after this perk then we won't detect the character being stunned. We can't check for `CurrentProperties.IsStunned` either because CurrentProperties are only updated at the end of the entire update cycle and if the stun comes from an effect that updates after this perk, we will still miss it.

However, this will now cause the things done by this perk to happen a bit too late for other skills which may depend on those stats (e.g. increased Initiative). But that seems correct because otherwise that also depends on SkillOrder which seems wrong from a design pov.

### DIFF
--- a/scripts/skills/perks/perk_rf_unstoppable.nut
+++ b/scripts/skills/perks/perk_rf_unstoppable.nut
@@ -104,7 +104,7 @@ this.perk_rf_unstoppable <- ::inherit("scripts/skills/skill", {
 		}
 	}
 
-	function onUpdate( _properties )
+	function onAfterUpdate( _properties )
 	{
 		if (_properties.IsStunned || _properties.IsRooted || this.getContainer().getActor().getMoraleState() == ::Const.MoraleState.Fleeing || this.getContainer().hasSkill("effects.staggered"))
 		{


Darxo (commented):
How about we implement the regular stat boost during `onUpdate`
But during `onAfterUpdate` we check for `IsStunned`
And when we detect that property being true AND `this.m.Stacks != 0`, then we set `this.m.Stacks = 0` and `setDirty(true)` on this actor.
That will force a second update in which we no longer add stats

LordMidas:
> `setDirty(true)` on this actor. That will force a second update in which we no longer add stats

`setDirty(true)` during a skill_container update does not trigger another update after it.

Darxo:
Then perhaps we trigger a `this.getContainer().update();`

LordMidas:
> Then perhaps we trigger a `this.getContainer().update();`

Doing that while already inside an update process does nothing because `container.m.IsUpdating` is true. And we can't manually set `container.m.IsUpdating` to false and then do an update as that can break things.

We have an event in Reforged called `onSkillsUpdated` which triggers after update and can be used to trigger new updates - but that has strong potential for infinite recursions so I actually want to delete that feature at some point.

Uberbagel (approved):

Darxo (approved):

PR #763, opened 2025-04-28
feat: include effect of traits and perm injuries in projected attributes

LordMidas:
Tested and seems to work nicely.

### DIFF
--- a/mod_reforged/hooks/entity/tactical/player.nut
+++ b/mod_reforged/hooks/entity/tactical/player.nut
@@ -48,6 +48,13 @@
 		}
 		this.getSkills().m.IsUpdating = wasUpdating;
 
+		local levelUpsRemaining = ::Math.max(::Const.XP.MaxLevelWithPerkpoints - this.getLevel() + this.getLevelUps(), 0);
+
+		// Fatigue and Hitpoints getter functions are in actor and they use CurrentProperties
+		// so we temporarily switcheroo the CurrentProperties to get our desired values
+		local original_CurrentProperties = this.m.CurrentProperties;
+		this.m.CurrentProperties = properties;
+
 		local ret = {};
 		foreach (attributeName, attribute in ::Const.Attributes)
 		{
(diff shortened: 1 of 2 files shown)


mod_reforged/hooks/entity/tactical/player.nut
```diff
@@ -37,7 +37,15 @@
 
 	q.getProjectedAttributes <- function()
 	{
-		local baseProperties = this.getBaseProperties();
+		local properties = this.getBaseProperties().getClone();
+
+		local wasUpdating = this.getSkills().m.IsUpdating;
+		this.getSkills().m.IsUpdating = true;
+		foreach (s in this.getSkills().getSkillsByFunction(@( _skill ) _skill.isType(::Const.SkillType.Trait) || _skill.isType(::Const.SkillType.PermanentInjury)))
```
Darxo, line 44: There should be a comment somewhere in this area or above the function definition, explaining what we are doing.
Mainly, that the returned attribute array will include changes from permanent injuries and traits

Darxo (approved):

Uberbagel:
I'm not sure I agree with this change because vanilla doesn't take them into account on the levelup attributes screen for example. This seems like we're creating an inconsistency.

LordMidas:
> I'm not sure I agree with this change because vanilla doesn't take them into account on the levelup attributes screen for example. This seems like we're creating an inconsistency.

We show the Base Attributes in the tooltip of the attribute itself in Reforged when the user hovers over that. So that is where you get that info. The main reason people look at these projected attributes is to compare bros or to assess the potential skill of a bro. And for that it is essential in my opinion to include this information. Granted, if you wanted to really know the Base Attribute at level 11 you won't be able to see that now anywhere. But that seems like a far less likely use case (e.g. if you want to learn a perk which depends on base attribute directly and you want to see where that will be at level 11). If this information is really necessary, we can consider adding that in the tooltip of the attributes along with the current base that we show there.

LordMidas:
> I'm still wary but also not that concerned about it overall and you both like it so fine by me.

In case it wasn't clear, this PR also adjusts the tooltip's wording to make it clear that the stats shown have the effects of traits and permanent injuries included in them.

Uberbagel:
> > I'm still wary but also not that concerned about it overall and you both like it so fine by me.
> 
> In case it wasn't clear, this PR also adjusts the tooltip's wording to make it clear that the stats shown have the effects of traits and permanent injuries included in them.

That makes me way more comfortable with it.

Darxo (approved):

PR #765, opened 2025-04-29
feat: prevent ai from swapping to ranged when near melee opponents

LordMidas:
Because it is a bad idea to lose your Reach when melee opponents can move to you and attack you.

I tested this in game and it works nicely. Enemies that ended their turn in a spot where I could reach them in 1 turn didn't swap to throwing weapons, but enemies which stayed behind them swapped properly.

### DIFF
--- a/mod_reforged/hooks/ai/tactical/behaviors/ai_switchto_ranged.nut
+++ b/mod_reforged/hooks/ai/tactical/behaviors/ai_switchto_ranged.nut
@@ -6,7 +6,6 @@
 		this.m.WeaponToEquip = null;
 
 		local myTile = _entity.getTile();
-		local switchMaxDangerDist = ::Math.max(_entity.getIdealRange(), ::Const.AI.Behavior.SwitchToRangedMaxDangerDist);
 
 		foreach (enemy in this.getAgent().getKnownOpponents())
 		{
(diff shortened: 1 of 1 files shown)


mod_reforged/hooks/ai/tactical/behaviors/ai_switchto_ranged.nut
```diff
@@ -0,0 +1,36 @@
...
+				continue;
+			}
+
+			if (enemy.Tile.getZoneOfControlCountOtherThan(enemy.Actor.getAlliedFactions()) >= ::Const.AI.Behavior.SwitchToRangedIgnoreDangerMinZones)
+			{
+				continue;
+			}
+
+			// If an enemy can reach us in 1 turn, don't swap to a ranged weapon
+			if (this.queryActorTurnsNearTarget(enemy.Actor, myTile, _entity).Turns <= 1)
```
Darxo, line 28: Does this include their ability to actually attack in the same turn?

Various enemie with throwing weapons also have Quickhands, which lets them instantly fix the bad position they found themselves in by switching to their bag melee weapon and attacking.

LordMidas: No. There's a separate thing for that in that query result called `.TurnWithAttack`. However, the thing is that even if the enemy can't attack me this turn, maybe next turn he can acts before me and I won't be able to swap to a melee before that. So I find that it's probably better to just check if they can reach me in 1 turn.

Darxo (approved):

Uberbagel (approved):

PR #767, opened 2025-04-29
fix: spear mastery not accounting for teleporation in tiles moved

LordMidas:


mod_reforged/hooks/skills/perks/perk_mastery_spear.nut
```diff
@@ -113,6 +131,7 @@
 		__original();
 		this.m.IsSpent = true;
 		this.m.TilesMoved = 0;
+		this.m.PrevTile = null;
```
Darxo, line 134: Is this really needed?
I dont know a single case where `onMovementStarted` is not then followed by a `onMovementFinished`
So `PrevTile` will always be set back to `null`

Darxo (changes requested):

mod_reforged/hooks/skills/perks/perk_mastery_spear.nut
LordMidas: Yeah probably unnecessary. I just added it as a bullet proof check but I can remove it.

Darxo (approved):

PR #769, opened 2025-05-04
refactor: add seed and company name as additional salt for seedRandom

Darxo:
These are two additional values unique to a specific campaign

### DIFF
--- a/mod_reforged/config/math/math.nut
+++ b/mod_reforged/config/math/math.nut
@@ -57,6 +57,8 @@
 		if ("Assets" in ::World && !::MSU.isNull(::World.Assets))
 		{
 			vargv.push(::World.Assets.getCampaignID());
+			vargv.push(::World.Assets.getSeedString());
+			vargv.push(::World.Assets.getName());
 		}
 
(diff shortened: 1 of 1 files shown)


LordMidas:
I am not sure about this. People can run multiple campaigns with the same Company Name and Seed. Isn't the CampaignID that we use already enough to guarantee different randomization in every campaign?

Darxo:
I dont trust that the ID is enough by itself.
Maybe IDs are reused, or incrementing it by 1 still produces predictable differences (despite multiplying it by a big number)
And I see no hard in adding more values, that are guaranteed to stay the same within the same campaign.

LordMidas:
My point was that neither the SeedString nor the Name is guaranteed to be different across different campaigns. Players _can_ have multiple campaigns with the same SeedString and same company Name. However, one can say that most probably players will run different campaigns with different seed and company name. So, I agree this sounds like an improvement for the overwhelming majority of cases to keep things different across different campaigns.

LordMidas (approved):

PR #770, opened 2025-05-04
feat: use fixed seed for location terrain

Darxo:
### Upside
- Prevents cheese of reloading autosaves in hopes for better terrain
- Prevents fleeing from battle when terrain is bad (mostly against hill locations)
- Introduces more realism/persistance into the world. Terrain doesnt magically change from one moment to the next
- Introduces option to scout location before sending everyone inside

### Potential Downside
- The option to scout locations might become required and unfun busywork if you want to play optimal
  - Can screenshot starting area to then 
    - position full squad perfectly in the real fight
    - visit the location much later into the campaign
- In some rare situation a location might come with an unbeatably hard terrain, bricking it effectively

All-in-All I'm all for trying this out.
Maybe we can discourage the need for scouting a location with other means down the line

### DIFF
--- a/mod_reforged/hooks/states/tactical_state.nut
+++ b/mod_reforged/hooks/states/tactical_state.nut
@@ -1,4 +1,28 @@
 ::Reforged.HooksMod.hook("scripts/states/tactical_state", function(q) {
+	q.initMap = @(__original) function()
+	{
+		// Feat: the same location will always produce the exact same layout
+		// This has no influence over troop spawning as those will be seeded by "combatSeed" later on
+		// Our seeding always happens first. If vanilla then chooses to use their own fixed seed, they will replace our attempt
+		foreach (party in this.m.StrategicProperties.Parties)
+		{
+			if (party.isLocation())
+			{
+				::Reforged.Math.seedRandom(
+					"RF_FixedLocationSeed",		// Fixed salt, specific to use-case
+					party.getTypeID(),			// Location specific salt
+					party.getName(),			// Location specific salt
+					party.getFaction(),			// Faction specific salt
+					party.getTile().X * 200,	// Position specific salt
+					party.getTile().Y			// Position specific salt
+				);
+				break;
+			}
+		}
+
+		__original();
+	}
+
 	q.showRetreatScreen = @(__original) function ( _tag = null )
 	{
 		this.m.TacticalScreen.getTopbarOptionsModule().changeFleeButtonToWin();


mod_reforged/hooks/states/tactical_state.nut
```diff
@@ -1,4 +1,29 @@
...
+		// This has no influence over troop spawning as those will be seeded  by "combatSeed" later on
+		if (this.m.StrategicProperties.MapSeed == 0)	// Other fixed seeds take priority
+		{
+			foreach (party in this.m.StrategicProperties.Parties)
+			{
+				if (party.isLocation())
+				{
+					::Reforged.Math.seedRandom(
+						"RF_FixedLocationSeed",		// Fixed salt, specific to use-case
+						party.getID(),				// Location specific salt
```
LordMidas, line 14: Are we sure that `party.getID()` is the same on save/load? I know that tactical entities `getID()` is changed on save/load.

Darxo: Right, it was meant to be getTypeID instead. its fixed now

Uberbagel (approved):
Nice idea. While it might bring about some frustration, I like the idea that some locations might just be a harder nut to crack than others. Players then have to actually make the choice of abandoning the contract, taking the risk, utilizing more saved resources (potions, bombs etc...) and yes, even retreating and coming back with their troops better aligned for the attack. This last part I don't even think is a bad thing, it might slow the game down a tad but allows for strategic play by the player.

mod_reforged/hooks/states/tactical_state.nut
```diff
@@ -1,4 +1,29 @@
...
+				if (party.isLocation())
+				{
+					::Reforged.Math.seedRandom(
+						"RF_FixedLocationSeed",		// Fixed salt, specific to use-case
+						party.getTypeID(),			// Location specific salt
+						party.getName(),			// Location specific salt
+						party.getFaction(),			// Faction specific salt
+						party.getTile().X * 200,	// Position specific salt
+						party.getTile().Y			// Position specific salt
+					);
```
LordMidas, line 19: Probably it should have a `break;` at the end of this `if`, no?

mod_reforged/hooks/states/tactical_state.nut
```diff
@@ -1,4 +1,30 @@
...
+				}
+				break;
```
LordMidas, line 21: The `break;` was added in the wrong place, it needs to be inside the `if` block.

mod_reforged/hooks/states/tactical_state.nut
```diff
@@ -1,4 +1,30 @@
 ::Reforged.HooksMod.hook("scripts/states/tactical_state", function(q) {
+	q.initMap = @(__original) function()
+	{
+		// Feat: the same location will always produce the exact same layout
+		// This has no influence over troop spawning as those will be seeded  by "combatSeed" later on
+		if (this.m.StrategicProperties.MapSeed == 0)	// Other fixed seeds take priority
```
LordMidas, line 6: In the original function I have seen that it also checks for `LocationTemplate != null` in addition to `MapSeed != 0` before it does the `seedRandom`. Do we not need that condition as well here?

LordMidas (changes requested):

mod_reforged/hooks/states/tactical_state.nut
Darxo: I dont remember why I only went with one check.
Upon further looking at the vanilla code I now chose to remove all condition checks, as our code runs before the vanilla code anyways. They will overwrite our seeding attempt.
And the performance impact of not checking the condition first is negligible, as this runs only once per combat initialization

PR #780, opened 2025-05-10
balance: change decisive to multiplicative resolve and initiative bonus

LordMidas:


mod_reforged/hooks/config/strings.nut
```diff
@@ -1132,7 +1132,7 @@ foreach (vanillaDesc in vanillaDescriptions)
 			Type = ::UPD.EffectType.Passive,
 			Description = [
 				"Gain a stack whenever you end your [turn|Concept.Turn] without having used [Wait,|Concept.Wait] up to a maximum of 3 stacks.",
-				"Gain " + ::MSU.Text.colorPositive("+15") + " [Resolve|Concept.Bravery] and [Initiative|Concept.Initiative] while you have at least 1 stack.",
+				"[Resolve|Concept.Bravery] and [Initiative|Concept.Initiative] are both increased by " + ::MSU.Text.colorPositive("15%") + " while you have at least 1 stack.",
```
Darxo, line 1135:
```suggestion
				"Gain " + ::MSU.Text.colorPositive("15%") + " more [Resolve|Concept.Bravery] and [Initiative|Concept.Initiative] while you have at least 1 stack.",
```
I would write it like this, utilizing our "more" term for multiplicative changes

Darxo (changes requested):

mod_reforged/hooks/config/strings.nut
LordMidas: Gain x% more sounds weird to me though. It has a feeling of "gain 15% more than what you were gaining before". The language I put in the text is more natural.

If we want to force in the use of "more" here perhaps the more appropriate sentence structure would be "You have x% more Resolve and Initiative while you have at least 1 stack."

LordMidas: In fact up on looking through the codebase we use this "increased by" quite often in perk tooltips where the sentence structure flows better with it. In status effects tooltips I agree that we should stick to "more".

Darxo: `You Have` seems better

LordMidas: After more deliberation I think the current one with "increased by" flows much better linguistically in the perk description so I will stay with that.

PR #782, opened 2025-05-10
feat: rework battle fervor to gain stack at turn start with morale check

LordMidas:
As long as you start each turn at Confident morale. Each stack increases attributes by 5%. Can have a maximum of 4 stacks. Expires upon losing Confident morale.

### DIFF
--- a/mod_reforged/hooks/config/strings.nut
+++ b/mod_reforged/hooks/config/strings.nut
@@ -900,6 +900,7 @@ foreach (vanillaDesc in vanillaDescriptions)
 			Type = ::UPD.EffectType.Passive,
 			Description = [
 				"Receive a [morale check|Concept.Morale] at the start of every [turn|Concept.Turn] if you are [Confident.|Concept.Morale] and if successful, gain a stack. For each stack, [Resolve,|Concept.Bravery] [Melee Skill,|Concept.MeleeSkill] [Ranged Skill,|Concept.RangeSkill] [Melee Defense|Concept.MeleeDefense] and [Ranged Defense|Concept.RangeDefense] are increased by " + ::MSU.Text.colorPositive("5%") + ".",
+				"Lose " + ::MSU.Text.colorNegative(1) + " stack upon ending a [turn|Concept.Turn] without having attacked.",
 				"Cannot have more than 4 stacks."
 			]
 		}]
--- a/mod_reforged/hooks/config/strings.nut
+++ b/mod_reforged/hooks/config/strings.nut
@@ -899,7 +899,7 @@ foreach (vanillaDesc in vanillaDescriptions)
 		Effects = [{
 			Type = ::UPD.EffectType.Passive,
 			Description = [
-				"Receive a [morale check|Concept.Morale] at the start of every [turn|Concept.Turn] if you are [Confident.|Concept.Morale] and if successful, gain a stack. For each stack, [Resolve,|Concept.Bravery] [Melee Skill,|Concept.MeleeSkill] [Ranged Skill,|Concept.RangeSkill] [Melee Defense|Concept.MeleeDefense] and [Ranged Defense|Concept.RangeDefense] are increased by " + ::MSU.Text.colorPositive("5%") + ".",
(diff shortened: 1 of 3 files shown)


Uberbagel (commented):
I think this needs a limiter of being 'engaged' with the enemy.
Engaged being a new concept that means you are adjacent to an enemy, or have made a melee attack against an enemy or taken a melee attack from an enemy. For range it should be you've made or taken a ranged attack at a distance of 4 tiles or _fewer._
Otherwise it will be too easy to start gaining stacks before the lines have clashed and getting up to 4 stacks. This also would incentivize running until those 4 stacks are gotten.

LordMidas:
That's a good point (similar to what we did with Unstoppable). I think the limiter should be: lose one stack in a turn where you do not perform an attack.

Uberbagel:
> That's a good point (similar to what we did with Unstoppable). I think the limiter should be: lose one stack in a turn where you do not perform an attack.

That's not a real limiter on ranged attackers though. Once they're confident it's just free damage.

LordMidas:
> That's not a real limiter on ranged attackers though. Once they're confident it's just free damage.

I think it's not that bad. It is similar to the condition on Unstoppable, and Unstoppable gains stacks guaranteed whereas Battle Fervor gains stacks only on successful morale check.

Also this perk gives attributes bonus, not bonus damage. Not sure if that was a typo on your end or if there was some misunderstanding.

We can also cause it to lose half the stacks if you didn't perform an attack. But I think losing 1 stack would probably be enough, as this perk ramps up rather slowly.

Uberbagel:
Ah I think I conflated this bonus with the proposed damage bonus on exploit opening. I think this is okay then because the player shouldn't be able to infinitely flee while also attacking each turn with ranged guys.

mod_reforged/hooks/config/strings.nut
```diff
@@ -899,9 +899,9 @@ foreach (vanillaDesc in vanillaDescriptions)
 		Effects = [{
 			Type = ::UPD.EffectType.Passive,
 			Description = [
-				"[Resolve|Concept.Bravery] is increased by " + ::MSU.Text.colorPositive("10%") + " at all times.",
-				"Additionally, at positive morale checks, [Resolve|Concept.Bravery] is increased by a further " + ::MSU.Text.colorPositive("+10") + ".",
-				"When at Confident Morale, all the bonuses of this perk are doubled, and [Melee Skill,|Concept.MeleeSkill] [Ranged Skill,|Concept.RangeSkill] [Melee Defense,|Concept.MeleeDefense] and [Ranged Defense|Concept.RangeDefense] are increased by " + ::MSU.Text.colorPositive("5%") + "."
+				"Receive a [morale check|Concept.Morale] at the start of every [turn|Concept.Turn] if you are [Confident.|Concept.Morale] and if successful, gain a stack. For each stack, [Resolve,|Concept.Bravery] [Melee Skill,|Concept.MeleeSkill] [Ranged Skill,|Concept.RangeSkill] [Melee Defense|Concept.MeleeDefense] and [Ranged Defense|Concept.RangeDefense] are increased by " + ::MSU.Text.colorPositive("5%") + ".",
```
Uberbagel, line 902: Sentence weirdness. '...start of every turn if you are Confident. and if successful...' Needs to be either a period or an and.

Uberbagel (approved):

mod_reforged/hooks/config/strings.nut
LordMidas: Good catch, that was a typo. Will remove the period.

LordMidas:
> because the player shouldn't be able to infinitely flee while also attacking each turn with ranged guys

I couldn't understand what you mean by "infinitely flee" :(

Uberbagel:
> > because the player shouldn't be able to infinitely flee while also attacking each turn with ranged guys
> 
> I couldn't understand what you mean by "infinitely flee" :(

If the limiter on the perk wasn't strong enough, the player could just run away from the enemy as their stacks build up.

scripts/skills/perks/perk_rf_battle_fervor.nut
```diff
@@ -1,9 +1,10 @@
 this.perk_rf_battle_fervor <- ::inherit("scripts/skills/skill", {
 	m = {
 		Stacks = 0,
+		HasAttackedThisTurn = false,
 
 		MaxStacks = 4,
-		MultPerStack = 0.05
```
Darxo: This renaming must also happen down below in the `RF_getMult` function

Darxo (changes requested):

scripts/skills/perks/perk_rf_battle_fervor.nut
LordMidas: It's already done, you might be looking at an older version of the code.

Darxo (approved):

Darxo (approved):

PR #801, opened 2025-05-19
Fix several issues with onAnySkillExecutedFully

LordMidas:


mod_reforged/hooks/experimental_modules/onAnySkillExecutedFully.nut
```diff
@@ -39,14 +39,43 @@
 	}
 };
 
+// Looks through the stackinfos and returns the skill which called a scheduling function e.g. ::Time.scheduleEvent.
+// This only includes skills which are present in ::Reforged.ScheduleSkills and that means only skills that are being used.
+// Returns null if the function was called by anything else.
+local function getSchedulerSkill()
+{
+	local level = 3;
```
Darxo, line 47: There should be some explanation of where this magic number comes from.
Something like 
`// 0 = "getstackinfos"; 1 = "getSchedulerSkill"; 2 = scheduler function wanting to know its related skill`

Darxo (approved):

PR #805, opened 2025-05-20
feat: allow AI to evaluate but not execute during delayed events

LordMidas:


LordMidas:
Closed in favor of #806

PR #806, opened 2025-05-20
feat: allow the AI to evaluate in parallel but rethink if state changed

LordMidas:
This should improve AI performance because we allow the AI to evaluate (but not execute) its behaviors while delayed events are scheduled from a used skill. However, if anything important changed in the state during the delayed events e.g. anyone died, or moved, or morale state changed, or Action Points changed, then we force the AI to reevaluate using this new state. This is because any important state change could make the already evaluated behaviors to be no longer valid for execution.

### DIFF
--- a/mod_reforged/hooks/experimental_modules/onAnySkillExecutedFully.nut
+++ b/mod_reforged/hooks/experimental_modules/onAnySkillExecutedFully.nut
@@ -31,9 +31,18 @@
 
 	function onScheduleComplete()
 	{
+		if (this.WasScheduled)
+			::Reforged.Mod.Debug.printLog(format("onScheduleComplete -- ID: %s, Count: %i", this.Skill.getID(), this.Count), "onAnySkillExecutedFully");
+		else
+			::Reforged.Mod.Debug.printLog(format("onScheduleComplete -- ID: %s, Count: %i -- Nothing was scheduled (normal skill)", this.Skill.getID(), this.Count), "onAnySkillExecutedFully");
 		// Check for <= 0 because when we call this manually, this.Count will be 0 and will drop to -1
 		if (--this.Count <= 0)
 		{
+			if (this.WasScheduled)
+				::Reforged.Mod.Debug.printLog(format("onScheduleComplete -- ID: %s, Count: %i - Schedule Complete", this.Skill.getID(), this.Count), "onAnySkillExecutedFully");
+			else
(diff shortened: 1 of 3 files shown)


mod_reforged/hooks/experimental_modules/onAnySkillExecutedFully.nut
```diff
@@ -16,15 +16,27 @@
...
 	ForFree = false;
+	Agent = null;
+
 	Count = 0;
 
+	// State of the actor as it was when the first delayed event was scheduled during a skill use
+	// Is used to tell the agent to restart its evaluation if something important changed on the battlefield
+	// which may make the results of previously evaluated behaviors potentially invalid.
+	APBefore = 0;
+	MoraleStateBefore = ::Const.MoraleState.Steady;
```
Suor, line 27: Don't care about Morale state, only care about fleeing or not.

mod_reforged/hooks/experimental_modules/onAnySkillExecutedFully.nut
```diff
@@ -69,6 +100,40 @@ local function getSchedulerSkill()
...
+// is wrapped to trigger the onScheduleComplete() function for the schedule.
+// Also increments the schedule Count of the relevant skill and saves the actor state when at Count 0.
+// Note: _countBump is necessary as a param because ::Tactical.switchEntities needs to increase count by 2 on every call
+// as its callback is triggered twice: once for each entity being switched.
+local function getWrapper( _caller, _func, _numArgs, _countBump = 1 )
+{
+	local schedule = ::Reforged.ScheduleSkills[_caller];
+	if (schedule.Count == 0)
+	{
+		schedule.saveStartingState();
```
Suor, line 113: A really weird place to do it.

mod_reforged/hooks/experimental_modules/onAnySkillExecutedFully.nut
```diff
@@ -34,9 +46,28 @@
 		{
 			if (!::MSU.isNull(this.Container))
 				this.Container.onAnySkillExecutedFully(this.Skill, this.TargetTile, this.TargetEntity, this.ForFree);
+			// this.Count == 0 check because if count is -1 that means no delayed event was scheduled.
+			if (this.Count == 0 && this.didStateChange())
+				this.Agent.m.RF_ForceReevaluate = true;
```
Suor, line 51: Why do we need this flag? May just do our reset here.

mod_reforged/hooks/experimental_modules/onAnySkillExecutedFully.nut
LordMidas: I have updated the PR by moving state save/check into an agent independent of scheduled events.

PR #819, opened 2025-06-17
feat: rework between the eyes and apply effect instead of proxy skill

LordMidas:
- Increase AP cost from 1 to 2.
- Now increases the headshot damage for the attack by 1.5x.

### DIFF
--- a/scripts/skills/effects/rf_between_the_eyes_effect.nut
+++ b/scripts/skills/effects/rf_between_the_eyes_effect.nut
@@ -1,0 +1,72 @@
+this.rf_between_the_eyes_effect <- ::inherit("scripts/skills/skill", {
+	m = {
+		MeleeSkillToHeadshotChancePct = 0.5
+		HeadshotDamageMult = 1.5
+	},
+	function create()
+	{
+		this.m.ID = "effects.rf_between_the_eyes";
+		this.m.Name = "Between the Eyes";
+		this.m.Description = "This character is prepared to land his next attack right between the target\'s eyes.";
+		this.m.Icon = "ui/perks//perk_rf_between_the_eyes.png";
+		this.m.Overlay = "rf_between_the_eyes_effect";
+		this.m.Type = ::Const.SkillType.StatusEffect;
+		this.m.IsRemovedAfterBattle = true;
+	}
+
+	function getTooltip()
+	{
+		local ret = this.skill.getTooltip();
+		local chanceStr;
+		if (::MSU.isEqual(this.getContainer().getActor(), ::MSU.getDummyPlayer()))
+		{
+			chanceStr = "chance equal to " + ::MSU.Text.colorizePct(this.m.MeleeSkillToHeadshotChancePct) + " of [Melee Skill|Concept.MeleeSkill]"
+		}
+		else
+		{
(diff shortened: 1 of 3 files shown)


scripts/skills/effects/rf_between_the_eyes_effect.nut
```diff
@@ -0,0 +1,72 @@
...
+		}
+		else
+		{
+			chanceStr = ::MSU.Text.colorizeValue(this.getHeadshotChanceAdd(), {AddSign = true, AddPercent = true}) + " chance"
+		}
+		ret.push({
+			id = 10,
+			type = "text",
+			icon = "ui/icons/chance_to_hit_head.png",
+			text = ::Reforged.Mod.Tooltips.parseString("The next melee attack has a " + chanceStr + " to hit the head and deals " + ::MSU.Text.colorizeMult(this.m.HeadshotDamageMult) + " damage on a such a hit")
```
Darxo, line 33: What kind of damage bonus you actually want here?

Right now this effect grants "50% more critical damage on a hit to the head".
That means it 
- only affects hitpoint damage
- is negated by steelbrow
- happens after armor mitigation

From this tooltip one would think that you just deal 50% more damage on a headshot (like a stronger killing frenzy)

scripts/skills/effects/rf_between_the_eyes_effect.nut
```diff
@@ -0,0 +1,72 @@
...
+
+		return ret;
+	}
+
+	function onAnySkillUsed( _skill, _targetEntity, _properties )
+	{
+		if (this.isSkillValid(_skill))
+		{
+			_properties.HitChance[::Const.BodyPart.Head] += this.getHeadshotChanceAdd();
+			_properties.DamageAgainstMult[::Const.BodyPart.Head] *= this.getHeadshotDamageMult();
```
Darxo, line 44: The existing vanilla effects, which utilize `DamageAgainstMult` only ever change it additively.
See `brute_trait.nut` or `chop.nut`

If you break that convention only at this place then there should be a comment here explaining why

Darxo (commented):
I would rework the convention around the vanilla `DamageAgainstMult` and make all changes to it always be multiplicative. Therefor the existing vanilla uses of it should be adjusted by reforged

scripts/skills/effects/rf_between_the_eyes_effect.nut
LordMidas: I don't really know why vanilla went the additive route. But perhaps it was because they wanted each of these traits to apply a fixed effect instead of multiplying each other's effects. I'll add a comment that we break the vanilla convention and apply multiplicative here.

Uberbagel (approved):

scripts/skills/effects/rf_between_the_eyes_effect.nut
Darxo: I do like the term **critical damage**

Here is a possible concept for it:
```
CriticalDamage = ::MSU.Class.BasicTooltip("Critical Damage", ::Reforged.Mod.Tooltips.parseString(
	"Critical Damage refers to bonus Hitpoint Damage inflicted when striking specific body parts. This damage is applied after armor mitigation.\n\n" +
	"By default, hits to the head deal " + ::MSU.Text.colorPositive("+50%") + " additional Critical Damage."
)),
```

PR #822, opened 2025-06-19
fix: non-settlement banner not receiving any banner offset after loading

Darxo:
Add brief explanation of onAfterInit limitations

This finally fixes a leftover bug from this PR: https://github.com/Battle-Modders/mod-reforged/pull/189


LordMidas (approved):

LordMidas:
Could you please move the comments to the line before the line to which they apply? Looks good otherwise.

PR #825, opened 2025-06-25
Hide combat information

Darxo:
As per midas
> The current proposal of the design is like this:
> - A "Discovered" entity is one that you have seen at least once in that battle. It doesn't matter whether the entity is currently visible or not. I believe vanilla actor.setDiscovered(true) and actor.isDiscovered() represents that already, so we could use that.
> - Idle sounds are only played by Discovered entities.
> - All other sounds are played by all entities with no new conditions.
> - We play a silent sound for music until the first enemy is Discovered. At that point the regular music starts. I can provide you with a silent .wav file for this.
> - Top bar shows enemy numbers as XX-YY where XX is enemies who are currently visible and YY is (visible + discovered). > - Maybe YY should simply be ?? instead. This might be simpler as then we don't have to worry about updating this number when invisible enemies die or retreat.

Currently in this PR these features are implemented:
- idle sounds no longer play for undiscovered enemies
- music is delayed until first enemy is discovered


LordMidas (approved):

LordMidas:
I fixed a small bug in your commits and pushed another commit to hide the number of enemies in the top bar. The PR is ready to go in from my side. It will need an `assets` update as well (due to the silent music file).

LordMidas:
- Fixed music not playing when all enemies are visible from the start.
- Changed music hiding to only be applied to location fights that hide defenders.

Both changes have been squashed into your music feat commit.

PR #832, opened 2025-07-11
refactor(style): sort hedgeknight perks similar to other enemy hooks

Darxo:
Add Reforged category with reforged additions
Sort Vanilla perks to to be in line with vanillas order 
Add commented versions of perks that vanilla added to them 
Remove commented non-existing perk_rf_heartless from hedgeknight


Uberbagel (changes requested):
They are not supposed to have battle flow in our design so it should be commented out.

PR #833, opened 2025-07-12
refactor: remove stun warning from line breaker skill

Darxo:
It commonly understood that someone who is stunned has no turn at all 
So it is redundant to write a warning about that in every active skill


LordMidas (approved):
Looks good. Could you please make the same change in rf_swordmaster_push_through_skill in the same commit and adjust the commit message accordingly before merging?

Darxo:
I found a third tooltip occurence which I also fixed now.
And I hyperlinked the Rooted term.
So as a little bit more changed a put you on review again

LordMidas (approved):

PR #842, opened 2025-09-07
Refactor attractionmult on npcs

Darxo:
The Heavy Arbalester had it coming with this mult, as the vanilla Crossbowman version already had this.
And the desert stalker getting this is in line with the regular nomad archer having the attraction multiplier.

There is an argument for also giving the Master Archer such an attraction multiplier, but I havent done that here


Uberbagel:
I expect Desert Stalker and Master Archer don't have this higher attraction mult due to their much higher ranged defenses (RDef + perks). Therefore I think the Heavy Arbalester should get the 1.1 mult but Desert Stalker should not.

PR #846, opened 2025-09-23
refactor: remove hard-coded hard-to-mod unit costs

Darxo:
All units we register automatically fetch a `Cost` value from the SpawnlistMaster entries, if no hard-coded one was defined for them.
This commit removes the hard-coded cost entries and it also moves those adjustments over into the `spawnlist_master`, where they differ from Vanilla. Hexen and Necromancer for example dont differ and therefor dont need entries there


LordMidas:
With Reforged 0.8.0 we moved to a new and more moddable system of units. So, this is no longer relevant.

PR #853, opened 2025-10-01
refactor: remove pauper and set wide attribute rolls on indebted

Uberbagel:


mod_reforged/hooks/skills/backgrounds/slave_background.nut
```diff
@@ -39,4 +37,43 @@
...
+                -10,
+                10
+            ],
+            RangedDefense = [
+                -10,
+                10
+            ],
+            Initiative = [
+                -15,
+                15
```
Darxo, line 74: This is a buff of +5 average Initiative for Slaves (from -5 to +-0). Is this intended? 
You dont buff their average in any other attribute.

Perhaps you want the range to be -20, 10 instead?

PR #855, opened 2025-10-03
fix(vanilla): getActiveEntity returning null while IsLocked == true

Darxo:
Effectively this will fix:
- actors being pushed aside by resurrecting zombies screwing field of view of active entity
- getActiveEntity returning null during onTurnStart and onTurnEnd events


LordMidas:
@Darxo Have you tried this in practice? Has it caused any issues? Should we include this in Reforged?

Darxo:
> @Darxo Have you tried this in practice? Has it caused any issues? Should we include this in Reforged?

I don't remember if that PR actually fixes the zombie bug; it should in theory

I'm not aware of side-effects. I have used a similar approach in Hardened for a long time now.
That variant was only used in a few dozen custom situations though. 

image

PR #947, opened 2025-12-15
feat(bal): rework barbarian slave in the same way as other slaves

Uberbagel:
- Replace Pauper perk group with the same as Barbarians
- Remove -1 weapon perk group
- Increase rolled range of combat attributes


Darxo (approved):

LordMidas (changes requested):
If we are going to treat all the slave backgrounds with the same attribute changes then it is better to define the `onChangeAttributes` function in the base slave class and then in the `onChangeAttributes` of the children you just call `this.slave.onChangeAttributes()`.

PR #964, opened 2026-02-20
feat: add chance to reduce condition of shields at spawn on some enemies

Uberbagel:
Discussed in VC. This brings back some capability to destroy shields, fits the Bandit/Nomad factions having some degree of shoddy/worn gear and adds a player cost (repairing the shield) to having the benefit of 'unbreakable' shields.

### DIFF
--- a/mod_reforged/hooks/entity/tactical/enemies/bandit_raider.nut
+++ b/mod_reforged/hooks/entity/tactical/enemies/bandit_raider.nut
@@ -60,7 +60,16 @@
 				[1.0, "scripts/items/shields/wooden_shield"],
 				[1.0, "scripts/items/shields/kite_shield"]
 			]).roll();
-			this.m.Items.equip(::new(shield));
+
+			if (shield != null)
+			{
+				shield = ::new(shield);
+				this.m.Items.equip(shield);
+				if (::Math.rand(1, 100) <= 25)
+				{
+					shield.setCondition(::Math.floor(shield.getCondition() * ::MSU.Math.randf(0.6, 0.8)));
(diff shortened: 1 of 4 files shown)


Darxo (commented):
Shields only have two visual states:
- "looks fully repaired"
- "looks damaged", when condition is below this value `ShowDamagedShieldThreshold`

Using a purely random range between a min- and max will make it much more annoying for the player to weigh their options when encountering enemies.
The player has to check every enemies shield to know the exact condition value when strategizing, as any condition point can matter.

I suggest do leave the condition either at 100% or 50% at random.
That way the player can see at a glance, whether a shield is damaged and worth breaking. And he will quickly learn all the break points from shields that have exactly 50% condition. No need to double-check the values

LordMidas:
I think Darxo's point is good. I'd be ok with such a change. The drawback of that change is that many shields have very high durability even at 50% condition e.g. Kite Shield will be 100.

Perhaps a middle ground between Darxo's proposal and the original design can be that all damaged shields are always dropped to a fixed value e.g. 50 durability, irrespective of their base durability. This is going off of Wooden Shield which has 100 durability in Reforged so it drops to 50 and will appear damaged at that condition.

LordMidas:
Here is a simpler thing we can do for now: show the shield durability in the Hit Factors tooltip when targeting the actor with the `split_shield` skill.

@Darxo Is it possible to add the shield durability as a bar in the actor overlay UI that one views by pressing Alt?

Darxo:
I have no experience with using hotkeys to trigger different tooltip views

PR #987, opened 2026-04-04
fix: infinite loop in npc combat if strength is non-zero float below 1

LordMidas:
Related bug report on Discord: https://discord.com/channels/1006908336991645757/1489663330036613273

### DIFF
--- a/scripts/!mods_preload/mod_reforged.nut
+++ b/scripts/!mods_preload/mod_reforged.nut
@@ -72,7 +72,7 @@ foreach (requirement in requiredMods)
 				// vanilla to roll damage per attack except these changes:
 				// We use `maxf` instead of vanilla `max` to allow dealing fractional damage as well.
 				// We use MSU.Math.randf to roll the damage instead of vanilla Math.rand as we allow
-				// utilize fractional Strength for rolling the damage.
+				// utilizing fractional Strength for rolling the damage.
 				// All of this is ok as vanilla serializes troop strength as F32.
 				local damage = ::Math.maxf(1.0, ::MSU.Math.randf(1.0, combatant.Strength) * ::Const.World.CombatSettings.CombatStrengthMult);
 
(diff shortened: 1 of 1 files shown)


LordMidas:
An alternative approach can be to use `Math.minf` when calculating `damageDealt`. The advantage of that approach is that the full strength of the opponent is utilized in receiving damage instead of wasting the remaining fraction below 1.0. The disadvantage is that it creates more opponents with fractional Strength. But maybe this disadvantage is not a problem because this will only happen if opponents with fractional Strength already existed.

However, if we do the `minf` approach in receiving damage, then we also need to do `maxf` approach in calculating the total damage a unit can do. This is currently integer-ized by the line 72
```squirrel
// This is the total damage this combatant did in this tick. Same calculation as vanilla to roll damage per attack.
local damage = ::Math.max(1, ::Math.rand(1, combatant.Strength) * ::Const.World.CombatSettings.CombatStrengthMult);
```
If we do the `minf` in receiving damage then we should change this line also to:
```squirrel
// This is the total damage this combatant did in this tick. Same calculation as vanilla to roll damage per attack.
local damage = ::Math.maxf(1.0, ::Math.rand(1, combatant.Strength) * ::Const.World.CombatSettings.CombatStrengthMult);
```
Which would mean that attackers can deal fractional damage as well. This will almost certain result in every troop having a fractional strength after a battle. Would we be ok with this?

Darxo:
For context. Vanilla already has `CombatStrengthMult = 0.4`, which is why the minf maxf variant will open the floodgates for fractional strengths

LordMidas:
Ok we decided to switch to the `minf` approach. Note:
- We now use `maxf` and `MSU.randf` in the damage rolling step. So entities will almost certainly roll fractional damage.
- We now use `minf` when dealing damage. So enemies can receive fractional damage up to their Strength.

PR #990, opened 2026-04-08
feat: show craftables for all items instead of only crafting type items

LordMidas:
This is how it looks for items that can be equipped:


Uberbagel:
I think 'equip item' and 'place item in bag' should come before the crafting lines.

Uberbagel (approved):

Darxo (approved):

PR #996, opened 2026-04-13
Fix skeleton resurrection values and equipment

Uberbagel:


mod_reforged/hooks/entity/tactical/enemies/skeleton_boss.nut
```diff
@@ -1,5 +1,10 @@
 // The Conqueror
 ::Reforged.HooksMod.hook("scripts/entity/tactical/enemies/skeleton_boss", function(q) {
+	q.create = @() { function create()
```
LordMidas: Should include `__original` in the hook and call it before your modifications. The same should be fixed in other files in this PR.

mod_reforged/hooks/entity/tactical/enemies/skeleton_boss.nut
```diff
@@ -1,5 +1,10 @@
...
+	{
+		this.m.ResurrectionValue = 20.0;
```
LordMidas: Please put the vanilla value in comments as well. The same for other files.

LordMidas (changes requested):

LordMidas (approved):

PR #997, opened 2026-04-15
feat: add ability to add dividers in tooltip text and hint entries

LordMidas:
Usage:
Pass `rf_divider` as value for the `text` key of a "text" or "hint" type tooltip entry.

image


TaroEld (approved):
Looks fine, but as commented, I would try making it visually distinct from the top/bottom dividers. You also don't necessarily need as much padding.

LordMidas:
> Looks fine, but as commented, I would try making it visually distinct from the top/bottom dividers. You also don't necessarily need as much padding.

I added the ability to pass custom cssClass name for the divider. That is rather flexible so one can use any kind of divider they want.

PR #999, opened 2026-04-16
feat: use nested names and icons during events for various classes

LordMidas:
Converts the following to nested tooltips in event and contract screens:
- Actor names
- Item names and icons
- Skill names and icons
- Faction names
- World entity names

image

### DIFF
--- a/mod_reforged/hooks/contracts/contract.nut
+++ b/mod_reforged/hooks/contracts/contract.nut
@@ -344,4 +344,20 @@
 		::Const.World.Common.RF_addSpawnlistInfo(_entity, _partyList);
 		__original(_entity, _partyList, _resources);
 	}}.addUnitsToEntity;
+
+	q.setScreen = @(__original) { function setScreen( _screen, _restartIfAlreadyActive = true )
+	{
+		::Reforged.NestedTooltips.setApplyNestingForEvents(true);
+		__original(_screen, _restartIfAlreadyActive);
+		::Reforged.NestedTooltips.setApplyNestingForEvents(false);
+	}}.setScreen;
+});
+
+::Reforged.HooksMod.hookTree("scripts/contracts/contract", function(q) {
+	q.onPrepareVariables = @(__original) { function onPrepareVariables( _vars )
+	{
+		::Reforged.NestedTooltips.setApplyNestingForEvents(true);
+		__original(_vars);
+		::Reforged.NestedTooltips.setApplyNestingForEvents(false);
+	}}.onPrepareVariables;
 });
--- a/mod_reforged/msu_systems/nested_tooltips.nut
+++ b/mod_reforged/msu_systems/nested_tooltips.nut
@@ -7,7 +7,88 @@ local getThresholdForInjury = function( _script )
 	}
 }
 
+// Apply nesting in events
+{
+	::Reforged.HooksMod.hookTree("scripts/entity/tactical/actor", function(q) {
+		q.getName = @(__original) { function getName()
+		{
+			return ::Reforged.NestedTooltips.isApplyingNestingForEvents() ? ::Reforged.Mod.Tooltips.parseString(format("[%s|EventActor+%i]", __original(), this.getID())) : __original();
+		}}.getName;
+
+		q.getNameOnly = @(__original) { function getNameOnly()
+		{
+			return ::Reforged.NestedTooltips.isApplyingNestingForEvents() ? ::Reforged.Mod.Tooltips.parseString(format("[%s|EventActor+%i]", __original(), this.getID())) : __original();
+		}}.getNameOnly;
+	});
+
+	::Reforged.HooksMod.hookTree("scripts/items/item", function(q) {
+		q.getName = @(__original) { function getName()
+		{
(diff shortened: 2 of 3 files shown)


mod_reforged/hooks/contracts/contract.nut
```diff
@@ -344,4 +344,27 @@
...
+		return ::Reforged.Mod.Tooltips.parseString(__original(_buildText));
+	}}.buildText;
+
+	q.setScreen = @(__original) { function setScreen( _screen, _restartIfAlreadyActive = true )
+	{
+		local original_IsApplyingNestingForEvents = ::Reforged.NestedTooltips.__IsApplyingNestingForEvents;
+		::Reforged.NestedTooltips.__IsApplyingNestingForEvents = true;
+		__original(_screen, _restartIfAlreadyActive);
+		::Reforged.NestedTooltips.__IsApplyingNestingForEvents = original_IsApplyingNestingForEvents;
+	}}.setScreen;
```
TaroEld, line 353: Can you wrap this inner function? It's repeated 4 times and pretty ugly. Something like "onBuildEventText" in NestedTooltip namespace.

mod_reforged/msu_systems/nested_tooltips.nut
```diff
@@ -7,7 +7,85 @@ local getThresholdForInjury = function( _script )
 	}
 }
```
TaroEld, line 9: One thing to note is that you add overhead on each call on each of these things every time. Idk if some conditional hook is possible.

mod_reforged/msu_systems/nested_tooltips.nut
```diff
@@ -7,7 +7,85 @@ local getThresholdForInjury = function( _script )
...
+
+::Reforged.HooksMod.hookTree("scripts/entity/world/world_entity", function(q) {
+	q.getName = @(__original) { function getName()
+	{
+		return ::Reforged.NestedTooltips.__IsApplyingNestingForEvents ? ::Reforged.Mod.Tooltips.parseString(format("[%s|WorldEntity+%i]", __original(), this.getID())) : __original();
+	}}.getName;
+});
+
 ::Reforged.NestedTooltips <- {
+	__IsApplyingNestingForEvents = false,
```
TaroEld: Doesn't __ denote private? You call that from outside. Should probably lose the __

mod_reforged/msu_systems/nested_tooltips.nut
TaroEld: (for the block below)

mod_reforged/msu_systems/nested_tooltips.nut
LordMidas: By "private" in this case I meant more like "Reforged-use-only" rather than other mods doing something with it. But  you're right in this case making it private doesn't make much sense. I'll remove the `__`.

mod_reforged/msu_systems/nested_tooltips.nut
LordMidas: That's true. I can't imagine any conditional hook to achieve this as it has to be checked at runtime. One thing we can do is constrain the hook on actor to `player` class only but then we lose the ability to nest non-player names e.g. Employer in contracts. Perhaps that isn't a big loss. But I'd rather not try to optimize this unless it really causes any performance issues.

mod_reforged/hooks/contracts/contract.nut
LordMidas: Every time it's a different `__original` with different parms e.g. `setScreen` of event has 1 param but `setScreen` of contract has 2 params. To make it a bit cleaner, an alternative is (I prefer this as this makes it explicit that you are applying nesting during some process which doesn't have to be a function call):
```squirrel
q.setScreen = @(__original) function( _screen )
{
  ::Reforged.NestedTooltips.setApplyNestingForEvents(true);
  __original(_screen);
  ::Reforged.NestedTooltips.setApplyNestingForEvents(false);
}
```

If I do your proposal and refactor it out then it I'll need to pass the `__original` along with its params to the extracted function. Something like this:

```squirrel
function applyNestingDuringFunction( _original, _env, ... )
{
   local original_IsApplyingNestingForEvents = ::Reforged.NestedTooltips.__IsApplyingNestingForEvents;
   ::Reforged.NestedTooltips.__IsApplyingNestingForEvents = true;
   vargv.insert(0, _env);
   _original.acall(vargv);
   ::Reforged.NestedTooltips.__IsApplyingNestingForEvents = false;
}
```
And then the 4 places where I use it I can do something like
```squirrel
q.setScreen = @(__original) function ( _screen )
{
  ::Reforged.NestedTooltips.applyNestingDuringFunction(_original, this, _screen)
}
```
The disadvantage of that would be that it would only work for calling other functions because we need to pass a function to it. You won't be able to use that method to enable nesting inside a function for part of its code without extracting that code into a separate function. That's why I feel inclined towards the first alternative I proposed above.

mod_reforged/msu_systems/nested_tooltips.nut
```diff
@@ -7,7 +7,85 @@ local getThresholdForInjury = function( _script )
 	}
 }
 
+::Reforged.HooksMod.hookTree("scripts/entity/tactical/actor", function(q) {
```
Darxo: You could wrap all of these different new hooks in something like
```js
{    // Hooks for "Nesting for Events"
[...]
}
```
to make is collapsible for easier traversing of this script, as nexted_tooltips.nut is growing large now

PR #1006, opened 2026-04-19
Feat add immunity to net in racial tooltip

LordMidas:


scripts/skills/special/rf_reach.nut
```diff
@@ -154,6 +154,17 @@ this.rf_reach <- ::inherit("scripts/skills/skill", {
 		}
 	}
 
+	function onQueryTooltip( _skill, _tooltip )
+	{
+		if (_skill.isType(::Const.SkillType.Racial) && this.getContainer().getActor().getBaseProperties().getReach() >= ::Reforged.Reach.Default.NetImmune)
+		{
+			_tooltip.push({
+				id = 100,	type = "text",	icon = "ui/icons/special.png",
```
Darxo, line 162: I suggest to use the warning icon here
```suggestion
				id = 100,	type = "text",	icon = "ui/tooltips/warning.png",
```

Darxo (changes requested):

scripts/skills/special/rf_reach.nut
LordMidas: This tooltip is for the Racial skill effect on the character himself (not the hit-factors tooltip). So, this is from the character's own point of view i.e. just like we show "Immune to Poison" in tooltips. The standard for such things is the `special` icon, not `warning`. If this was the hit-factors tooltip then I would use `warning`.

Darxo (approved):

PR #1008, opened 2026-04-19
fix: remove invalid contracts from last visit contracts

LordMidas:


mod_reforged/hooks/contracts/contract_manager.nut
```diff
@@ -7,6 +7,63 @@
...
+		// destinations or targets which no longer exist or are no longer valid.
+		foreach (contracts in this.m.RF_LastVisitContracts)
+		{
+			for (local i = contracts.len() - 1; i >= 0; i--)
+			{
+				if (contracts[i].isValid())
+					continue;
+
+				local c = contracts.remove(i);
+				::logInfo(format("Reforged: last visit contract removed: %s (%i)", c.getName(), c.getID()));
```
Darxo: Is this log intended to stay in release?

mod_reforged/hooks/contracts/contract_manager.nut
```diff
@@ -7,6 +7,63 @@
...
+		if (!_force && ::World.State.getMenuStack().hasBacksteps())
+		{
+			return;
+		}
+
+		if (!_force && ("State" in ::Tactical) && ::Tactical.State != null)
+		{
+			return;
+		}
+
```
Darxo, line 30: From my understanding of the vanilla `world_state::onUpdate` the following logic runs every "tick" currently, which is concerning from a performance pov.

The `contract_manager` for comparison only does one check every virtual BB second (105 times a day)

How important is the immedaitey/timely cleanup of those contracts?
Is this allowed to happen 1 BB second delayed, or 5?

mod_reforged/hooks/contracts/contract_manager.nut
LordMidas: I just put it in there to mirror the log that vanilla prints when removing contracts from contract_manager. It doesn't _have_ to be in there. What do you think?

mod_reforged/hooks/contracts/contract_manager.nut
LordMidas: I see, I missed that time part in the vanilla contract_manager. It is somewhat important to clear these contracts the moment they become invalid because otherwise we will get bad tooltips. The current tooltip implementation I have for `contract.nut` handles _most_ of such cases gracefully i.e. it will ignore `null` destinations etc. But there can be cases where this may lead to the tooltip not appearing at all if someone has a custom tooltip implementation and doesn't expect these things to be null.

I think we can implement the same time check as vanilla contract manager.

mod_reforged/hooks/contracts/contract_manager.nut
Darxo: I would remove it. Its just more noise in the log which I dont like

Darxo: Maybe gate it behind some debug flag?

Darxo (approved):

mod_reforged/hooks/contracts/contract_manager.nut
LordMidas: That's a nice idea. Perhaps in the future we can add that (along with some other flags here and there). For now I just removed it.

PR #1009, opened 2026-04-22
Uber changes

Uberbagel:


LordMidas (approved):
The last commit about the marshal should be changed to `fix`. The rest looks good.

scripts/entity/tactical/enemies/rf_draugr_warrior_bodyguard.nut
```diff
@@ -7,4 +7,14 @@ this.rf_draugr_warrior_bodyguard <- ::inherit("scripts/entity/tactical/enemies/r
...
 	}
+
+	function assignRandomEquipment()
+	{
+		this.rf_draugr_warrior.assignRandomEquipment();
+
+		if (this.m.Items.hasEmptySlot(::Const.ItemSlot.Offhand))
+		{
+			this.m.Items.equip(::new("scripts/items/shields/rf_draugr/rf_draugr_round_shield"));
+		}
```
LordMidas, line 16: This won't work because the parent's function can equip a 2-handed weapon. You need to first equip the shield and then call the parent's function afterward.

LordMidas (changes requested):

scripts/entity/tactical/enemies/rf_draugr_warrior_bodyguard.nut
```diff
@@ -7,4 +7,14 @@ this.rf_draugr_warrior_bodyguard <- ::inherit("scripts/entity/tactical/enemies/r
 		this.m.AIAgent = ::new("scripts/ai/tactical/agents/rf_draugr_bodyguard_agent");
 		this.m.AIAgent.setActor(this);
 	}
+
+	function assignRandomEquipment()
+	{
+		this.rf_draugr_warrior.assignRandomEquipment();
```
LordMidas: Probably draugr huskarl bodyguard needs a similar treatment.

Uberbagel: It did not in fact need a similar treatment.

LordMidas (approved):

PR #1010, opened 2026-04-25
feat: add a standardized patch registration file

LordMidas:
This makes it easy to track patches in the log.html via their type and versioning and to ensure that users delete those patches when updating Reforged to the next build.

### DIFF
--- a/scripts/!mods_preload/mod_reforged_patch/mod_reforged_patch_registration.nut
+++ b/scripts/!mods_preload/mod_reforged_patch/mod_reforged_patch_registration.nut
@@ -1,0 +1,33 @@
+/*
+## What is a patch?
+A patch is NOT a complete Reforged build. Instead, it is meant to contain only a few files
+that have changes in them meant to fix a specific issue for a user or to "field-test" the fix
+for a specific issue before it is included in the main build.
+
+## How to make a patch:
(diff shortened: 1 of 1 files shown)


scripts/!mods_preload/mod_reforged_patch/mod_reforged_patch_registration.nut
```diff
@@ -0,0 +1,31 @@
...
+  - Command: BBBuilder.exe build "REFORGED_PATH" -diff development,patch
+*/
+local t = {
+	Version = "1.0.0", // Each patch has its own versioning starting at 1.0.0.
+	ID = "", // ID must be unique for each patch.
+	Name = "Reforged Patch", // Give each patch a unique name.
+};
+
+if (t.ID == "")
+	return;
```
Darxo, line 26: this could also trigger a popup, so that the modder knows immediately, that their patch is not doing anything at all

LordMidas: The purpose of this guard is to prevent this patch registration file from registering with hooks in a regular build of the mod. The idea is that when shipping an actual patch, the ID will not be an empty string. But in a regular build, the ID will be an empty string here (unless you exclude this file from the build).

PR #1012, opened 2026-04-29
fix(vanilla): slash_lightning triggering lightning during enemy turn

Darxo:
Todo @self: Test this out once


LordMidas:
Should this not be in Modular Vanilla instead? Btw, I have opened a bug report for it now for vanilla:
https://steamcommunity.com/app/365360/discussions/1/841752762653941021/

LordMidas:
I have implemented this fix in a [commit](https://github.com/Battle-Modders/mod_modular_vanilla/commit/0773560b1fe791428c7691d7ad8fec384dea88cd) over at Modular Vanilla and have set you as the author of that commit.

PR #1015, opened 2026-04-29
feat: show predicted hitchances in tile tooltip during movement preview

LordMidas:
The main changes are in the first commit. The second commit is simply adapting some existing skills to use this system.

Now whenever a character is previewing movement, hovering over any empty tile will show the predicted chances to hit and to be hit from that tile against relevant enemies (i.e. attackable enemies and enemies that can attack you).

The system uses some switcheroos and hacky methods to achieve as accurate of a prediction as possible including the effects of terrain, zone of control, skills/perks and their effects etc.

Note: Any skill that has triggers for things affecting potential hitchances in `onMovementFinished` etc. will need to properly account for those to make the prediction work reliably. Most skills don't fall into this category, but those that do will need to implement support for this system otherwise they won't be accurately taken into account. See the second commit to see how I adjust some relevant Reforged skills accordingly e.g. Formidable Approach.

In addition to the screenshot below, there is also a [video demonstration](https://discord.com/channels/1006908336991645757/1496970977677152296/1498921679311863839) of the system on the discord.

### DIFF
--- a/mod_reforged/hooks/native/TacticalTile.nut
+++ b/mod_reforged/hooks/native/TacticalTile.nut
@@ -1,0 +1,47 @@
+// Hook to return as if the previewing entity has ZoC on this tile. Relevant for hit-chance prediction during movement preview.
+local hasZoneOfControlOtherThan = ::TacticalTile.hasZoneOfControlOtherThan;
+::TacticalTile.hasZoneOfControlOtherThan <- { function hasZoneOfControlOtherThan( _factions )
+{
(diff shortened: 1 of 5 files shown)


Uberbagel:
Can this whole system be turned off in Mod Settings?

LordMidas:
> Can this whole system be turned off in Mod Settings?

Not yet. But that seems like a good thing to include. Currently, what a user can set in the mod settings is the "Threshold" for hit-chance below which enemies collapse into a single entry. By default it is set to 5%.

LordMidas:
I have added mod setting to enable/disable the predicted hit-chance display. Can enable/disable it by setting it to All, Player Only, AI only, or None.

Uberbagel (approved):

PR #1019, opened 2026-04-30
feat: allow unused throwing weapons to drop from bag

LordMidas:
Vanilla does not allow throwing weapons to drop from bag slot if they are at full ammo.

See [related discussion](https://discord.com/channels/1006908336991645757/1296562347066003560/1499459302857900102) on the Reforged Discord.

### DIFF
--- a/mod_reforged/hooks/items/weapons/weapon.nut
+++ b/mod_reforged/hooks/items/weapons/weapon.nut
@@ -15,7 +15,7 @@
 	q.isDroppedAsLoot = @(__original) { function isDroppable()
 	{
 		local original_CurrentSlotType = this.m.CurrentSlotType;
-		this.m.CurrentSlotType = ::Const.ItemSlot.Mainhand;
+		this.m.CurrentSlotType = this.m.SlotType;
 		local ret = __original();
 		this.m.CurrentSlotType = original_CurrentSlotType;
 		return ret;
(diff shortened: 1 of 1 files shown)


LordMidas:
I have converted this to a draft. I have just realized that this has major balance conern because this means that throwing weapons in bag will drop 90% of the time. So we probably need a different method to tune that roll chance.

mod_reforged/hooks/items/weapons/weapon.nut
```diff
@@ -8,6 +8,18 @@
...
 		}
 	}}.create;
+
+	// Vanilla does not allow weapons at their AmmoMax to drop from bag slot.
+	// We want such weapons to drop, so we switcheroo their CurrentSlotType
+	// before running the original function.
+	q.isDroppedAsLoot = @(__original) { function isDroppable()
+	{
+		local original_CurrentSlotType = this.m.CurrentSlotType;
+		this.m.CurrentSlotType = ::Const.ItemSlot.Mainhand;
```
Darxo: There are also a few weapons, who are not equipped in the `Mainhand`.
One of them is `throwing_net`

It is more correct and probably slightly safer to only switcheroo them into SlotTypes, which make sense on them.
```suggestion
		this.m.CurrentSlotType = this.m.SlotType;
```

Darxo (changes requested):

PR #1024, opened 2026-05-04
feat(bal): increase xp of bandit units

LordMidas:
We had dropped the XP compared to vanilla some time ago to "encourage" the players to fight other factions. But in my experience this leads to the opposite i.e. more grind. So, to reduce early game grind we increase the XP again.

As discussed in [this thread](https://discord.com/channels/996482927531671692/1496953625510023208/1496953625510023208) on the dev discord.

### DIFF
--- a/mod_reforged/hooks/config/faction_bandits.nut
+++ b/mod_reforged/hooks/config/faction_bandits.nut
@@ -1,7 +1,7 @@
 ::Reforged.Entities.editEntity("BanditThug",
 	null, // Vanilla Cost 9
 	{
-		XP = 120, // vanilla 150
+		XP = 150,
 		ActionPoints = 9,
 		Hitpoints = 55,
 		Bravery = 40,
@@ -34,7 +34,7 @@
 		Script = "scripts/entity/tactical/enemies/rf_bandit_pillager"
 	},
 	{
-		XP = 200,
+		XP = 225,
 		ActionPoints = 9,
 		Hitpoints = 75,
 		Bravery = 50,
@@ -59,7 +59,7 @@
 		Strength = 23 // vanilla 20
 	},
 	{
-		XP = 250,
+		XP = 300,
 		ActionPoints = 9,
 		Hitpoints = 80, // vanilla 75
 		Bravery = 60, // vanilla 55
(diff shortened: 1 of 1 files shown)


Uberbagel:
These are now higher than before 0.8.0. I think they only need to go back to pre 0.8.0 numbers.

LordMidas:
Vanilla:
- Thug 150
- RaiderLOW 250
- Raider 250
- Leader 375

In Reforged, the Pillager is comparable to RaiderLOW while our Raider is slightly stronger than vanilla Raider. In my new setup I have set Pillager below 250 i.e. at 225, and that makes space for Raider to be at 300. The player encounters fewer Raiders than in vanilla (because they are now more expensive) and encounters more Pillagers because they are cheaper. So, he gains less XP from Pillagers, but the less frequent encountesr with Raiders grants 300 XP so overall I expect this to behave in a similar ballpark as the 250 from both RaiderLOW and Raider in vanilla. If the player were encountering equal numbers of Pillagers and Raiders than setting one to 225 and the other 275 would be more accurate. But the encountering is skewed towards Pillagers.

Marauder is higher tier than Raider, so if Raider is 300, then 375 seems fitting for Marauder. We have already bumped Leader to 400 in Reforged anyway. Marauder is comparable to Barbarian Chosen (which are 350) but has better stats than Chosen in some things e.g. Melee Skill.

Therefore, I think the new numbers I have proposed are a better fit than any we have had before.

150 -> 225 -> 300 -> 375 is also a nicely even +75 increase per tier (but this is mostly an aesthetic point :P ).

Darxo (approved):

Uberbagel:
This does not match with our initiative for 0.8.0 to normalize our xp values with vanilla (excepting leaders and elites/specials). If Highwayman goes above 350 the consistent thing to do is allow that for all the factions.

LordMidas:
Upon further discussion we decided to set the highest tier at 350, in line with other factions. This seems fine as this faction is front-loaded on XP on the lower tiers so it is ok to taper it off at the highest tier. 350 and 375 is only a 7% difference anyway.

Uberbagel (approved):

PR #1027, opened 2026-05-06
feat: add ability to auto-negotiate and skip screens in contracts

LordMidas:
This is a quality of life feature as [discussed](https://discord.com/channels/1006908336991645757/1296562347066003560/1500868199804305599) with the players on the Reforged Discord.

- Add a mod setting to auto-skip screens in contracts to either Negotiation or Overview (final screen).
- Add a mod setting to auto-negotiate contracts a given number of times until the employer gives his final offer or the negotiations fail.

Default values for both are "Overview" and "1".

EDIT: Default values changed to "Disabled" and "0" after some discussion.

image
image

### DIFF
--- a/mod_reforged/hooks/contracts/contract.nut
+++ b/mod_reforged/hooks/contracts/contract.nut
@@ -1,5 +1,29 @@
 ::Reforged.HooksMod.hook("scripts/contracts/contract", function(q)
 {
+	// Will auto-negotiate this contract for _numAttempts times.
+	// Will stop if the employer gives his final offer or if negotiations fail.
+	q.RF_autoNegotiate <- { function RF_autoNegotiate( _numAttempts = 1 )
+	{
+		if (!this.isStarted())
+			this.start();
+
+		if (this.m.ActiveScreen.ID == "Negotiation.Fail" || this.m.Payment.IsFinal)
+			return;
+
+		if (this.m.ActiveScreen.ID != "Negotiation")
+			this.setScreen("Negotiation");
+
+		for (local i = 0; i < _numAttempts; i++)
+		{
+			this.setScreen(this.getScreen(this.m.ActiveScreen.Options[1].getResult()));
+			if (this.m.ActiveScreen.ID == "Negotiation.Fail")
+				return;
+
+			if (this.m.Payment.IsFinal)
+				break;
+		}
+	}}.RF_autoNegotiate;
+
 	// Is used to "start" a contract so that its home, origin, destination, payment etc.
 	// is fully set and generated.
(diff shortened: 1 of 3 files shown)


Darxo:
I dont like the autonegotiate feature.
By the time we automate certain game mechanics, we might aswell remove them completely. 
Negotiations might aswell not exist and contracts pay 5% more by default (and perhaps give 10% less relation) to simulate that.

LordMidas:
> I dont like the autonegotiate feature. By the time we automate certain game mechanics, we might aswell remove them completely. Negotiations might aswell not exist and contracts pay 5% more by default (and perhaps give 10% less relation) to simulate that.

I must say your point is fair. But unlike removal of negotiation, which is a sledgehammer change, this is an optional mod setting. Someone may want to do it the old "organic" way and click on those negotiating options manually. So, removing Negotiating from the game like that completely robs that player of that possibility. Meanwhile, these settings add a way for others who want to automate it to achieve it in a QoL way without having to click 5 buttons every time. Of course, these automators are paying the price of building relations slowly. So, for them the result is the same as complete removal of negotiation, higher contract pay, and slower relations gain. But the organic players retain their ability to play organically.

mod_reforged/hooks/contracts/contract.nut
```diff
@@ -1,5 +1,29 @@
...
+		if (this.m.ActiveScreen.ID != "Negotiation")
+			this.setScreen("Negotiation");
+
+		for (local i = 0; i < _numAttempts; i++)
+		{
+			this.setScreen(this.getScreen(this.m.ActiveScreen.Options[1].getResult()));
+			if (this.m.ActiveScreen.ID == "Negotiation.Fail")
+				return;
+
+			if (this.m.Payment.IsFinal)
```
Darxo, line 24: What does `IsFinal` has to do with the autonegotiate feature?

From my understanding this variable only ever determines, which response-text you get.
It has no influence over whether you can continue negotiating and may still get an additional raise.

So "... as soon as the emplyoer gives his final offer ..." from the mod setting is not correct from my understanding.

Of course it is an indirect indicator for how well you do. Because on average, if you succeed your first negotiation, you will have a higher chance to also succeed your next one, because you will have built less annoyance (on average).
If this indirect indicator is the reason to use that (and I understand all of the correctly), then we should add a comment  explaining that.

LordMidas: You are right. I checked and it seems that `IsFinal` only really determines the text that the negotation screen shows (and it also shows that your last attempt of negotiation did not increase the payment i.e. that attempt was refused by the employer). I will add a comment.

Darxo (approved):

PR #1029, opened 2026-05-11
feat: improve vanilla mult logic in properties for negative values

LordMidas:
In vanilla negative values are always multiplied with (1.0 / mult) This causes mults between 0.0 and 1.0 i.e. those that would reduce a positive value to increase the negative value by double the amount. E.g. a 50% reduction on a positive value becomes a 2x increase on the negative value (in the negative direction). We implement a more intricate system whereby this behavior is only for mults > 1.0. For mults between 0.0 and 1.0 we instead multiply by 2.0 - mult. This results in a 50% reduction on a positive value to be a 50% increase on the negative value in the negative direction. So it behaves more intuitively.

### DIFF
--- a/mod_reforged/hooks/config/character.nut
+++ b/mod_reforged/hooks/config/character.nut
@@ -10,7 +10,17 @@ local original_getClone = ::Const.CharacterProperties.getClone;
 	DefensiveReachIgnore = 0,
 	OffensiveReachIgnore = 0,
 	BonusPerReachAdvantage = 0,
-	RF_BleedingEffectMult = 1.0
+	RF_BleedingEffectMult = 1.0,
+
+	function getClone()
+	{
+		local ret = original_getClone();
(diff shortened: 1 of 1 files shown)


mod_reforged/hooks/config/character.nut
```diff
@@ -22,14 +32,47 @@ local original_getClone = ::Const.CharacterProperties.getClone;
 		return ::Math.floor(this.FatigueRecoveryRate * this.FatigueRecoveryRateMult);
 	}
 
-	function getClone()
+	// In vanilla negative values are always multiplied with (1.0 / mult)
+	// This causes mults between 0.0 and 1.0 i.e. those that would reduce a positive value
```
Darxo, line 36: You should also write somewhere, that this fixes a division by 0 vanilla bug, when a multiplier of 0.0 is being applied on a negative value

Darxo (approved):

PR #1044, opened 2026-06-20
feat(chg): normalize zone of control tooltips in a few skills

Darxo:
These three skills are outliers in their phrasing compared to 12 other zone of control related tooltips

### DIFF
--- a/mod_reforged/hooks/skills/actives/charge.nut
+++ b/mod_reforged/hooks/skills/actives/charge.nut
@@ -38,7 +38,7 @@
 				id = 20,
 				type = "text",
 				icon = "ui/icons/warning.png",
-				text = ::Reforged.Mod.Tooltips.parseString("Cannot be used when [engaged|Concept.ZoneOfControl] in melee")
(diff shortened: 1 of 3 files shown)


mod_reforged/hooks/skills/actives/charge.nut
```diff
@@ -38,7 +38,7 @@
 				id = 20,
 				type = "text",
 				icon = "ui/icons/warning.png",
-				text = ::Reforged.Mod.Tooltips.parseString("Cannot be used when [engaged|Concept.ZoneOfControl] in melee")
+				text = ::Reforged.Mod.Tooltips.parseString("Cannot be used because this character is [engaged|Concept.ZoneOfControl] in melee")
```
LordMidas, line 41: I prefer the wording `when engaged in melee` instead of `because this character is engaged` because that keeps the tooltip more appropriate when this appears as a nested tooltip (e.g. using the dummy player).

Darxo: This tooltip will never show up for the dummy player because the dummy player is never engaged in melee

LordMidas (approved):

PR #1046, opened 2026-06-23
fix(vanilla): serialize last tavern rumor

Darxo:
Currently the value of `tavern_building::m.LastRumor` is not serialized. It will default to "" whenever you load a save.
This results in you getting a free tavern rumor when visiting that same tavern again.
This cheese can be used to endlessly receive new tavern rumors for free.

We can't serialize that value during tavern_building::onSerialize, because at that time, the flags of our settlement has already been serialized 

We create a second getRumor hook, because adding our serialization into the first one requires adding the code twice
- once after the __original call
- once at the very end

Also the existing getRumor hook is already very large


Darxo:
Moved into Modular Vanilla https://github.com/Battle-Modders/mod_modular_vanilla/pull/100

PR #1051, opened 2026-06-24
Adjust code for vanilla 1.5.2.2 update

LordMidas:
Look at the individual commits for a cleaner look at what's happening.


mod_reforged/hooks/entity/world/settlements/buildings/marketplace_oriental_building.nut
```diff
@@ -0,0 +1,47 @@
+::Reforged.HooksMod.hook("scripts/entity/world/settlements/buildings/marketplace_oriental_building", function(q) {
+	q.getDefaultShopList = @(__original) { function getDefaultShopList()
+	{
+		local ret = __original();
+
+		if (this.getSettlement().getTile().SquareCoords.Y > ::World.getMapSize().Y * 0.7)
```
Darxo: In Vanilla and Reforged, oriental cities are never even close to the upper 30% of the map.
Also they are, by design, desert cities.
Should the upper 30% ever be desert, then this rule would also stop making sense.

I think we can safely remove the `if` block

sSs1897:
Is there any reason to keep the warhound check in marketplace_oriental?

sSs1897 (commented):
Tested the shop and seems good

mod_reforged/hooks/entity/world/settlements/buildings/marketplace_oriental_building.nut
LordMidas: Good point. I've removed that block.

Darxo (approved):

PR #1054, opened 2026-06-30
feat: adapt vanilla executioner background

LordMidas:
The `executioner_southern_background` inherits from this background so it gets the same treatment automatically.

- This background spawns with axes and swords, so I gave him Axe and Sword guaranteed (it also seems thematic for this background). I gave the Power fighting style guaranteed too. Note: The southern variant can spawn with 2h scimitar etc. which is a cleaver, so I've given 2x multiplier for cleaver for this background.
- I gave him +1 weapon perk group so a total of 4. This makes up for the 2 guaranteed axe, sword groups to give some variety in the different recruits of this background.
- He is only a melee background, so will never roll ranged perk groups.
- To make him interesting, I've given him the Executioner perk for free (like we give Sword Mastery for free to Swordmasters) and that perk seems fitting for this background.
- The 2x multiplier for Vicious and Unstoppable also were given for thematic reasons because of the perks they contain (Berserk, Fearsome, Executioner etc.).

### DIFF
--- a/mod_reforged/hooks/skills/backgrounds/executioner_background.nut
+++ b/mod_reforged/hooks/skills/backgrounds/executioner_background.nut
@@ -1,0 +1,71 @@
+::Reforged.HooksMod.hook("scripts/skills/backgrounds/executioner_background", function(q) {
+	q.createPerkTreeBlueprint = @() { function createPerkTreeBlueprint()
+	{
+		return ::new(::DynamicPerks.Class.PerkTree).init({
+			DynamicMap = {
+				"pgc.rf_exclusive_1": [],
+				"pgc.rf_shared_1": [],
+				"pgc.rf_weapon": [
+					"pg.rf_axe",
+					"pg.rf_sword"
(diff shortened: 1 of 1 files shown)


Darxo:
But this Background already comes with a built-in +10% Headshot chance as a background bonus (like Killer on the Run)
Are you sure, that you want to give him a free perk on top of that?

sSs1897:
When you buff a bg the easiest way to balance is to increase its price, and I think there is room for a price increase on executioner.

LordMidas:
I was not aware that this background had +10% chance to hit head. After giving it more thought I incline towards the following:
- Remove the +10% chance to hit head from this background. I don't see a reason why Executioner background should have such a bonus. They are trained to chop someone's head off at the chopping block. They don't have combat training or other dexterity training to target heads.
- Keep the Executioner perk that I added in this PR. It gives the background character and is not too big of a buff as the bonus is rather conditional.

PR #1055, opened 2026-07-05
feat: replace filename checking with file count checking

LordMidas:
Benefits:
- Avoids issues due to NexusMods changing their file naming scheme.
- Avoids false positives due to some submods naming themselves as `mod_reforged_something`.
- Achieves the original goal of preventing people from running multiple copies (e.g. older and newer versions) of Reforged by mistake.

### DIFF
--- a/mod_reforged/mod_conflicts.nut
+++ b/mod_reforged/mod_conflicts.nut
@@ -53,7 +53,7 @@
 	foreach (filePath in ::IO.enumerateFiles("data/"))
 	{
 		// Skip patch files
-		if (filePath.find("data/mod_reforged") != null && filePath.find("mod_reforged_patch") == null)
+		if (filePath.find("data/mod_reforged_core") != null || filePath.find("data/mod_reforged_assets") != null)
 		{
 			reforgedFiles.push(filePath.slice(5));
 			continue;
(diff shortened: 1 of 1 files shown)


Darxo (approved):

mod_reforged/mod_conflicts.nut
```diff
@@ -68,55 +69,32 @@
...
-		if (filename.len() > 22 && filename.slice(0, 18) == "mod_reforged_core_")
 		{
-			nameToMatch = filename.slice(0, 18 + ::Reforged.Version.len());
-		}
-		else if (filename.len() > 24 && filename.slice(0, 20) == "mod_reforged_assets_")
-		{
-			nameToMatch = filename.slice(0, 20 + ::Reforged.Assets.Version.len());
+			if (reforgedFiles.len() > 1)
+			{
+				::Hooks.errorAndQuit("You have extra Reforged zip files in your data folder. Delete the extra ones and keep the latest version only.");
```
Darxo, line 80: We need a comment here explaining this case.
Afaik the name `mod_reforged` is only relevant for developers and it means that both core and assets are part of the same singular file.

Darxo (approved):

Darxo (approved):

PR #1056, opened 2026-07-05
fix: revert the usage of handleInvoluntaryMovement

LordMidas:
This reverts commits e2699128b1ccfcd45a9b98f6c5bdb6f07f11f5e2 and 24d2fc0014d31a9c0b1fd0cf227fc7b246f9ce4d.

The vanilla function requires the `_onMovementCompleted` callback to have a single parameter, so our callbacks were not working. The vanilla function is also limited in terms of the information passed in the tags and we need extra information.

See my detailed criticism of the vanilla function in this Discord thread: https://discord.com/channels/965324395851694140/1523321576739569716


sSs1897:
Depending on Sato's reply but I think it might make more sense to change the handler if they wont do it? This should work for now.

sSs1897 (approved):

Darxo (approved):

PR #1061, opened 2026-07-12
feat(chg): exclude flying skulls from receiving injuries

Darxo:
Currently Flying Skulls can receive injuries. As they are only a skull, almost all injuries look out of place.
Also because of their low hitpoints, injuries dont make much sense to have play a role on them

Because every undead gets the injury receiver effect, we need to manually remove it again, when we want a specific undead to be totally immune to injuries


LordMidas (approved):
Should be moved to the bottom of our onInit hook into the `// Reforged` section. Then looks good to merge after that.

PR #1062, opened 2026-07-13
fix: ctrl hotkey reminder showing up on enemy active skills

Darxo:
This solution is not perfect but it catches most problematic tooltips.
The only thing it can't catch is when you view the actor tooltip of the active entity and view active skills in here

## Before

image

## After

image

### DIFF
--- a/mod_reforged/hooks/skills/skill.nut
+++ b/mod_reforged/hooks/skills/skill.nut
@@ -192,7 +192,7 @@
 	{
 		local ret = __original();
 
-		if (::Tactical.isActive() && this.isActive() && !this.m.IsTargeted)
+		if (::Tactical.isActive() && this.isActive() && !this.m.IsTargeted && ::Tactical.TurnSequenceBar.isActiveEntity(this.getContainer().getActor()))
 		{
 			ret.push({
 				id = 100,


mod_reforged/hooks/skills/skill.nut
```diff
@@ -192,7 +192,7 @@
 	{
 		local ret = __original();
 
-		if (::Tactical.isActive() && this.isActive() && !this.m.IsTargeted)
+		if (::Tactical.isActive() && this.isActive() && !this.m.IsTargeted && this.getContainer().getActor() == ::Tactical.TurnSequenceBar.getActiveEntity())
```
LordMidas: 1. `container.getActor()` should return an instance of `::WeakTableRef` whereas `TurnSequenceBar.getActiveEntity()` should return a table (at least in vanilla BB). So I am surprised this is even working. Do you have a mod that changes the turn sequence bar to hold `WeakTableRef` instances, or `container.getActor()` to hold a table instead?
2. How about checking that the container's actor is player controlled?

Darxo: I have removed Hardened when taking the screenshots.
Might still be another of my mods causing the weird behavior: Can't think of one though

Player controlled actors catches less problematic situations than this approach.

LordMidas: We can simplify the code by using the MSU function `isActiveEntity`.
```suggestion
		if (::Tactical.isActive() && this.isActive() && !this.m.IsTargeted && ::Tactical.TurnSequenceBar.isActiveEntity(this.getContainer().getActor())
```

LordMidas (approved):

PR #1063, opened 2026-07-14
feat: rewrite double grip condition check

sSs1897:
This is changed such that it can be compatible with other mods. Mod prompted this is refillable net in tweaks and fixes. I also pinged sato to change the code on that end to use similar tactics. 
People asking about compatibility in rf discord:
https://discord.com/channels/1006908336991645757/1006908337981509725/1521986074119770285
Ping to sato:
https://discord.com/channels/965324395851694140/965325048015646770/1526622665044004894

### DIFF
--- a/mod_reforged/hooks/skills/special/double_grip.nut
+++ b/mod_reforged/hooks/skills/special/double_grip.nut
@@ -14,27 +14,31 @@
 	}}.create;
 
 	// Overwrite vanilla function to allow double-gripping with southern swords with the perk_rf_en_garde perk
-	q.canDoubleGrip = @() { function canDoubleGrip()
+	// Return original function for compatibility with other mods
+	q.canDoubleGrip = @(__original) { function canDoubleGrip()
 	{
-		local actor = this.getContainer().getActor();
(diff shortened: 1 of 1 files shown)


LordMidas (approved):

mod_reforged/hooks/skills/special/double_grip.nut
```diff
@@ -14,27 +14,31 @@
 	}}.create;
 
 	// Overwrite vanilla function to allow double-gripping with southern swords with the perk_rf_en_garde perk
-	q.canDoubleGrip = @() { function canDoubleGrip()
+	// Return original function for compatibility with other mods
```
Darxo: That comment is not necessary.
We call the original function in the majority of cases to retain compatibility. That should be seen as the default.
The exception, e.g. Overwrites, however should be noted, as it makes this less compatible.

As a result I suggest to replace both comments with something like this
```
// Allow double-gripping with southern swords with the perk_rf_en_garde perk
```

Darxo (changes requested):

Darxo (approved):

PR #1065, opened 2026-07-17
feat: support new damage type specific damage mitigation properties

Darxo:
The goal here is to make damage type specific mitigation code streamlined and shorter

### DIFF
--- a/mod_reforged/hooks/config/character.nut
+++ b/mod_reforged/hooks/config/character.nut
@@ -12,6 +12,10 @@ local original_getClone = ::Const.CharacterProperties.getClone;
 	BonusPerReachAdvantage = 0,
 	RF_BleedingEffectMult = 1.0,
 
+	RF_DamageReceivedRegularMultByType = array(::Const.Damage.DamageType.len(), 1.0);
+	RF_DamageReceivedArmorMultByType = array(::Const.Damage.DamageType.len(), 1.0);
+	RF_DamageReceivedTotalMultByType = array(::Const.Damage.DamageType.len(), 1.0);
+
 	function getClone()
 	{
 		local ret = original_getClone();
--- a/mod_reforged/hooks/config/character.nut
+++ b/mod_reforged/hooks/config/character.nut
@@ -12,9 +12,9 @@ local original_getClone = ::Const.CharacterProperties.getClone;
 	BonusPerReachAdvantage = 0,
 	RF_BleedingEffectMult = 1.0,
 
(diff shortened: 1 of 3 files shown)


mod_reforged/hooks/config/character.nut
```diff
@@ -12,6 +12,10 @@ local original_getClone = ::Const.CharacterProperties.getClone;
 	BonusPerReachAdvantage = 0,
 	RF_BleedingEffectMult = 1.0,
 
+	RF_DamageReceivedRegularMultByType = array(::Const.Damage.DamageType.len(), 1.0);
+	RF_DamageReceivedArmorMultByType = array(::Const.Damage.DamageType.len(), 1.0);
+	RF_DamageReceivedTotalMultByType = array(::Const.Damage.DamageType.len(), 1.0);
```
LordMidas: I would say something like `RF_DamageTypeReceivedXYZMult` is a cleaner name.  So:
- `RF_DamageTypeReceivedRegularMult`
- `RF_DamageTypeReceivedArmorMult`
- `RF_DamageTypeReceivedTotalMult`

Then on the attacker side we should add similar properties to allow tweaking outgoing buffs/maluses to specific damage types.
- `RF_DamageTypeRegularMult`
- `RF_DamageTypeArmorMult`
- `RF_DamageTypeTotalMult`

mod_reforged/hooks/entity/tactical/actor.nut
```diff
@@ -16,6 +16,7 @@
 		this.getSkills().add(::new("scripts/skills/special/rf_reach"));
 		this.getSkills().add(::new("scripts/skills/special/rf_formidable_approach_manager"));
 		this.getSkills().add(::new("scripts/skills/special/rf_direct_damage_limiter"));
+		this.getSkills().add(::new("scripts/skills/special/rf_damage_type_mitigation"));
```
LordMidas, line 19: It doesn't have to be mitigation (e.g. it could be increased damage from a certain damage type), so a better name for the effect could be `rf_damage_type_mult_manager`.

scripts/skills/special/rf_damage_type_mitigation.nut
```diff
@@ -0,0 +1,23 @@
+// Implements Damage Resistences on a damage type basis
+this.rf_damage_type_mitigation <- ::inherit("scripts/skills/skill", {
+	m = {
+	},
```
LordMidas:
```suggestion
	m = {},
```

PR #1067, opened 2026-07-17
Wip/ci test workflow

Darxo:
This new workflow will compile Reforged Code every time someone creates or pushes onto a Pull Request.
When that compilation fails, then 
- you see a red warning under the PR
- you can view the BBBuilder debug output on the website

This currently doesnt yet prevent merging on a fail. But we can enable that seperately in the github settings for this repo.

I will squash the first commits, if the argument support is really not required

## Example

Look into this PR: https://github.com/Battle-Modders/mod-reforged/pull/1066

image

### DIFF
--- a/.github/workflows/test-compile.yml
+++ b/.github/workflows/test-compile.yml
@@ -1,0 +1,34 @@
+name: Verify Compilation
+
+on:
+  workflow_dispatch:
+  pull_request:
+
+jobs:
+  compile:
+    runs-on: windows-latest
+
+    env:
+      MOD_ROOT: ${{ github.workspace }}\
+      MOD_PATH: ${{ github.workspace }}\repo
+      GAME_PATH: ${{ github.workspace }}\_game_fake
+
+    steps:
+      - uses: actions/checkout@v7
+        with:
+          path: repo
+          ref: ${{ github.event.pull_request.head.sha }} # Checks out the exact PR commit
+
+      - name: Setup BBBuilder
+        id: bbbuilder
+        uses: ./repo/.github/actions/setup-bbbuilder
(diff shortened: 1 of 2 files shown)


.github/workflows/test-compile.yml
```diff
@@ -0,0 +1,33 @@
...
+  compile:
+    runs-on: windows-latest
+
+    env:
+      MOD_ROOT: ${{ github.workspace }}\
+      MOD_PATH: ${{ github.workspace }}\repo
+      GAME_PATH: ${{ github.workspace }}\_game_fake
+
+    steps:
+      - uses: actions/checkout@v4
```
LordMidas: Use a newer version of the checkout action to prevent NodeJS outdated warnings.
```suggestion
      - uses: actions/checkout@v7
```

LordMidas (changes requested):

.github/workflows/test-compile.yml
```diff
@@ -0,0 +1,33 @@
...
+        with:
+          path: repo
```
LordMidas, line 19:
```suggestion
        with:
          path: repo
          ref: ${{ github.event.pull_request.head.sha }} # Checks out the exact PR commit
```
Based on a quick study of what the `checkout` action does in the context of a `pull_request` event, it seems it creates a temporary merge commit from this PR onto the target branch and checks that out. This is problematic because:
- The action will not run if the PR has merge conflicts with the target.
- The compilation will fail if there is a compilation error in the target branch, even if the code in this PR is correct.
So, we should explicitly check out the latest commit on the PR.

LordMidas (approved):

PR #1069, opened 2026-07-18
fix: hired bros being able to roll non-sword perks in swordmaster origin

LordMidas:
Things that roll new weapon perk groups into the perk tree, e.g. Promised Potential, should not give the bros non-sword melee groups in the Old Swordmaster origin.


Darxo (approved):

sSs1897:
So, a bro can start with mace + cleaver and on realizing his potential he will almost always get sword only and keep mace + cleaver, and he still would not get the swordmaster perk group.
If a bro started with sword however, he will almost always get two ranged perk groups on realize, which is likely very heavily biased towards throwing on 50-60 ratk at lvl11?

PR #1079, opened 2026-08-10
fix: polearm triarii not assigned script

sSs1897:


LordMidas (approved):

LordMidas:
A more descriptive commit message could be `fix: polearm triarii troop using regular triarii script"

PR #1081, opened 2026-08-16
fix: release falcon was missing effect

sSs1897:


LordMidas (approved):
Commit message needs adjustment. For "fixes" we allow using `fix:` as part of the commit message's verb. So this could be:
```
fix: release falcon tooltip not working due to missing variable
```

Darxo (approved):

PR #1082, opened 2026-08-27
feat(chg): Combo does not remove itself on free skill use.

sSs1897:
All usecases of useForFree in rf:
en garde
net pull
swm charge
swm kick
swm push through
swm tackle
follow up
flail spinner
kingfisher
rebuke
whirling death

The only ones happen in the entity's turn are swm skills, trapper skills and flail skills. I think flail skills already work with combo. Not super sure about current status of trapper or swm skills but I dont see why they should interrupt combo.

In vanilla:
none

https://discord.com/channels/996482927531671692/996858365726691369/1542178187683831973


LordMidas (approved):
Commit message needs rewording:
`feat(chg): do not expire combo on forFree skill use`

PR #1084, opened 2026-09-07
feat: add new ambition to fight champion

Darxo:
I verified, that 
- this ambition spawns (by commenting out the make nobles aware condition).
- it provided its champion spawn chance effect while active
- it can be completed
- it provides the smaller champion spawn buff after completion

## Images

I censored a line, that was added by a different mod and is unrelated to this feature.
image

image

image


Uberbagel:
Is it possible to use the champion's name in the ambition fulfilled event?

sSs1897:
1. oathtakers would need an event similar to their own version of make nobles aware
2. asset manager's resetToDefaults is only called on completing an ambition, what would be a good way to revert the vanilla 90day bonus?

Darxo:
> Is it possible to use the champion's name in the ambition fulfilled event?

Yes, that is pretty easily implemented.
I made it now so, that the name of the last champion you killed will show up in the contract:

image

PR #1085, opened 2026-09-13
feat: nested tooltip highlights entity temporarily 

sSs1897:
Actors with tooltip data requested will highlight with showarrow(true) and schedules an event to turn it off in 2 seconds.

Tested with reach advantage, properly highlights and disappears.
https://discord.com/channels/996482927531671692/1544541743444131941


mod_reforged/hooks/ui/screens/tooltip/tooltip_events.nut
```diff
@@ -1,4 +1,18 @@
...
+	{
+		local entity = ::Tactical.getEntityByID(_data.entityId);
+		if (::MSU.isNull(entity)) return null;
+
+		local ret = __original(_data);
+		if (ret != null && ::isKindOf(entity, "actor") && entity.isAlive() && !entity.isDying()
+			&& entity.isPlacedOnMap() && entity.isDiscovered() && !entity.isHiddenToPlayer())
+			{
+				entity.RF_showArrowTemporary();
+			}
```
LordMidas: 1. Guard statement returns should be on new line.
2. Move entity.isPlacedOnMap() check to the first guard as it is most relevant i.e. it ensures that we are in tactical state and that the entity is on the map.

```suggestion
		if (::MSU.isNull(entity) || !entity.isPlacedOnMap())
			return null;

		local ret = __original(_data);
		if (ret != null && ::isKindOf(entity, "actor") && entity.isAlive() && !entity.isDying()
			&& entity.isDiscovered() && !entity.isHiddenToPlayer())
			{
				entity.RF_showArrowTemporary();
			}
```

mod_reforged/hooks/entity/tactical/actor.nut
```diff
@@ -425,6 +436,13 @@
 
 		__original(_killer, _skill, _tile, _fatalityType);
 	}}.onDeath;
+	
+	q.showArrow = @(__original) { function showArrow( _v )
+	{
+		if (this.m.RF_IsShowingArrow == _v) return;
```
LordMidas: Guard statement returns should be on new line.
```suggestion
		if (this.m.RF_IsShowingArrow == _v)
			return;
```

mod_reforged/hooks/entity/tactical/actor.nut
```diff
@@ -360,6 +362,15 @@
 		// how it is calculated in actor.onMissed).
 		return this.getTile().getZoneOfControlCountOtherThan(this.getAlliedFactions()) * ::Math.round(::Const.Combat.FatigueLossOnBeingMissed * this.getCurrentProperties().FatigueEffectMult * this.getCurrentProperties().FatigueLossOnAnyAttackMult);
 	}}.RF_getZOCEvasionFatigue;
+	
+	// Used for entity highlighting
+	q.RF_showArrowTemporary <- function()
+	{
+		this.showArrow(true);
+		::Time.scheduleEvent(::TimeUnit.Real, 2000, function( _tag ) {
```
LordMidas: Use `_` for parameters that are just place holders and aren't actually used in the function body.
```suggestion
		::Time.scheduleEvent(::TimeUnit.Real, 2000, function( _ ) {
```

mod_reforged/hooks/entity/tactical/actor.nut
```diff
@@ -11,6 +12,7 @@
 
 	q.onInit = @(__original) { function onInit()
 	{
+		this.m.RF_IsShowingArrow = false;
```
LordMidas: Why is this necessary when the value is already set to `false` by default in the `m` table declaration above?

mod_reforged/hooks/entity/tactical/actor.nut
```diff
@@ -360,6 +362,15 @@
 		// how it is calculated in actor.onMissed).
 		return this.getTile().getZoneOfControlCountOtherThan(this.getAlliedFactions()) * ::Math.round(::Const.Combat.FatigueLossOnBeingMissed * this.getCurrentProperties().FatigueEffectMult * this.getCurrentProperties().FatigueLossOnAnyAttackMult);
 	}}.RF_getZOCEvasionFatigue;
+	
+	// Used for entity highlighting
+	q.RF_showArrowTemporary <- function()
```
LordMidas: Would be nice to add a `_duration = 2000` parameter to this function and use that to determine the delay for the arrow to hide.

mod_reforged/hooks/entity/tactical/actor.nut
```diff
@@ -425,6 +436,13 @@
 
 		__original(_killer, _skill, _tile, _fatalityType);
 	}}.onDeath;
+	
+	q.showArrow = @(__original) { function showArrow( _v )
+	{
+		if (this.m.RF_IsShowingArrow == _v) return;
+		this.m.RF_IsShowingArrow = _v;
+		__original(_v);
+	}}.showArrow;
```
LordMidas: A better system would be that if a temporary arrow is being shown for a character and someone mouses out and mouses over the tooltip of the same entity, then we "ignore" the last scheduled event to hide the arrow and instead schedule a new one. This can be accomplished by using a "counter" system for how many times the temporary arrow was triggered on. And in each scheduledEvent to hide the arrow you drop the counter by 1. And only when the counter reaches 0 do you actually hide the arrow.

So instead of a `RF_IsShowingArrow` you have `RF_TemporaryArrowTriggerCount`. It increments by 1 on every call to `RF_showArrowTemporary` and decrements by 1 during the scheduledEvent from that function.

This way we ensure that the arrow only hides 2 seconds after the last triggering of the tooltip.

LordMidas (changes requested):

mod_reforged/hooks/entity/tactical/actor.nut
sSs1897: Do we want this in the configs or just a parameter?

mod_reforged/hooks/entity/tactical/actor.nut
sSs1897: It is not. Was trying to vibe code and I let this slip past. Fixed.

mod_reforged/hooks/entity/tactical/actor.nut
sSs1897: Ideally we would also want the showarrow fade in animation to be a little bit more eye catching like it moves a bit instead of just appear - similar to the net emerge animation, I will look into it. 
Not fully sure how would the counter system work, does nested tooltip query the data again if the entity is queried recently or already exist in the nested stack (ie any sort of caching etc)?

sSs1897: I made it separate as IsShowingArrow is directly used by showArrow(_v) for animation and accepts input from turn sequence bar.

sSs1897:
Update on test result: rapidly hovering in and out of an entity on turn sequence bar can cause showArrow(_v) to not show the arrow properly, which can be resolved by rapidly hovering in and out again. Attempted to fix but not working super well.
Issue is related to async call stack. As it does not show up in normal gameplay and can resolve itself, I will leave it at that.

